PR #9, opened 2026-08-19
Harden the install gates and the sweep's noise floor (Phase 0)

jizhi0v0:
Three pieces of hardening that stand on their own — none of them depends on the
remote-recipe work they were scoped out of, and the first two close gaps that
exist today.

## 1. A `.pkg` is gated on where it installs, not just who signed it

The `.app` routes pair the Team ID check with a bundle-identifier check against
the installed app (`SignatureVerifier` Gate 4). The `.pkg` route had only the
Team ID — and Team ID alone permits any app from the same vendor, which Gate 4's
own comment spells out as Chrome vs Earth. A package runs its install scripts as
root the moment the user confirms.

**Comparing bundle identifiers is not the fix.** Measured against six real vendor
packages, it rejects legitimate updates:

| package | declares the installed app's bundle id? |
|---|---|
| Microsoft Word | yes — one of **206**, so it screens almost nothing |
| WeChat DevTools | yes, but as `com.github.Electron` (stock Electron identity) |
| **Tailscale** | **no** — only the helpers nested inside the bundle |

Tailscale's `pkg-ref` identifier is `com.tailscale.ipn.macsys`; the installed app
is `io.tailscale.ipn.macsys`. A bundle-id gate would refuse a perfectly good
update.

What every package *does* declare is where it writes. This parses the destination
`.app` paths out of each `PackageInfo` (`install-location` as the payload root,
each `<bundle path>` relative to it, and the root itself when it *is* the bundle)
and requires the app being updated to be one of them. Word's package resolves to
Word and AutoUpdate — never Excel — so a same-Team substitution no longer clears
the gate.

Verified through the production code path against all three package shapes:

```
Tailscale.pkg   -> /Applications/Tailscale.app
Word.pkg        -> /Applications/Microsoft Word.app, .../Microsoft AutoUpdate.app
Edge.pkg        -> /Applications/Microsoft Edge.app
OneDrive.pkg    -> /Applications/OneDrive.app
UURemote.pkg    -> /Applications/UURemote.app
WeChat DevTools -> /Applications/wechatwebdevtools.app
```

**Deliberate limitation:** a package whose layout cannot be read yields no
destinations, and that falls back to the Team-only gate rather than blocking an
install that works today — the other 13 pkg recipes are untested. The comment
says to tighten to fail-closed once they are confirmed. Worth a second opinion on
whether that trade is the right one.

## 2. Changelog URLs must be plain https before they are rendered

`changelogURL` is the one recipe field that becomes an attacker-choosable origin
loaded in-process, in a window sitting next to credentials for other services.
Every recipe is compiled into the signed binary today, so this is defence in
depth rather than a live hole — but it is the field most likely to be served
remotely first.

Rejects non-https, URLs carrying credentials (`https://vendor.com@evil.example/`),
IP-literal hosts, and non-standard ports. A rejected URL returns nil so the caller
falls through to its existing empty state.

No host allowlist: that only becomes meaningful once recipes stop being compiled
in, and guessing at vendor domains now would break the many products whose docs
site is on a different domain than their download endpoint.

Regression tests walk `VendorProbeRegistry`, `ChangelogRecipeRegistry` and
`ChangelogCatalog.pages` rather than a hand-written list, so a recipe added later
with an `http://` URL fails here instead of silently losing its notes pane.

## 3. The sweep's infra gate is raised ahead of a faster cadence

The sweep is the real ceiling on how fast a breakage is noticed — Zotero 10.0 sat
undetected because detection failures are silent by design. Going to four sweeps a
day fixes that, but the noise gates are counted in *sweeps*, so the frequency
change shortens them fourfold. `infraThreshold` 5 -> 20 restores the original ~5
days of wall clock; `actionableThreshold` stays at 2 on purpose (5xx/429 are
classified `.infra`, not actionable).

**The schedule itself is not in this repo.** `.github/workflows/recipe-verify.yml`
is gitignored and no longer runs anything — Actions last ran it 2026-08-13, and
since 2026-08-16 the sweep is a launchd agent on the mini that pulls and rebuilds
before each run. So this must merge *first*; the cadence is raised on the mini
afterwards, or there is a window where 6-hourly sweeps run against a 30-hour infra
gate.

## Verification

```
make test    -> 768 tests / 62 suites passed (core)
                101 tests / 16 suites passed (CLI)
                2 known issues, both pre-existing
make install -> builds, signs (Developer ID), deploys, runs
```

Real packages were downloaded and parsed through the shipping code, not fixtures
alone; the temporary test that did so was removed after it passed.

### DIFF
--- a/App/Sources/WorkbenchWindowView.swift
+++ b/App/Sources/WorkbenchWindowView.swift
@@ -1028,7 +1028,7 @@ private struct FormulaDetailPane: View {
         case .loaded(let release):
             if let changelog = release.changelog {
                 ChangelogEntriesView(changelog: changelog)
-            } else if let url = release.pageURL {
+            } else if let url = ChangelogURLPolicy.displayable(release.pageURL) {
                 CachedWebView(url: url)
             } else {
                 noNotes
--- a/App/Sources/WorkbenchWindowView.swift
+++ b/App/Sources/WorkbenchWindowView.swift
@@ -1059,7 +1059,8 @@ private struct DetailHeader: View {
 
     /// The vendor page to link out to in Release Notes mode.
     private var changelogURL: URL? {
-        result.remote?.changelogURL ?? ChangelogCatalog.url(forBundleID: result.app.bundleID)
+        ChangelogURLPolicy.displayable(
+            result.remote?.changelogURL ?? ChangelogCatalog.url(forBundleID: result.app.bundleID))
     }
 
     var body: some View {
@@ -1149,7 +1150,8 @@ private struct ReleaseNotesPane: View {
     @Bindable var model: AppListModel
 
     private var changelogURL: URL? {
-        result.remote?.changelogURL ?? ChangelogCatalog.url(forBundleID: result.app.bundleID)
+        ChangelogURLPolicy.displayable(
+            result.remote?.changelogURL ?? ChangelogCatalog.url(forBundleID: result.app.bundleID))
     }
 
     var body: some View {
--- a/CLI/Sources/DuoKit/Baseline.swift
+++ b/CLI/Sources/DuoKit/Baseline.swift
@@ -85,7 +85,14 @@ public struct Baseline: Codable, Sendable {
     }
 
     /// Actionable sweeps below this never produce an issue. One sweep of trouble
-    /// is noise; two consecutive sweeps a day apart is a pattern.
+    /// is noise; two consecutive sweeps is a pattern.
+    ///
+    /// Left at two when the sweep moved from nightly to every six hours, which
+    /// tightens the wait from two days to twelve hours. That is the point of the
+    /// change — a broken recipe is invisible until a sweep reports it — and it
+    /// stays safe because the noisy failures it could otherwise catch (a vendor's
+    /// maintenance window returning 5xx or 429) are classified `.infra`, not
+    /// actionable, and are gated separately below.
     public static let actionableThreshold = 2
 
     /// Consecutive *unreachable* sweeps before an endpoint is reported as gone.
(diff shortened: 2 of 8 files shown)


jizhi0v0:
Review found four real defects. All four reproduced against the shipping code before fixing, and all four are fixed in the three commits just pushed.

### The parser was wrong in the dangerous direction

The `.app`-substring rule produced **non-empty but incorrect** destination sets, which the empty-set fallback does not catch — so it was a hard refusal of legitimate updates, not merely a weaker gate. Reproduced:

```
./Library/Application Support/com.vendor.app/Real.app  ->  [".../com.vendor.app"]   real bundle dropped
./Widget.appex  (install-location=/Applications)       ->  ["/Applications/Widget.app"]  fabricated
<bundle search-path="/Evil.app" path="./Good.app">     ->  both, via the unanchored regex
```

Worth noting the suggested fix — take the first component that *ends with* `.app` — has the same failure, because `com.vendor.app` genuinely ends with `.app` and is indistinguishable from a bundle by name. Neither first nor last is reliable: first truncates on a reverse-DNS directory, last picks a helper out of a package whose payload root is the app itself (Tailscale). Now collects **every** prefix ending in an app bundle, so the real destination survives either shape. The extra entries are paths the package really writes and cannot make a different product match.

Parsing moved to `XMLDocument`, which removes the whole table of regex defects at once (entities, comments, single quotes, `search-path`).

Re-verified through the shipping code against all six real packages.

### Apps outside `/Applications` were being refused

Confirmed: `AppScanner.defaultLocations` includes `~/Applications`, `/Applications/Utilities` and both Input Methods directories, while every real vendor package names the system location. Path equality alone refused those updates. Now falls back to comparing the bundle name, which is what actually separates one product from another.

### The gated web views were the wrong ones

Confirmed and the sharpest finding: `WorkbenchWindowView.swift:1032` renders `FormulaRelease.pageURL`, sourced at runtime from `brew info` homepage metadata or the GitHub API — the only one of the three that is already remote. And socat's Homebrew homepage really is `http://www.dest-unreach.org/socat/`, so this was live, not hypothetical. Now gated.

### `commentEverySweeps`

Confirmed, 7 -> 28. Agreed that two constants needing the same manual rescaling argues for gating on the already-persisted `infraSince` timestamp instead; left as a follow-up rather than widening this PR.

### Not taken

- **Per-recipe `destinationsVerified` opt-in** — good idea, but it adds a field to the install spec and that is beyond this change. The six tested packages and the thirteen untested ones still share one code path.
- **Reading the Bom instead of `PackageInfo`** — agreed it is structurally better; follow-up.
- **Fail-closed when a `PackageInfo` parses but yields no `.app`** — deferred with the above, since the parser rewrite already removes the cases that made this most likely.

Also corrected: the Word fixture is now labelled constructed rather than captured, and the verification table in the PR body overstated where AutoUpdate lands — it is under `/Library/Application Support/Microsoft/MAU2.0`, not `/Applications`.

`make test` -> 777 core / 101 CLI passing, 2 pre-existing known issues. `make install` builds, signs and runs.

PR #55, opened 2026-08-24
A locked file must not make a backup undeletable

jizhi0v0:
`ditto` copies BSD file flags faithfully — correct for a backup — so a `uchg`
file the user set on an installed app arrives in the store along with the
bundle. `removeItem` then fails with `EPERM` on that one file and refuses the
whole tree.

Every removal in `BackupStore` was `try?`, so the failure was silent. The
backup stayed, the space was never reclaimed, and the delete sheet reported
success while deleting nothing.

### How it turned up

On a real machine, ToDesk's `Contents/advInfo.json` had been locked by hand
(there is a `.bak-…-noads` sitting beside it). Every ToDesk backup had become
undeletable, along with a `.staging-` remnant a refused save had left behind.

```
$ ls -lO …/Backups/.staging-com.youqu.todesk…/ToDesk.app/Contents
-rw-r--r--@ 1 bobby admin uchg 340 May 21 17:15 advInfo.json

$ ditto <that directory> /tmp/uchgtest && rm -rf /tmp/uchgtest
rm: uchgtest/ToDesk.app/Contents/advInfo.json: Operation not permitted
rm: uchgtest/ToDesk.app/Contents: Directory not empty
```

### The fix

`forceRemove` tries the plain removal first and only on failure clears the
user-settable flags (`UF_IMMUTABLE`, `UF_APPEND`) and retries. `lstat`/
`lchflags` rather than the follow-the-link pair, so a symlink inside a bundle
has its own flags cleared and a link cannot reach outside the store.

Clearing is safe **here specifically** because the target is always a copy this
store made, never the user's own file. The system flags (`schg`) need root and
their presence is a genuine reason to stop — those still fail, but now they say
so in the log instead of being swallowed.

A restore is unaffected: it copies its own bundle out, flags and all.

### Verification

`aBackupHoldingALockedFileCanStillBeRemoved` goes through the real
`save` → `remove(forKey:)` path rather than deleting a fixture, so it fails if
the flag handling is ever dropped from the store's own path. Confirmed red
before green by commenting out the one `clearUserFlags` call.

```
DuoUpdaterCore  1022 tests / 81 suites passed
CLI              119 tests / 18 suites passed
localizable keys agree — 410 in the catalog, all reachable
```

Verified on the real store afterwards: the `.staging-` remnant that had
survived every previous sweep was removed.

### DIFF
--- a/DuoUpdaterCore/Sources/DuoUpdaterCore/Install/BackupStore.swift
+++ b/DuoUpdaterCore/Sources/DuoUpdaterCore/Install/BackupStore.swift
@@ -169,7 +169,7 @@ public enum BackupStore {
         // across a crashed prior attempt.
         try fm.createDirectory(at: root, withIntermediateDirectories: true)
         let staging = root.appendingPathComponent(".staging-\(key)", isDirectory: true)
-        try? fm.removeItem(at: staging)
+        forceRemove(staging)
         try fm.createDirectory(at: staging, withIntermediateDirectories: true)
 
         let name = appPath.lastPathComponent
@@ -183,7 +183,7 @@ public enum BackupStore {
         // nothing worth storing.
         let unreadable = BackupManifest.unreadableFiles(in: appPath)
         guard unreadable.sealed.isEmpty else {
-            try? fm.removeItem(at: staging)
+            forceRemove(staging)
             throw BackupError.payloadUnreadable(unreadable.sealed.first ?? appPath.path)
         }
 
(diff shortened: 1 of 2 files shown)


jizhi0v0:
Superseded by #752.

Half of this landed differently in `4259432` (v0.3.66, 2026-08-26), as
`clearUserImmutableFlags(under:)` — a `chflags -R nouchg` on the copy before it
is stored, on the generation it supersedes, and on the staging leftovers it
sweeps. The `BackupStore` this PR was written against no longer exists, so it
has conflicted with `main` since, and a conflicting PR does not even get a
Claude review: GitHub withholds `pull_request` events from one.

What `4259432` did **not** cover is the half this PR started from: the paths
that *delete* a backup. `remove(forKey:)`, `pruneOrphans()`, the legacy-key
cleanup and the rollback scratch were all still a bare `try? removeItem`, so a
backup taken before v0.3.66 could not be deleted, cleaned up, or rolled back
onto — silently. #752 carries that across to today's `BackupStore`, keeps the
`lstat`/`lchflags` detail from here, and adds the regression tests.

PR #57, opened 2026-08-24
Keep the rollback backups on another disk

jizhi0v0:
> Stacked on #55 — base retargets to `main` once that merges. The diff here is
> the backup-destination work only.

Backups are a real second copy of every app that gets updated, and on a Mac
that is short of space they are the main reason people turn the safety net off.
This lets the store point at an external disk or a NAS, while the boot volume
keeps only what has not been moved yet.

### The destination holds archives, not bundles

A `.app` is a directory whose meaning lives as much in its metadata as in its
bytes. An exFAT stick or an SMB share stores that approximately at best, which
breaks the restored app's code signature *and* makes the manifest recorded at
save time disagree with the tree read back — both of which surface on the day
the backup is actually needed.

Packing each bundle into a single Apple Archive (`/usr/bin/aa`) moves that
problem out of the destination's hands: the metadata is encoded inside the byte
stream, so the destination only has to hold one opaque file. It also keeps
`integrityHolds` untouched — all three rollback paths now compute their
manifest on APFS, so the comparison never straddles two filesystems.

Measured on Claude.app (802 MB, 3421 files):

| algorithm | archive | time |
|---|---|---|
| `lzfse` (default) | 330 MB (41%) | 1.8 s |
| `lzma` ("smallest") | 242 MB (30%) | 22 s |
| extract (`lzfse`) | — | 1.1 s |

3421 files becoming one is the bigger win on a share, where per-file round
trips rather than bandwidth are what make a NAS backup slow.

Round-tripping preserves the vendor code signature, asserted separately from
the manifest comparison — a tree can match byte for byte and still restore an
app whose seal is broken, because the manifest does not hash extended
attributes. AppCleaner, TestFlight, Pearcleaner and Claude all come back
`codesign --verify --deep --strict` clean.

### An unplugged disk is a delay, not a missing rollback point

`save` always writes to the boot volume first, exactly as it always has; a
background queue drains that outbox onto the disk afterwards. Install latency
is unchanged, the boot volume is reclaimed a minute later, and a disk that is
in a bag simply means the backup waits.

The transfer's ordering is the whole safety argument: the archive lands, its
digest is read back **from the destination**, the sidecar is written, and only
then is the local copy removed. At every point before that last step there is a
complete, restorable backup somewhere.

An install with the disk away still takes a backup. Only when the boot volume
is *also* tight does it skip — and then it says so on the row rather than
leaving the user to discover it at rollback time.

### Things a real machine taught this, which are now regression tests

- **Never conjure a directory at an unreachable destination.** `save` used to
  reach it through `createDirectory(withIntermediateDirectories:)`, which on a
  detached disk does not fail — it builds the path on the boot volume, squats
  the mount point, and the real disk then mounts beside the decoy as
  `Archive 1` with backups split across two places that each look correct.
- **The store owns its own subdirectory.** Rooted at the folder the user
  picked, a Documents folder read as 31 GB of backup storage and Clean Up would
  have `removeItem`'d their folders.
- **Adopting is idempotent.** Re-picking the shown path nested a second store
  inside the first; the newly empty one became the destination while every
  existing backup sat one level up, invisible.
- **What is owed is read from the sidecars, not from a flag written at save
  time.** Otherwise backups taken before a disk was ever chosen never move —
  on the machine this was built against, that was 23.87 GB, which is exactly
  the part the user was trying to reclaim.
- **A drain skips a failing entry and continues.** One backup that could never
  transfer — a remnant a refused save left behind — used to hold all 17
  remaining backups on the boot volume indefinitely.
- **Volume identity is a UUID we write ourselves.** Not
  `volumeUUIDStringKey` (not guaranteed on exFAT or SMB), and emphatically not
  `volumeIdentifierKey`, which is a mount-scoped token that changes on replug
  and would report a different disk every morning.
- **A pre-existing restore race**, surfaced by the new tests: the scratch
  directory was keyed on the app alone, so two concurrent restores of the same
  key deleted each other's working tree mid-copy.

### CLI

`duo backups sync` / `verify [--deep]` / `probe <path>`, and `list`/`verify`
now say on stderr when the disk is not reachable instead of reporting an empty
store. `probe` writes nothing, so it is safe to point at a share you are only
considering — and it is the honest answer to "will this work on my NAS?", since
SMB and NFS cannot be faked in a test.

`verify` reports three states, not two. A backup taken before fingerprints
existed is neither a pass nor a failure; calling it either would be a
reassurance nobody earned, or an error that trains people to ignore the exit
code.

### Verification

```
DuoUpdaterCore  1093 tests / 86 suites passed
CLI              119 tests / 18 suites passed
localizable keys agree — 448 in the catalog, all reachable
```

Exercised end to end on a real machine with a USB stick: 52 backups moved,
9.18 GB from ~23.4 GB (39%), every SHA-256 matching its sidecar, no nesting and
no sidecar-less directories left in the store.

### Not in scope

`CHANGELOG.md` — this repo names each section by version, so the user-facing
prose goes in when the version is bumped.

### DIFF
--- a/App/Resources/Localizable.xcstrings
+++ b/App/Resources/Localizable.xcstrings
@@ -3117,6 +3117,47 @@
         }
       }
     },
+    "A different disk is mounted here": {
+      "extractionState": "manual",
+      "localizations": {
+        "de": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Hier ist eine andere Festplatte eingebunden"
+          }
+        },
+        "es": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Aquí hay montado otro disco"
+          }
+        },
+        "fr": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Un autre disque est monté ici"
+          }
+        },
+        "ja": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "別のディスクがマウントされています"
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Здесь подключён другой диск"
+          }
+        },
+        "zh-Hans": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "此处挂载的是另一块磁盘"
+          }
+        }
+      }
+    },
     "A few quick steps let Duo Updater keep your apps updated without interrupting you. You set these once — they persist across launches.": {
       "extractionState": "manual",
       "localizations": {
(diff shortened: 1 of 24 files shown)


jizhi0v0:
Superseded by #763, merged as `57a7aa39`.

This PR held the first four commits of the work (`50b1d01f..b274d6e0`, against `fix/backup-locked-file-undeletable`). Everything in it is now on `main`, along with the 38 commits that followed it and never reached a pull request — disk selection, the transfer queue's reporting, the comparison against a backup on a disk, and the record that keeps that comparison from unpacking the archive.

Its branch `feat/backup-to-external-disk` is left in place, as is `fix/backup-locked-file-undeletable`.

PR #65, opened 2026-08-25
Give App Store apps a rollback point, and make restoring one possible

### LINKED ISSUE #56: App Store apps get no rollback point, and the stated reason is wrong
`InstallCoordinator.wantsBackup` excludes the App Store route:

```swift
case .appStore:  return false
```

The doc comment above it explains why: *"the store can always re-fetch a prior
build, so a local copy of a multi-gigabyte bundle is dead weight."*

**That is not true.** The App Store serves only the current version of an app.
There is no user-facing way to download or install a previous build; the single
exception is the "last compatible version" prompt Apple shows when the current
release requires a newer macOS than the machine is running, which is not
rollback.

- [Apple Community — revert to a previous app version](https://discussions.apple.com/thread/255068136)
- [Apple Developer Forums — Install old app versions from the App Store](https://developer.apple.com/forums/thread/700304)

So for every App Store app, Duo Updater applies an update with no way back,
while the product's premise is that an update can be undone. That is the gap
worth closing, not the comment.

### It is technically feasible — measured, not assumed

App Store apps are `root:wheel drwxr-xr-x` in `/Applications`, and every file
in them is world-readable, so a copy needs no privileges:

```
$ find /Applications/Keka.app ! -perm -o+r        # no output
$ ditto /Applications/Keka.app <scratch>/Keka.app                    OK
$ codesign --verify --deep --strict <scratch>/Keka.app               OK
$ ls <scratch>/Keka.app/Contents/_MASReceipt/receipt                 present
```

The receipt travels with the bundle, so a restored older copy carries the
receipt that was issued for it rather than a mismatched newer one.

The size objection is also weaker than it looks: on APFS, `ditto` between
`/Applications` and the backup store is already a copy-on-write clone. Time
tracks entry count, not bytes — Docker.app, 2218 MB across 267 entries, copies
in 0.07 s, which is 31 GB/s and therefore not real I/O. A 2193 MB clone
consumed no measurable space (`df` delta within noise, and of the wrong sign).
The blocks only become the backup's own once the update replaces the original.

### What actually needs deciding

1. **Restoring writes over a `root:wheel` bundle**, which needs elevation. The
   machinery exists — the `.installer` route already handles root-owned bundles
   and records a declined prompt per install path — so this is wiring, not new
   capability.
2. **The store will immediately re-offer the update** after a rollback. The
   rollback row should probably say so, the way the `.pkg` note already warns
   that helpers and daemons stay at the newer version.
3. **`mas` is not involved in a restore** — the swap is `InPlaceSwap`, the same
   as every other route.

### Scope

Independent of the external-disk backup work: `wantsBackup` and
`BackupStore.save` are both on `main` today, and nothing here touches the
backup destination.

jizhi0v0:
Closes #56.

## What the issue got right, and what it missed

The issue's premise checks out: `wantsBackup` excluded `.appStore` on the
grounds that "the store can always re-fetch a prior build", and the App Store
serves only an app's *current* version. So the store route was the one route
that applied an update with no way back.

It also flagged the right open question — "restoring writes over a `root:wheel`
bundle, which needs elevation" — but assumed the machinery already handled it.
It did not, and that turned out to be a live bug wider than this issue.

## The measurement that changed the shape of the fix

`InPlaceSwap.needsElevatedReplace` tested only the **parent directory**. For a
store app that reads: `/Applications` is `root:admin` 775 → writable → "no
password needed". But the bundle itself is `root:wheel` 755, and the swap ends
by deleting the bundle it replaced — which needs write permission on the
directories *inside* it.

Built with those exact permission bits, `FileManager.replaceItemAt`:

- throws `NSCocoaErrorDomain 513` (`NSFileWriteNoPermissionError`),
- leaves the app on disk **unchanged**,
- leaves its `.duoupdater-staged-*` sibling behind.

513 is exactly what `isAppManagementDenial` matches — so every such failure was
reported as `AppManagementRequiredError`, sending the user to grant an App
Management permission that could never help. The obstacle is POSIX ownership,
not TCC.

On this machine that is **22 of 22** store-installed apps plus **6** further
root-owned ones. Note who it was biting: a *standard* user cannot write
`/Applications` at all, so the parent-only test already said yes for them. It
was **admin accounts** — the ordinary single-user Mac — that took the path that
could not succeed.

The doc comment already described the correct rule ("writability of the bundle
*and* of its parent"); only the code disagreed. Sparkle's
`SPUSystemNeedsAuthorizationAccessForBundlePath` is the same pair.

## Changes

1. **`needsElevatedReplace`** requires both the bundle and its parent to be
   writable. Root-owned bundles now take the privileged swap, which works.
2. **`privilegedReplace` keeps the replaced bundle's owner.** `ditto` as root
   preserves the *source's* owner, and our sources belong to the user — so
   without a `chown` the elevated path silently converted a `root:wheel` app
   into a user-owned one. Observed on a real store app. On a shared Mac that
   would leave one admin able to edit an app another user runs.
3. **`wantsBackup(.appStore)` is true**, and the doc comment that stated the
   wrong reason is replaced with the measurements behind the right one.
4. **The route is recorded** (`fromAppStore` in the sidecar) so the restore path
   can say the one thing that is specific to it: the store re-offers the update
   immediately, and re-applies it unattended when automatic app updates are on.
   Surfaced in `duo backups restore` (before the confirmation, beside the `.pkg`
   note) and on the menu-bar row after a rollback lands. Localized in all 7
   languages.

## Verification

`make test` — 1090 tests / 87 suites and 125 / 19 pass (2 pre-existing known
issues), plus `check_localizable_keys.py` green at 414 keys. `make install`
builds, signs and deploys with all 7 languages landing.

Beyond the unit tests, against real bundles:

| check | result |
|---|---|
| all 22 store apps readable without privileges | 0 unreadable files in any of them |
| predicate, before → after | 0/22 → 22/22 report `needsElevatedReplace` |
| save → mutate → restore, real store bundle | signature verifies, `_MASReceipt` intact, retention intact |
| live privileged rollback over `root:wheel` `/Applications/Developer.app` | version unchanged, codesign OK, receipt present, **owner still `root:wheel`**, no leftovers |

The ownership fix has its own red→green: the same live rollback run *before* it
left the bundle `bobby:staff`; after it, `root:wheel` both sides.

## Not in scope

- `CHANGELOG.md` — this repo writes release notes in the `release: prepare X`
  commit, not in feature PRs.
- Disk cost is real but unchanged in kind: `ditto` into the backup store is an
  APFS clone (2750 MB of Word in 8.7 s, ~16 MB of new blocks), and the blocks
  become the backup's own only once the update replaces the original. Gated by
  `keepBackups`, retention stays at one, and the Backups sheet shows the total.

### DIFF
--- a/App/Resources/Localizable.xcstrings
+++ b/App/Resources/Localizable.xcstrings
@@ -13717,6 +13717,47 @@
         }
       }
     },
+    "Rolled back, but %@ updates through the App Store — it will offer this update again, and re-install it on its own if automatic app updates are on.": {
+      "extractionState": "manual",
+      "localizations": {
+        "de": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Zurückgesetzt – aber %@ wird über den App Store aktualisiert: Der Store bietet dieses Update sofort wieder an und installiert es von sich aus erneut, wenn automatische App-Updates aktiviert sind."
+          }
+        },
+        "es": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Se revirtió, pero %@ se actualiza a través del App Store: volverá a ofrecer esta actualización y la instalará por su cuenta si las actualizaciones automáticas de apps están activadas."
+          }
+        },
+        "fr": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Version restaurée, mais %@ se met à jour via l’App Store : celui-ci proposera de nouveau cette mise à jour et la réinstallera de lui-même si les mises à jour automatiques des apps sont activées."
+          }
+        },
+        "ja": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "ロールバックしましたが、%@ は App Store 経由で更新されます。ストアはこのアップデートをすぐに再提示し、App の自動アップデートがオンなら自動で再インストールします。"
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Откат выполнен, но %@ обновляется через App Store: он снова предложит это обновление и установит его сам, если включены автоматические обновления приложений."
+          }
+        },
+        "zh-Hans": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "已回滚，但 %@ 通过 App Store 更新：商店会立刻再次提示这次更新；如果开启了 App 自动更新，它还会自行装回去。"
+          }
+        }
+      }
+    },
     "Rollback": {
       "extractionState": "manual",
       "localizations": {
--- a/App/Sources/AppListModel.swift
+++ b/App/Sources/AppListModel.swift
@@ -2237,7 +2237,8 @@ final class AppListModel {
         // backup timestamp hadn't moved.
         let backupRoute = InstallCoordinator.route(
             for: result, requiresInstaller: requiresInstaller(result))
-        if prefs.keepBackups, InstallCoordinator.wantsBackup(backupRoute) {
+        if prefs.keepBackups, InstallCoordinator.wantsBackup(backupRoute),
+           !(backupRoute == .appStore && appStoreRouteWillNotInstall(result)) {
             await backupCurrent(result, route: backupRoute)
         }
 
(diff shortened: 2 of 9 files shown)


jizhi0v0:
## Code review pass — 4 findings, all fixed

| # | finding | fix |
|---|---|---|
| 1 | Store backup ran before the `.appStore` route's own early-outs, so a clone + whole-bundle SHA-256 was paid and thrown away — and left a rollback point whose version equals the one still installed | `appStoreRouteWillNotInstall` predicts the three cheap bail conditions (7271906) |
| 2 | `chown -R` on `.duoupdater-new` before the renames made an interrupted swap leave a root-owned leftover that `recoverInterruptedSwaps` silently cannot delete | chown moved to after the final rename (8d37e36) |
| 3 | The "APFS clones it, so it's free" cost note omitted `BackupManifest.compute`, which is O(bytes) — 8.7 s for Word on top of 8.7 s of ditto | doc comment now carries the hashing cost (7271906) |
| 4 | Every root-owned bundle now takes the privileged path, and the apply pool allows 2 concurrent swaps → two stacked administrator panels, neither naming its app | a lock held across the panel serialises them (8d37e36) |

Re-verified live against `/Applications/Developer.app` (`root:wheel`) after the
script was reordered: still `root:wheel`, signature and receipt intact, no
leftovers. `swift test --package-path CLI` 125/125,
`check_localizable_keys.py` green at 414 keys, `make install` builds and
deploys with all 7 languages.

### One pre-existing failure, not from this branch

`vendorResolvesInstallPlans` fails on `com.google.android.studio [canary]`:

```
resolved .../2026.1.4.6/android-studio-quail4-rc2-mac_arm.dmg, which carries
no canary marker (expected /-canary[0-9]*-mac/) — the install may be crossing channels
```

Reproduced identically on unmodified `main` in a detached worktree, so it is
a vendor-side condition (Google currently serving the beta RC artifact on the
canary channel), not a regression here. Worth its own issue.

PR #78, opened 2026-08-27
fix(vendor-probe): Android Studio canary/beta ignore feed publish-date ordering

### LINKED ISSUE #76: Android Studio preview probes assume the releases feed is version-ordered; it is publication-ordered
The Android Studio preview probes take the **first** match of their channel set out of `jb.gg/android-studio-releases-list.json`, and `VendorProbeRecipe.swift` states the assumption that makes that correct:

> The feed is newest-first, so the FIRST channel-set match is that set's newest build.

The feed is ordered by **publication date**, not by version. When two feature trains are open at once, those are not the same thing.

## Observed

Fetched 2026-08-27 (671 items), in feed order:

```
[0] RC       Android Studio Quail 4  | 2026.1.4 RC 2      build=AI-261.26222.65.2614.16166871
[1] Canary   Android Studio Rabbit 1 | 2026.2.1 Canary 2  build=AI-262.9437.185.2621.16128175
[2] Canary   Android Studio Rabbit 1 | 2026.2.1 Canary 1  build=AI-262.9437.185.2621.16076543
[3] RC       Android Studio Quail 4  | 2026.1.4 RC 1      build=AI-261.26222.65.2614.16087418
```

2026.1 has reached RC while 2026.2 has just opened on Canary, and the 261 RC was published *after* the 262 Canary — so it holds `[0]`. The canary recipe's set is `(?:Canary|Beta|RC)`, so its first match is that RC, on the older train.

`duo verify` caught it as a backwards move:

```
⚠ WARN  vendor:com.google.android.studio:canary
    resolved  2026.1.4 RC 2
    warning   version went BACKWARDS since the last sweep (2026.2.1 Canary 2 → 2026.1.4 RC 2)
```

## What this is NOT

Worth stating, because "canary resolved to an RC" looks alarming and is not the bug:

- **Not a channel leak.** The STABILITY FLOOR design is intact and deliberate: Canary accepts `{Canary, Beta, RC}`, so an RC is a legal answer for a canary install. Beta's pattern is `(?:Beta|RC)` — a Canary build cannot match it textually. Stable is a separate recipe on a separate URL.
- **Not a downgrade.** Checked against `VersionComparator` with the real build ids:
  ```
  isNewer(AI-261.26222.65.2614.16166871, than: AI-262.9437.185.2621.16128175) = false
  isNewer(AI-261.26222.65.2614.16166871, than: AI-261.26222.65.2614.16027044) = true
  isNewer(AI-262.9437.185.2621.16128175, than: AI-261.26222.65.2614.16166871) = true
  ```
  Someone on 2026.2.1 Canary 2 is correctly told they are up to date; the 261 RC is never offered to them.
- **Not a version/download mismatch.** All three patterns land on the same item against the real body:
  ```
  version 1st match : AI-261.26222.65.2614.16166871   @off 1310
  display 1st match : 2026.1.4 RC 2
  dmg     1st match : 2026.1.4.6/android-studio-quail4-rc2-mac_arm.dmg   @off 315
  ```

## What is actually wrong

Someone still on the 261 canary train (e.g. 2026.1.4 Canary 4) is offered **2026.1.4 RC 2** rather than **2026.2.1 Canary 2**. The RC is a legal, more-stable, same-train build, so installing it does no harm — but it is not what the stability floor intends. Its own doc says a canary install moves onto an RC "when no newer Canary exists", and here a newer Canary does exist; it just was not published most recently.

Once they take the RC, the probe keeps answering with the same build, so they sit on 261 until Google publishes something on 262 that is chronologically newest. Self-correcting, but the lag is unbounded in principle.

## Why this is not a one-line change

`selectHighest` (`VendorProbeRecipe.swift:233`) is exactly the mechanism for "decide the reported version by comparison rather than by position", and flipping it on the canary recipe would pick `AI-262.9437…` correctly.

But the install spec's `urlSource: .bodyPattern(…)` is *also* first-match, and it has no such flag. Set `selectHighest` on the version alone and the row would read `2026.2.1 Canary 2` while the button fetched `quail4-rc2-mac_arm.dmg` — a genuine wrong-build install, which is worse than the lag it fixes. `VendorProbeRecipe.swift:53` already warns about this exact combination.

These installs are Toolbox-managed → detection-only, so the install spec is suppressed at the source today. That makes the mismatch latent rather than live, which is a reason to fix it carefully, not a reason to ignore it.

## Impact

Low. No wrong channel, no downgrade, no mismatched download. A canary install on the outgoing train lags one feature train behind until the incoming one publishes again.

Found while reviewing 0.3.67; unrelated to that release, which is why it is filed rather than folded in.

jizhi0v0:
## What was wrong

`jb.gg/android-studio-releases-list.json` is ordered by **publication date**, not by version. `VendorProbeRecipe.swift` documented the opposite assumption ("the feed is newest-first, so the FIRST channel-set match is that set's newest build"). When two feature trains are open at once — a common state, not an edge case — that assumption breaks.

**Real evidence**, fetched 2026-08-27 (671 items):

```
[0] RC      2026.1.4 RC 2      build AI-261.26222.65.2614.16166871   Aug 25
[1] Canary  2026.2.1 Canary 2  build AI-262.9437.185.2621.16128175   Aug 20
[2] Canary  2026.2.1 Canary 1  build AI-262.9437.185.2621.16076543   Aug 17
[3] RC      2026.1.4 RC 1      build AI-261.26222.65.2614.16087418   Aug 17
```

`262 > 261` (`VersionComparator`), so `[1]` is the actually-newest build — but it was published *before* the RC at `[0]`, so the canary recipe's `(?:Canary|Beta|RC)` first-match landed on the older 261 train's RC. This is exactly what `duo verify` caught:

```
⚠ WARN  vendor:com.google.android.studio:canary
    resolved  2026.1.4 RC 2
    warning   version went BACKWARDS since the last sweep (2026.2.1 Canary 2 → 2026.1.4 RC 2)
```

Not a channel leak (the stability floor — Canary → newest of {Canary, Beta, RC} — is intact and an RC is a legal answer) and not a downgrade (`VersionComparator` correctly ranks 262 > 261, so nobody on 262 was ever offered the 261 RC). It's a lag: a Canary-train user sits one feature version behind until the newer train happens to publish something more recently than the older train's next RC.

## Why the naive fix (flip `selectHighest`) is wrong

`VendorProbeSource.probeOutcome` reads **three** things from the response body by first-match, and none of them respect `selectHighest`:

1. `version` — the `versionPattern` capture (line ~382, pre-fix)
2. `display` — the `displayVersionPattern` capture (~line 409)
3. the install spec's `.bodyPattern` capture, for the download URL (`resolveInstall`)

Setting `selectHighest: true` on the canary recipe's version alone would have picked `AI-262.9437…` correctly for the *row*, while `displayVersionPattern` and the install URL stayed first-match and kept resolving the RC's own entry — the row would read `2026.2.1 Canary 2` while the download button fetched `quail4-rc2-mac_arm.dmg`. `VendorProbeRecipe.swift:53` already documents this exact failure mode (measured on Docker's appcast) for the general case; this issue is the same shape hitting Android Studio.

## The fix

A new opt-in primitive, `VendorProbeRecipe.entryStartPattern: String?` (default `nil`, so every other recipe in the registry is byte-identical to before). When set, `VendorProbeSource.probeOutcome` computes one `scope`:

- slice the body into entries at every match of `entryStartPattern` (that match's start → the next match's start; last entry runs to the end of the body)
- keep the entries where `versionPattern` matches
- pick the entry whose extracted version compares highest, via the same `VersionComparator` path `VendorProbeRecipe.highestVersion`/`highestVersionedURL` already use (no second comparator)

`version`, `displayVersionPattern`, `publishedAtPattern`, **and** the install spec's URL are then all resolved against that *one* winning entry substring — so they can never land on different releases. Falls back to today's whole-body behavior when the pattern matches fewer than two entries, or when no entry's `versionPattern` matches (better a possibly-stale first-match answer than no answer at all).

Applied to both Android Studio preview recipes (canary and beta) with `entryStartPattern: #"\{"date":""#` — each feed item starts with its own `"date"` key, and slicing there keeps a whole item (its download links included) together. Verified against the real 1MB/671-item feed that this pattern produces exactly 671 entries, one per item.

## What I deliberately did NOT touch

- `ReleaseChannel.detect`, `AppScanner`, `ToolboxSource`, or Toolbox gating.
- The stable Android Studio website recipe (unaffected — single-version page, no ordering ambiguity).
- Any other order-fragile recipe I noticed in passing. None jumped out during this change, but I did not go looking — out of scope for this issue.
- No new flags/knobs beyond `entryStartPattern` itself.
- `VendorAppcastDeltas.patches` still reads `body.text` (not `scope`) — deliberately: it parses the whole response as a Sparkle XML appcast, which a JSON entry fragment wouldn't parse as, and it does its own version-match safety independent of scoping. Commented in place.

## Tests

Added to `DuoUpdaterCoreTests/VendorProbeTests.swift`:
- `androidStudioCanaryPicksTheNewestTrainNotTheMostRecentlyPublishedEntry` — red→green regression using the real captured feed ordering above. Runs the *actual shipping* canary recipe (only `url`/`entryStartPattern` swapped) end-to-end through `VendorProbeSource.probeDiagnostic` against a loopback stub of that real body. Confirmed red by temporarily reverting the `VendorProbeSource` engine change alone (keeping the registry field) — it still reproduced the bug, proving the source-side change is what matters, not just the recipe flag.
- `androidStudioPreviewRecipesBothOptIntoEntryScopedResolution` — registry-derived, loops every non-stable Android Studio recipe rather than hand-listing canary/beta, so a future third preview channel is covered automatically.
- `highestVersionEntryFallsBackWhenThereIsNothingToDisambiguate` — unit coverage of the primitive's two fallback cases.
- `androidStudioChannelProofsSpanTheWholeStabilityFloor` (pre-existing) still passes unmodified — it asserts `recipe.install?.urlSource` is still `.bodyPattern`, which this fix does not change.

## Verification (raw output)

Full suite:

```
make test
```
```
✔ Test run with 1164 tests in 92 suites passed after 39.736 seconds with 2 known issues.
   (the 2 known issues are pre-existing tracked-flaky tests unrelated to this change:
    keepassxcRespinIsTheAssetSelected, ankiZeroPaddedTagComparesEqualToInstalled)
✔ Test run with 125 tests in 19 suites passed after 0.067 seconds.   [CLI package]
✓ localizable keys agree — 411 in the catalog, all reachable
```

(An earlier full run hit two total-misses on `com.google.android.studio [beta]` under the suite's full concurrent fan-out — re-ran that test alone and it passed cleanly in 7-10s each time; a second full `make test` run came back fully green with no android.studio issue at all. Network contention from ~150 concurrent live probes, not a code regression — the test file's own comments already document this exact failure shape for a differently-contended run.)

Per this repo's convention, a full `duo verify` sweep is ~150 requests / ~3 minutes and shouldn't be declared green off a partial run — so I did **not** run the full sweep, only the changed recipes:

```
swift run --package-path CLI duo verify --only android.studio
```
```
  duo verify
  3 vendor probes  0 GitHub rules  0 changelogs
  vendor probe  ✓ 3  ⚠ 0  ✗ 0  ~ 0  - 0
```

And fetched the live feed once directly to confirm the resolved versions against the real 671-item body (via `--report`/`--baseline`, which records the resolved version per recipe):

```json
"vendor:com.google.android.studio:canary": { "lastGoodVersion": "2026.2.1 Canary 2", ... }
"vendor:com.google.android.studio:beta":   { "lastGoodVersion": "2026.1.4 RC 2", ... }
"vendor:com.google.android.studio:stable": { "lastGoodVersion": "2026.1.3", ... }
```

Canary now resolves the newer 262 train's Canary 2 (previously would have resolved the stale 261 RC per the issue). Beta correctly stays on the RC — it doesn't accept Canary builds, so it was never exposed to this bug in the first place.

Fixes #76.

### DIFF
--- a/DuoUpdaterCore/Sources/DuoUpdaterCore/Sources/ChannelArtifactProof.swift
+++ b/DuoUpdaterCore/Sources/DuoUpdaterCore/Sources/ChannelArtifactProof.swift
@@ -63,11 +63,17 @@ public enum ChannelProofRegistry {
         // MORE STABLE (the stability floor documented on the recipes), so the marker
         // has to be that whole ladder, not just the channel's own name — Canary
         // resolves the newest of {Canary, Beta, RC}, Beta the newest of {Beta, RC}.
-        // That is not a bug being papered over: it is why a Canary install correctly
-        // moves onto `2026.1.4 RC 2` when it is newer than any open Canary, which is
-        // exactly what the feed served on 2026-08-26. Google's Beta train has in
-        // practice shipped RELEASE CANDIDATES for years (no `Beta` item since
-        // 2025-03-18), so `-rc<N>-` is the marker actually seen on both.
+        // That is not a bug being papered over: an RC genuinely is the legitimate
+        // answer for a Canary install once it is the highest build on the ladder —
+        // e.g. before a newer feature version's Canary train has opened. (It is NOT
+        // legitimate merely because the RC was the most recently PUBLISHED item —
+        // the feed is ordered by publish date, not by version, which is what made
+        // the canary recipe land on `2026.1.4 RC 2` over the already-published,
+        // already-newer `2026.2.1 Canary 2` on 2026-08-26; see issue #76 and
+        // `VendorProbeRecipe.entryStartPattern`, which now resolves that correctly.)
+        // Google's Beta train has in practice shipped RELEASE CANDIDATES for years
+        // (no `Beta` item since 2025-03-18), so `-rc<N>-` is the marker actually
+        // seen on both.
         // What the marker still excludes is the pair that WOULD be a cross-channel
         // install: `Release` (`android-studio-quail3-mac_arm.dmg`) and `Patch`
         // (`…-patch1-mac_arm.dmg`) — neither carries a ladder token.
--- a/DuoUpdaterCore/Sources/DuoUpdaterCore/Sources/VendorProbeDiagnostics.swift
+++ b/DuoUpdaterCore/Sources/DuoUpdaterCore/Sources/VendorProbeDiagnostics.swift
@@ -150,7 +150,11 @@ public struct ProbeOutcome: Sendable {
     public let httpStatus: Int?
     /// The text the version pattern was run against, capped at
     /// ``ProbeOutcome/maxSampleBytes``. This is what a human (or a triage step)
-    /// needs to see to fix a pattern.
+    /// needs to see to fix a pattern. For a recipe with `entryStartPattern` set,
+    /// this is the ONE winning entry `VendorProbeSource` scoped everything else
+    /// to — not the whole fetched body — so it matches what the recipe actually
+    /// resolved against, not just whatever fits in the first
+    /// ``ProbeOutcome/maxSampleBytes`` of the feed.
     public let bodySample: String?
     public let elapsedMs: Int
 
--- a/DuoUpdaterCore/Sources/DuoUpdaterCore/Sources/VendorProbeRecipe.swift
+++ b/DuoUpdaterCore/Sources/DuoUpdaterCore/Sources/VendorProbeRecipe.swift
@@ -273,6 +273,35 @@ public struct VendorProbeRecipe: Sendable {
     /// probed text is a URL or a single plist value rather than a document.
     public let publishedAtPattern: String?
 
+    /// Optional regex marking where each entry begins in a body that lists
+    /// several releases — e.g. `\{"date":"` for a JSON feed whose items each
+    /// start with a `date` key. When set, the source slices the body into
+    /// entries at every match (one match's start to the next match's start,
+    /// last entry running to the end of the body), keeps the entries where
+    /// `versionPattern` matches, and picks the one whose extracted version
+    /// compares highest (the same `VersionComparator` ordering
+    /// `highestVersion`/`highestVersionedURL` use). `versionPattern`,
+    /// `displayVersionPattern`, `publishedAtPattern`, and the install spec's
+    /// `.bodyPattern`/`.bodyPatternRelative`/`.bodyTemplate` URL are then all
+    /// resolved against that ONE winning entry, so they can never land on
+    /// different releases — see `VendorProbeRecipe.highestVersionEntry`.
+    ///
+    /// This exists because a feed ordered by *publication date* rather than by
+    /// *version* breaks every first-match pattern above at once whenever two
+    /// release trains are open simultaneously (Android Studio: a newer feature
+    /// version's Canary can be published before an older version's RC, so the
+    /// RC — not the newer Canary — sits first in the feed). Flipping
+    /// `selectHighest` alone does not fix this: it would pick the version by
+    /// comparison while `displayVersionPattern` and the install URL stayed
+    /// first-match, landing on three different releases' worth of data.
+    ///
+    /// Nil (the default) leaves every pattern reading the whole body,
+    /// first-match, exactly as before this field existed. Also the fallback
+    /// when the pattern matches fewer than two entries, or when no entry's
+    /// `versionPattern` matches — better a possibly-stale first-match answer
+    /// than no answer at all.
+    public let entryStartPattern: String?
+
     /// When present, the app can be updated in place through its own channel: the
     /// source resolves the installer URL (and optional checksum) and hands it to
     /// `VendorInstaller`. Absent → detection only (the user is sent to download
@@ -337,6 +366,7 @@ public struct VendorProbeRecipe: Sendable {
         versionIsBuild: Bool = false,
         displayVersionPattern: String? = nil,
         publishedAtPattern: String? = nil,
+        entryStartPattern: String? = nil,
         install: VendorInstallSpec? = nil,
         requestBody: RequestBody? = nil,
         requestHeaders: [String: String] = [:],
@@ -360,6 +390,7 @@ public struct VendorProbeRecipe: Sendable {
         self.versionIsBuild = versionIsBuild
         self.displayVersionPattern = displayVersionPattern
         self.publishedAtPattern = publishedAtPattern
+        self.entryStartPattern = entryStartPattern
         self.install = install
         self.requestBody = requestBody
         self.requestHeaders = requestHeaders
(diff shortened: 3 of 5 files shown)


jizhi0v0:
Addressed the adversarial review findings, all verified against the real code / live feed before implementing (offsets, ASCII-only feed, `555`/`42`/`198` stats — all reproduced independently).

**Must-fix (A–F):**
- **A** — `trackVersion` (`rolloutTrackVerdict`) now goes through the same `scopedBody` helper `probeOutcome` uses (factored into one shared choke point), so its "a divergence means the tracks differ, not that two code paths read the body differently" doc stays true even if a future recipe combines `track` + `entryStartPattern`.
- **B** — added `entryStartPatternsInTheRegistryAreValidRegexes`, a registry-wide (not just Android Studio) test that every `entryStartPattern` actually compiles — closes the silent-typo-forever hole.
- **C** — `highestVersionEntry` now declines (falls back to whole-body) when the winning entry's own `versionPattern` matches more than once, defending the "one entry, one release" claim against a future feed nesting the start marker mid-item. Tested directly, plus a clean control case.
- **D** — moved `ProbeOutcome.sample(...)` to after `scope` is computed and sample `scope`, not `body.text` — fixes the diagnostic sample and, traced through `Finding.bodySample`, `duo triage`'s pattern-validation step in `Triage.swift:343`. Confirmed the failure-path reasoning: `scope` only ever differs from `body.text` when extraction already succeeded against it, so the miss branch's sample is unaffected. Updated `bodySample`'s doc.
- **E** — added `VendorProbeRecipe.with(entryStartPattern:)` / `.with(url:)`, backed by one private `copy(...)` listing every field once, and switched the A/B test to use them — removes the hazard (16-of-19-field hand copy) rather than patching that one instance.
- **F** — `highestVersionEntry` now takes `selectHighest` and scores entries with the matching extractor, so entry-selection and the final reported version can't disagree about what "highest" means. Reproduced your exact 1.0/9.0-vs-2.0 scenario as a test.

**Doc-only (G–I):** expanded `entryStartPattern`'s doc with the `.bodyPatternLast`/`.bodyPatternHighestVersioned` collapse trap, the `^`/`$`/lookbehind re-anchoring warning (verified: 6 registry recipes pin `^…$` today), and the unbounded-history/no-self-correction note (verified against the live feed: 555 matching items, 42 two-segment 2018-era build ids, 198 missing the `displayVersionPattern` separator).

**Did not touch:** `GitHubReleasesSource`, `ChannelArtifactProof`'s `.recipeAnchor` surface, no new `ProbeWarning` case — per the "do not do" list.

**Verification (raw):**

```
make test
```
```
✔ Test run with 1167 tests in 92 suites passed after 70.703 seconds with 2 known issues.
   (pre-existing tracked-flaky, unrelated: keepassxcRespinIsTheAssetSelected, ankiZeroPaddedTagComparesEqualToInstalled)
✔ Test run with 125 tests in 19 suites passed after 0.072 seconds.
✓ localizable keys agree — 411 in the catalog, all reachable
```

```
swift run --package-path CLI duo verify --only android.studio
```
```
vendor probe  ✓ 3  ⚠ 0  ✗ 0  ~ 0  - 0
```
Live-resolved (`--report`): `canary → "2026.2.1 Canary 2"`, `beta → "2026.1.4 RC 2"`, `stable → "2026.1.3"` — canary is still correctly picking the newer 262 train.

Pushed as 3 new commits on `fix/android-studio-canary-feed-order-76` (`596c55c`, `28a36f3`, `217050d`). Not merging.

Issue #81, opened 2026-08-27
ChannelArtifactProof .recipeAnchor hand-enumerates recipe fields, so new fields silently shrink the guard


jizhi0v0:
`RecipeSanity.crossChannelArtifact`'s `.recipeAnchor` check builds the text it searches by **hand-listing** recipe fields (`ChannelArtifactProof.swift:213-220`):

```swift
recipe.url … recipe.versionPattern … recipe.install?.urlSource
```

`entryStartPattern` (added in #78) is not in that list, and neither will be the next field anyone adds.

## Why this is worth fixing rather than patching

The repo already argues against this exact shape, in `VendorProbeRecipe.localReads` (`VendorProbeRecipe.swift:197-200`):

> Splitting the plan out of `identities` broke two such guards the day it happened, which is the argument for deriving rather than enumerating.

This is the same trap in the same file family, and it is the closest thing the codebase has to a digest over recipe fields. Adding a field silently shrinks what the guard inspects — the guard keeps passing while covering less, which is the worst failure mode for a guard.

## Impact today

**None.** No recipe currently combines a `.recipeAnchor` proof with an `entryStartPattern`. This is filed as a latent structural issue, not a live defect.

## Proposal

Derive the anchor surface from the recipe rather than enumerating it, so a new field is included by construction. If deriving is genuinely impractical here, the fallback is a `Mirror(reflecting:).children.count` assertion in the tests, so adding a field fails loudly and forces the author to decide whether the guard should see it — the same alarm suggested for the A/B recipe copy in #78 (solved there by adding `VendorProbeRecipe.with(...)` copy methods).

Deliberately left out of #78: touching the proof surface is unrelated to the feed-ordering fix and belongs in its own change.

Found during the adversarial review of #78.

PR #122, opened 2026-08-28
fix(core): a .recipeAnchor must name the fields it relies on, and match in each (#110)

### LINKED ISSUE #110: A .recipeAnchor passes if any field matches, so an anchor meant to guard the install spec can be satisfied by the version pattern
Found by the pre-release review of the v0.3.67→0.3.68 range. Not live today — a next-edit hazard.

## The guard

`com.tencent.wechatdevtools` `.rc` is proven by an anchor, not by the artifact, and the file says why (`ChannelArtifactProof.swift:130-133`): RC's artifact is byte-for-byte shaped like Stable's — same host, same directory, same filename template, differing only in the version. So this anchor is the **only** thing standing between an RC user and a Stable build.

```swift
ChannelProofKey("com.tencent.wechatdevtools", .rc): .recipeAnchor(#""id":.*"rc""#),
```

## Why it can't fail the way it should

`.recipeAnchor` is matched against `channelAnchorSurface`, which joins every non-labelling field one per line and passes if **any** line matches. Today `"id":\s*"rc"` appears in *both* the `versionPattern` and the install `bodyPattern`.

That redundancy is what makes it weak: the install pattern is the half that picks the artifact, and if it alone drifts — the vendor renames the block, someone rewrites that regex — the version pattern still satisfies the anchor, the proof stays green, and the install falls back to whichever channel `config.json` lists first. Which is Stable.

The guard would be reporting on a field that no longer decides anything.

## Verified
- The three WeChat recipes each carry the token in both fields (read them).
- `channelAnchorSurface` joins per-field and `crossChannelArtifact` uses a plain `range(of:)` over the join — one matching line is enough (`VendorProbeRecipe.swift`, `ChannelArtifactProof.swift`).
- Not currently misresolving: `duo verify` is green on these.

## NOT verified
- Whether a per-field anchor is the right shape, or whether RC would be better served by an `.artifact` proof if the vendor's URL ever gains a token. Worth deciding rather than assuming the first option.

## Suggested shape
Let a `.recipeAnchor` say *which* field it must appear in, so an anchor whose job is to prove the install spec is checked against the install spec. That also makes the existing entries state what they actually rely on. `ChannelProofRegistry` has five anchors today, so the migration is small.

jizhi0v0:
Closes #110.

## The defect, as measured

`.recipeAnchor` was matched against `channelAnchorSurface` — every anchorable field joined one per line — and passed if **any** line matched.

WeChat DevTools `.rc` is proven by an anchor rather than by its artifact because its RC build is byte-for-byte shaped like Stable's (same host, same directory, same filename template). That anchor is the only thing between an RC user and a Stable build. `"id": "rc"` appears in **both** `versionPattern` and the install `bodyPattern`, so rewriting the install regex left the version pattern satisfying the anchor while the install fell back to whichever channel `config.json` lists first. Which is Stable.

## Where this differs from the issue as filed

The issue proposed checking "an anchor whose job is to prove the install spec against the install spec". I measured all five registered anchors before implementing, and that rule cannot be applied uniformly:

| proof | pattern | fields that actually carry it |
|---|---|---|
| `com.jetbrains.intellij-EAP` `.preview` | `type=eap` | **`url` only** |
| `com.runningwithcrayons.Alfred` `.beta` | `prerelease\.xml` | **`url` only** |
| `com.tencent.wechatdevtools` `.rc` | `"id":.*"rc"` | `versionPattern` + `install` |
| `dev.kdrag0n.MacVirt` `.beta` / `.canary` | `<sparkle:channel>…` | `versionPattern` + `install` |

IntelliJ EAP and Alfred beta have **no channel token in their install spec at all** — their endpoints return only that channel's builds, so their *install* patterns are byte-identical to their stable siblings' and neither their install nor their version pattern names a channel. Requiring an install-spec anchor would have failed the two recipes that are correctly built.

(Corrected after review: an earlier revision of this description said the *version* patterns were byte-identical too. They are not — IntelliJ EAP reads `"build"` where stable reads `"version"`, deliberately, because `versionIsBuild` compares the build; Alfred beta's differs in strictness. Neither difference names a channel, which is the property the scope actually rests on.)

So the shape here is: **the proof names the fields its channel identity lives in, and the regex must match in every one of them.** Endpoint-keyed recipes name `["url"]`; shared-feed recipes name `["versionPattern", "install"]`.

## Also closed

Both degenerate anchors are now findings rather than passes — an empty field set, and a name that is not an anchorable field (typo, rename, or one of the labelling `nonAnchorFields`). Each would be a proof that cannot fail for any input, which reads as green forever. `everyRegisteredAnchorNamesRealFields` catches them in a PR; `recipeAnchorFailure` catches them at runtime.

Both `crossChannelArtifact` overloads share one matcher, so the two registries cannot drift into different notions of what an anchor asserts.

The fields remain **derived by reflection** — a proof may name any field the recipe reads, including one added after this lands. That keeps the `entryStartPattern` property from #81, which now has its own test saying so explicitly.

## Verification

`swift test` — DuoUpdaterCore:
```
Test run with 1317 tests in 103 suites passed after 46.233 seconds with 1 known issue.
```
(the known issue is the pre-existing `GitHubReleaseRuleTests` xfail)

`swift test --package-path CLI`:
```
Test run with 161 tests in 22 suites passed after 0.075 seconds.
```

Live endpoints, with a `duo` built from **this branch** (the installed binary carries the old registry):
```
wechatdevtools       total  ✓ 6  ⚠ 0  ✗ 0
intellij-EAP         total  ✓ 1  ⚠ 0  ✗ 0
Alfred               total  ✓ 2  ⚠ 0  ✗ 0
MacVirt              total  ✓ 4  ⚠ 0  ✗ 0
```

### Red first, then the counterfactual

A green sweep proves nothing on its own, so I drifted the WeChat RC install pattern — dropped its `"id": "rc"` block, left `versionPattern` intact — and ran the live sweep on **this branch**:

```
⚠ WARN  vendor:com.tencent.wechatdevtools:rc
    warning   the recipe's install is no longer anchored on /"id":.*"rc"/ — nothing there ties its rc install to that channel
```

Then applied the **same drift to `main`'s code** and ran the same sweep:

```
vendor probe  ✓ 3  ⚠ 0  ✗ 0  ~ 0  - 0
```

Silent. That is the defect, reproduced end to end and then closed.

No recipe content changed — `git diff main -- VendorProbeRecipe.swift` touches only the anchor-surface block.

---

## Review round 2

An adversarial review of the first pass found a defect in the comments this PR itself added, plus three gaps. All fixed in `6be9…`:

- **DEFECT — the "byte-identical" claim was false**, in three comments and in this description. Measured: IntelliJ EAP `"build"` vs stable `"version"`; Alfred beta `([0-9]+(?:\.[0-9]+)+)` vs stable `([0-9][0-9.]*)`. The conclusion held, the evidence did not. Narrowed to "neither pattern names a channel", which is what is true and what the scope rests on. In a repo where comments are specification this is a defect, not a nit — the next reader trusts "byte-identical" and normalises the two IntelliJ recipes.
- **A third degenerate anchor was still a silent pass.** The description claimed "both" were closed; there are three. A *pattern* matching any text (`""`, `a?`, `.*`) satisfies every field of every recipe forever. `noRegisteredAnchorMatchesEverything` now tests both registries against the empty string. Verified red-first — swapping IntelliJ's anchor to `.*` fails it with the right message.
- **`recipeAnchorFailure` took three hand-paired strings** (`subject`, type name, surface closure) that had to agree, in the function whose stated purpose is preventing exactly that drift. Replaced with a `ChannelAnchorSubject` protocol, so the type supplies all three.
- **Granularity limit now stated:** `install` is anchored as a whole field, so #110's shape survives inside a `.bodyTemplate`'s `fields`. Not reachable today (no anchored recipe uses `bodyTemplate`); documented rather than left to be rediscovered.
- The `entryStartPattern` negative case now drifts a **present** field rather than only nilling it — a nil field's surface is the string `"nil"`, so the old form would have passed whether or not the field was checked. The GitHub half of the field-name check no longer skips silently when a proof has no rule.
- `docs/app-audits/com-tencent-wechatdevtools.md` records the RC proof's new field scope.

Re-verified after the fixes: `swift test` 1318 passing (DuoUpdaterCore) + 161 (CLI), and the four anchored recipes green against live endpoints.

### DIFF
--- a/DuoUpdaterCore/Sources/DuoUpdaterCore/Sources/ChannelArtifactProof.swift
+++ b/DuoUpdaterCore/Sources/DuoUpdaterCore/Sources/ChannelArtifactProof.swift
@@ -24,9 +24,33 @@ public enum ChannelArtifactProof: Sendable, Hashable {
     /// The vendor ships ONE artifact to several channels, so the URL carries no
     /// channel token and there is nothing to assert on it. The proof is instead
     /// that the recipe reads a channel-dedicated endpoint (or a channel-tagged
-    /// block of a shared feed): the regex must appear in the recipe's endpoint
-    /// URL, version pattern, or install URL source.
-    case recipeAnchor(String)
+    /// block of a shared feed).
+    ///
+    /// `fields` names the recipe fields the channel identity actually lives in
+    /// (their `Mirror` labels — `"url"`, `"versionPattern"`, `"install"`, …),
+    /// and the regex must match in EVERY one of them. That is the whole point:
+    /// matching against the joined surface passes if any single line matches, so
+    /// a token that happens to sit in two fields kept the proof green while
+    /// either one drifted. WeChat DevTools RC is the live case — `"id": "rc"` is
+    /// in both `versionPattern` and the install `bodyPattern`, and only the
+    /// install half picks the artifact (issue #110).
+    ///
+    /// Which fields to name is a per-recipe fact, not a rule to apply uniformly.
+    /// Measured across the five anchors registered on 2026-08-28:
+    ///   * **Endpoint-keyed** (IntelliJ EAP, Alfred beta) — the response body
+    ///     holds only that channel's builds, so `versionPattern` and the install
+    ///     pattern are byte-identical to their stable siblings' and contain no
+    ///     channel token to anchor on. `["url"]` is the whole proof, and naming
+    ///     anything else would be an assertion those recipes cannot satisfy.
+    ///   * **Shared-feed** (WeChat DevTools RC, OrbStack beta/canary) — one
+    ///     endpoint serves every channel, and the per-field patterns are what
+    ///     select. Both the version half and the install half must stay anchored,
+    ///     so both are named.
+    ///
+    /// An empty set, or a name that is not an anchorable field of the recipe, is
+    /// a hard finding rather than a pass — either would be a proof that cannot
+    /// fail. See `RecipeSanity.recipeAnchorFailure`.
+    case recipeAnchor(String, in: Set<String>)
 }
 
 /// A `(bundleID, channel)` pair — the same key `VendorProbeSource` selects a
@@ -58,7 +82,17 @@ public enum ChannelProofRegistry {
         // uses, so the filename proves nothing. The endpoint does: `type=eap` vs
         // stable's `type=release`, and the install pattern reads the dmg out of
         // that response body.
-        ChannelProofKey("com.jetbrains.intellij-EAP", .preview): .recipeAnchor(#"type=eap"#),
+        //
+        // `url` alone, and that is the honest scope rather than a weaker one: the
+        // response to `type=eap` contains ONLY EAP builds, so `versionPattern`
+        // (`"build": "…"`) and the install pattern (`"macM1" … "link"`) carry no
+        // channel token at all — they are what a `type=release` recipe would use
+        // verbatim. Naming them would assert something this recipe cannot satisfy.
+        // What naming `url` buys is that swapping the endpoint back to
+        // `type=release` fails the proof instead of passing on patterns that never
+        // mentioned a channel.
+        ChannelProofKey("com.jetbrains.intellij-EAP", .preview):
+            .recipeAnchor(#"type=eap"#, in: ["url"]),
         // Android Studio: each preview channel accepts builds at its own quality OR
         // MORE STABLE (the stability floor documented on the recipes), so the marker
         // has to be that whole ladder, not just the channel's own name — Canary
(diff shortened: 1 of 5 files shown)


jizhi0v0:
Note for merge ordering: the adversarial re-review found that `noRegisteredAnchorMatchesEverything` (added here) has a hole — it probes only the empty string, so `.+` and `.` pass it while asserting nothing beyond "this field is non-empty". The fix lives in the stacked #123 rather than here, because the strengthened test also iterates `bindingProofs`, which only exists on that branch. If this PR merges alone, that gap stands until #123 lands. Everything else the review raised against this PR is fixed in `d9e6e0d`.

PR #124, opened 2026-08-28
fix(install): preserve elevated bundle directory modes

jizhi0v0:
## Summary

- preserve the installed app bundle and Contents directory modes during an elevated in-place swap
- apply those modes to the staged sibling before it occupies the registered app path
- keep the existing atomic rename and ownership-preservation behavior unchanged
- add a regression test that runs the complete privileged shell transaction in a temporary directory

## Why

While investigating the withdrawn input-method one-click path, I reproduced an independent bug in the generic privileged swap.

Both locally installed third-party input methods use the same filesystem shape:

- WeType.app: root:staff 775 for both the bundle and Contents
- DoubaoIme.app: root:staff 775 for both the bundle and Contents

Their own updaters keep the outer app and stage a replacement Contents directory inside it. Official downloaded archives, however, unpack as 755. The elevated DuoUpdater path copied the archive with ditto and restored ownership only, so the live install silently changed from 775 to 755. For WeType that removes the group-write bit its next self-update needs to create the staged Contents directory.

Repairing the mode after the live rename is not reliable. In the controlled reproduction, macOS attached com.apple.macl to the registered app path and a later chmod returned EPERM even with administrator privileges.

This change reads the authoritative modes from the existing install while it is still present, applies them to the staged sibling, and only then starts the two-rename exchange. A chmod failure therefore leaves the original app untouched, and an interrupted staged copy remains user-owned and recoverable.

The input-method policy remains unchanged: this PR does not re-enable in-place installation for WeType, DoubaoIme, or the input-method class.

## Validation

- controlled same-version WeType replacement with configuration backup and post-swap signature/configuration checks
- verified WeType and DoubaoIme both use root:staff 775 and Contents staging
- new PrivilegedReplacementShellTests regression test
- full DuoUpdaterCore test suite
- CLI test suite: 161 tests

### DIFF
--- a/DuoUpdaterCore/Sources/DuoUpdaterCore/Install/InPlaceSwap.swift
+++ b/DuoUpdaterCore/Sources/DuoUpdaterCore/Install/InPlaceSwap.swift
@@ -318,6 +318,46 @@ public enum InPlaceSwap {
     }
 
     private static func privilegedReplace(newApp: URL, target: URL) throws {
+        let shell = try privilegedReplacementShell(newApp: newApp, target: target)
+        let appleScript = "do shell script \"\(escapeForAppleScript(shell))\" with administrator privileges"
+
+        // One password panel at a time. The apply pool allows two concurrent
+        // swaps, and now that every root-owned bundle takes this path (28 apps in
+        // `/Applications` on the development machine, 22 of them store-installed)
+        // a batch can reach it twice at once — two system panels stacked over each
+        // other, neither saying which app it belongs to. Blocking here rather than
+        // making `replace` async is deliberate: the callers are synchronous, and
+        // the thread this parks was going to sit in `waitUntilExit` waiting on the
+        // same human anyway, so this moves the wait rather than adding one.
+        elevationPanel.lock()
+        defer { elevationPanel.unlock() }
+
+        let process = Process()
+        process.executableURL = URL(fileURLWithPath: "/usr/bin/osascript")
+        process.arguments = ["-e", appleScript]
+        let errPipe = Pipe()
+        process.standardError = errPipe
+        try process.run()
+        let errData = errPipe.fileHandleForReading.readDataToEndOfFile()
+        process.waitUntilExit()
+        guard process.terminationStatus == 0 else {
+            let msg = String(data: errData, encoding: .utf8) ?? "unknown error"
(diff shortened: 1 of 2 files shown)


jizhi0v0:
Superseded by #126 — closing this one rather than merging is the cleaner path.

#126 contains this commit unchanged as its first commit, plus a follow-up correcting two things about it that only showed up once it was measured against real bundles:

**Best-effort instead of fatal.** `directoryMode` threw, so a `stat` that came back empty refused the whole elevated swap. That path is taken by *every* root-owned app (28 on the development machine), and the ownership block right beside it is deliberately non-fatal for exactly this reason — refusing to install a verified build over a metadata detail is the worse trade.

**Two levels instead of recursive.** The comment here justifies mode preservation from WeType and DoubaoIme needing the group-write bit. That turned out to be true, and *more* true than stated: both vendors' installers end with `chmod -R 775`, recursively, and it is load-bearing at the **end** of their next update — after rotating `Contents` their updater has to delete the copy it displaced, and unlinking a tree needs write permission on every directory inside it (WeType's updater carries `cleanup old Contents failed:` for when it can't).

But those apps no longer take this path at all in #126 — they rotate `Contents` instead — so carrying the bit recursively *here* would apply an input-method-derived rule to everything else. Measured across 137 bundles on this machine: 5 have a group-writable root, and 2 of those take the privileged path (Microsoft Word and Excel, `root:wheel` 775). Their entire interiors would have been opened to the group on the next update, for no stated reason. The recursion moved into `rotateContents`, where it was measured; this path now guarantees exactly two levels, in both directions, with a test pinning the limit as well as the rule.

One more thing worth recording here, since it is the reason the input-method work needed a different shape than a bundle swap at all: writing *inside* another app's bundle is blocked by App Management even for root. One root shell, one directory:

```
uid 0  ditto → /Library/Input Methods/.probe                 status 0
uid 0  ditto → /Library/Input Methods/DoubaoIme.app/.probe   Operation not permitted
```

The whole-bundle path this PR touches is unaffected — everything it writes and renames sits *beside* a bundle, never within one.

jizhi0v0:
Closing rather than merging: #126 contains this commit unchanged as its first commit, plus the two follow-up corrections described above. Merging here would replay it.

PR #165, opened 2026-08-30
docs(audit): Supacode — Sparkle covers both channels, verified on real builds

jizhi0v0:
## What

[Supacode](https://supacode.sh/) — native terminal for coding agents. Homebrew 90d installs 6,658, cask `auto_updates`. Its dmg carries `SUFeedURL` (`supacode.sh/download/latest/appcast.xml`), so the **generic Sparkle source already covers it — zero recipe**. The deliverable here is the verified audit doc, the same shape as the earlier Sparkle-covered apps (Superwhisper, OpenClaw, Dia Browser).

## Why it is not just "has a feed"

The feed is a channel-tagged one: 21 items, 10 default + 11 `<sparkle:channel>tip</sparkle:channel>` (unstable history builds), and `pubDate` is not reliably ordered. Two traps verified against real artifacts:

- **Both tracks report the same short version** (default and tip both say `0.10.8`); only `CFBundleVersion` moves (timestamp builds `1785775286` vs `1787740786`). The feed's `sparkle:version` equals that build and `comparisonKey` prefers it, so updates are detected on the build — no change needed, but worth pinning so nobody "fixes" it into a marketing-only comparison (the Amp/Surge trap).
- **Tip users stay on tip** via the existing build-matching inference (`allowedChannels` → `channel(ofInstalled:)`), not a new ChannelBinding. A default install never sees tip items.

## Verification (real artifacts)

- Default dmg + feed enclosure zip: `app.supabit.supacode`, build `1785775286`, Team `9ZLSJ2GN2B`, notarized.
- Tip zip: same bundle, build `1787740786`, same Team, notarized.
- `channel-verify --check app.supabit.supacode` on the default build: **Sparkle wins, up to date**; on the tip build: still up to date, no cross-channel push.
- `check_app_audits.py` green.

No code changes — docs only.

### DIFF
--- a/docs/app-audits/README.md
+++ b/docs/app-audits/README.md
@@ -147,6 +147,7 @@ Per-app audit checklist. Run `/app-audit <App>` for each, then check off.
 - [x] [**Superwhisper**](com-superduper-superwhisper.md) · `com.superduper.superwhisper` — S · real zip/feed verified ✓ · 2026-08-17
 - [x] [**Perplexity Mac**](ai-perplexity-macv3.md) · `ai.perplexity.macv3` — S detection; current dmg unnotarized, no one-click · 2026-08-17
 - [x] [**Dia Browser**](company-thebrowser-dia.md) · `company.thebrowser.dia` — S · real DMG/feed verified ✓ · 2026-08-17
+- [x] [**Supacode**](app-supabit-supacode.md) · `app.supabit.supacode` — S · default+tip 两轨真包验证 ✓（tip 经 build 反查推断，零 recipe）· 2026-08-30
 
 ## Investigated — blocked safely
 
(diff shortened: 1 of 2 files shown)


jizhi0v0:
Blocked after real-feed verification: Supacode stable and tip share short version 0.10.8. The current channel inference examines each item with build-or-short matching in one pass, so when a tip item appears before the exact stable-build item, a stable install matches the tip item by short version and unlocks the tip channel. Reproduction: installed 0.10.8/1785775286 against the current feed selects tip build 1788000000 instead of staying on default. This PR's zero-core-change channel claim is therefore not safe to merge. The core fix should perform a complete build-number pass before any short-version fallback, with an offline regression fixture for this exact feed shape.

PR #207, opened 2026-09-01
fix: an app is Tauri when the binary says so, not when the packaging resembles it (#206)

### LINKED ISSUE #206: Tauri detection has a false positive: a wry-only app (Longbridge) is labelled Tauri
`AppRuntimeDetector` labels an app **Tauri** from a packaging fingerprint plus a WebKit link:

```
Info.plist has LSRequiresCarbon == true && CSResourcesFileMapped == true   // cargo-bundle / tauri-bundler
&& the executable links WebKit
```

That is a heuristic, and it has a real false positive on this machine.

## Measured

Five installed apps match the fingerprint. Four are Tauri; one is not:

| app | `tauri-*` crate in the binary | verdict |
|---|---|---|
| Conductor | `tauri-2.11.1` | Tauri |
| Clash Verge | `tauri-2.11.5` | Tauri |
| CC Switch | `tauri-2.10.3` | Tauri |
| Handy | `tauri-2.11.5` | Tauri |
| **Longbridge** | **none — only `wry-0.53.3`** | **not Tauri** |

`wry` is Tauri's WebView layer and is perfectly usable on its own, which is what Longbridge does. It is cargo-bundled and links WebKit, so it is indistinguishable from Tauri by the current rule.

There is no cheaper marker. Comparing Longbridge against the four real Tauri apps, the `Contents/` layout and the `Info.plist` key set are the same shape — same `CSResourcesFileMapped` / `LSRequiresCarbon` / `LSMinimumSystemVersion` / `CFBundleURLTypes`, same `Resources/` with an icon and nothing distinctive.

## Current behaviour

`RuntimeVersion.read(.tauri, …)` already scans the executable for the `tauri-<semver>` crate path Cargo bakes into panic metadata, so **the detail popover is honest**: Longbridge gets no version, because there is no Tauri in there to version. What stays wrong is the **row's mark and label**, which still say Tauri.

## Why it wasn't fixed with the rest

The proof is only available by walking the executable — 289 ms for Longbridge's 261 MB — which cannot go on the scan path that runs over every app in the library. The on-demand reader now caches its answer keyed on `path | app version | runtime`, so the cost is paid once per app per update.

## Possible fixes

1. **Consult the same cache from the scanner.** Once a detail has been opened, the crate answer is already known; a later scan could downgrade the label for an app the cache says has no Tauri. Only helps after the user has looked at that app once.
2. **Persist the cache** (defaults, keyed the same way) and let the scanner read it. Then the correction survives relaunches and the walk is paid once ever, per app version.
3. **Find a cheap positive marker for wry-without-Tauri.** Nothing found so far.
4. Accept it and hedge the popover copy for the unconfirmed case.

(2) looks like the right shape: the key already exists, and the failure direction stays safe — an app is only ever *demoted* out of Tauri by positive evidence that the crate is absent.

## Not a correctness risk elsewhere

Nothing routes on `runtime`: it is descriptive only, read by no source, policy or install path. The blast radius is one wrong word and one wrong logo on one row.

jizhi0v0:
Closes #206.

## What was wrong

Tauri bundles nothing distinctive — its frontend compiles into the binary and it draws into the system WebView — so `AppRuntimeDetector` inferred it from two cheap markers: the tauri-bundler plist pair (`LSRequiresCarbon` + `CSResourcesFileMapped`) and a WebKit link. The plist pair alone only says "cargo-bundle"; the WebKit link was there to separate Warp and Zed, which are cargo-bundled Rust apps with their own renderers.

The counterexample was already installed. **Longbridge draws with GPUI — the same renderer as Zed, pulled from the same repo — and embeds a renamed wry fork for incidental web content:**

```
~/.cargo/git/checkouts/zed-a70e2ad075855582/6ae5231
~/.cargo/git/checkouts/gpui-component-95ce574d8a0da8b8/20a1bb4   # github.com/longbridge/gpui-component
~/.cargo/registry/src/rsproxy.cn-…/lb-wry-0.53.3/src/wkwebview/…
```

`strings` on the 0.19.1 binary: `gpui` 2140, `zed` 1238, **`tauri` 0**. Cargo-bundled, links WebKit, is not Tauri. Zed and Warp escaped only by not happening to embed a web view — which the old comment predicted almost word for word: *"a future cargo-bundled app that embeds a WebView for something incidental would read as Tauri."*

## The fix

Keep the cheap markers as a **filter** and add the proof: the `tauri-<semver>` Cargo path Rust bakes into the binary. Positive evidence, and it yields the version besides.

`RuntimeVersion.carriesTauriCrate` routes through the same cache the detail view reads, so the walk happens once per app version rather than once per asker — which also closes the second half of #206 (the version shown in the popover is now already paid for by the scan).

## Cost

The worry that deferred this was scanning binaries during a library-wide scan. It doesn't apply, because almost nothing reaches the scan — on a 144-app library the filter admits six bundles:

| bundle | binary | crate found | warm |
|---|---|---|---|
| STCM Editor | 10 MB | `tauri-2.10.1` | 0.006s |
| Conductor | 65 MB | `tauri-2.11.1` | 0.039s |
| Longbridge | 262 MB | **none** — full walk to prove absence | 0.25s |
| CC Switch / Clash Verge / Handy | — | found | — |

Whole sweep of 144 bundles, **release build**: `0.50s`, and a repeat sweep `0.05s` (cache). Note that a debug build is ~200× slower on this path (`-Onone` byte loop, 104s for the same sweep) — the new tests use tiny stub executables, so `make test` is unaffected.

## Verification

```
make test
```

Core: `1771 tests in 139 suites passed ... with 1 known issue`. CLI: `184 tests in 24 suites passed`. All three scripts green (`check_localizable_keys`, `check_staged_version_use`, `check_app_audits`).

Real-library sweep, release build, comparing every bundle's verdict before and after:

```
SWEEP total=144
SWEEP changed=["Longbridge.app: tauri -> native"]
counts: electron 36, native 78, qt 10, tauri 5, chromium 3, flutter 3, java 3, catalyst 1, (none) 5
```

Exactly one classification changes, and it is the one that was wrong.

## Residual risk, stated plainly

A Tauri app whose binary keeps no `tauri-` crate path would now read as native. That is the safer direction — an app reading as itself-minus-a-label, rather than an unrelated app claiming to be Tauri — and nothing routes on this value; it is a label in the UI. Both Tauri versions observed here (v2.10, v2.11) carry the path, and the mechanism is Rust's, not Tauri's, so v1 should behave the same; I have no v1 app installed to confirm that, and have not asserted it in the code comments.

### DIFF
--- a/App/Sources/MenuContentView.swift
+++ b/App/Sources/MenuContentView.swift
@@ -900,7 +900,6 @@ private struct AppRow: View {
             // by its edges rather than its centre. See `RunningIndicator.opticalNudge`.
             if let runtime {
                 RuntimeTag(runtime: runtime, bundle: result.app.path,
-                           appVersion: result.app.shortVersion,
                            frameworks: result.app.linkedFrameworks)
             }
         }
--- a/App/Sources/Preferences.swift
+++ b/App/Sources/Preferences.swift
@@ -212,10 +212,16 @@ final class Preferences {
     /// Show a small chip beside each app's name naming what it is built with —
     /// Electron, Tauri, Qt, native, … See `AppRuntime`.
     ///
-    /// Default ON. The information is free (the scan already reads every bundle
-    /// this comes from) and it explains rows that otherwise look alike: an app the
+    /// Default ON, because it explains rows that otherwise look alike: an app the
     /// App Store manages already says so on the right-hand side, while everything
     /// else was indistinguishable.
+    ///
+    /// This gates the *display* only. The scan works the runtime out either way —
+    /// the workbench's own filters read it — so turning this off hides the mark
+    /// without saving the work behind it. Worth knowing because that work is no
+    /// longer nothing: since #206, proving an app is Tauri means reading its
+    /// executable through. It is confined to the bundles a cheap plist-and-link
+    /// filter admits, six of about a hundred and fifty here.
     var showRuntimeTags: Bool {
         didSet { defaults.set(showRuntimeTags, forKey: Key.showRuntimeTags) }
     }
--- a/App/Sources/RuntimeTag.swift
+++ b/App/Sources/RuntimeTag.swift
@@ -29,9 +29,6 @@ struct RuntimeTag: View {
     /// runtime's own version lives inside it. Nil where there is no real app behind
     /// the mark.
     var bundle: URL?
-    /// The app's own version, used only as the cache key for the runtime-version
-    /// read: an app that has not been updated cannot have changed its runtime.
-    var appVersion: String?
     /// What the app's binary actually links, shown in the click-through detail for
     /// the cases where the runtime label alone is not the whole answer. Empty is
     /// normal — nothing is claimed when nothing was read.
@@ -147,8 +144,7 @@ struct RuntimeTag: View {
             // the same cached entry back out.
             guard !versionLoaded, let bundle else { return }
             version = await Task.detached(priority: .utility) {
-                RuntimeVersion.read(runtime, bundleAt: bundle,
-                                    appVersion: appVersion, scanningBinaries: true)
+                RuntimeVersion.read(runtime, bundleAt: bundle, scanningBinaries: true)
             }.value
             versionLoaded = true
         }
--- a/App/Sources/RuntimeTag.swift
+++ b/App/Sources/RuntimeTag.swift
@@ -55,10 +55,10 @@ struct RuntimeTag: View {
 
     @State private var showingDetail = false
     @State private var version: String?
-    /// Separate from `version` being nil, which is a real answer: a runtime with no
-    /// trustworthy version, or a Tauri fingerprint with no Tauri crate behind it.
-    /// Without this the nil answer is never remembered and every re-open pays for
-    /// the search again — a quarter of a second, each time, on a large binary.
+    /// Separate from `version` being nil, which is a real answer: Chromium,
+    /// Flutter and the Apple-platform runtimes have no version worth printing, and
+    /// a rebranded Electron framework can hide the one it has. Without this the nil
+    /// answer is never remembered and every re-open pays for the search again.
     @State private var versionLoaded = false
 
     var body: some View {
@@ -139,10 +139,12 @@ struct RuntimeTag: View {
         .padding(14)
         .frame(width: 300, alignment: .leading)
         .task {
-            // Off the main thread and only once the detail is actually open: for
-            // Tauri this walks the whole executable, which is a quarter of a second
-            // on a large one. That cost is fine for a single app someone asked
-            // about and would be indefensible during a scan of the whole library.
+            // Off the main thread and only once the detail is actually open: a
+            // rebranded Electron framework is found by walking a binary, which is
+            // fine for a single app someone asked about and would be indefensible
+            // during a scan of the whole library. Tauri's walk is already paid for
+            // — the scan had to do it to reach the verdict at all, and this reads
+            // the same cached entry back out.
             guard !versionLoaded, let bundle else { return }
             version = await Task.detached(priority: .utility) {
                 RuntimeVersion.read(runtime, bundleAt: bundle,
--- a/App/Sources/RuntimeTag.swift
+++ b/App/Sources/RuntimeTag.swift
@@ -52,10 +52,13 @@ struct RuntimeTag: View {
 
     @State private var showingDetail = false
     @State private var version: String?
-    /// Separate from `version` being nil, which is a real answer: Chromium,
-    /// Flutter and the Apple-platform runtimes have no version worth printing, and
-    /// a rebranded Electron framework can hide the one it has. Without this the nil
-    /// answer is never remembered and every re-open pays for the search again.
+    /// Separate from `version` being nil, which is a real answer: Flutter and the
+    /// Apple-platform runtimes have no version worth printing, and a rebranded
+    /// Electron framework can hide the one it has. A Tauri mark can reach it too,
+    /// though it is now rare rather than ordinary — the binary was unreadable at
+    /// this moment, or it changed between the scan that produced the verdict and
+    /// this read. Without this the nil answer is never remembered and every re-open
+    /// pays for the search again.
     @State private var versionLoaded = false
 
     var body: some View {
--- a/App/Sources/WorkbenchWindowView.swift
+++ b/App/Sources/WorkbenchWindowView.swift
@@ -1108,7 +1108,6 @@ private struct DetailHeader: View {
                         // so they scale without losing their stroke ratio.
                         if model.prefs.showRuntimeTags, let runtime = result.app.runtime {
                             RuntimeTag(runtime: runtime, bundle: result.app.path,
-                                       appVersion: result.app.shortVersion,
                                        frameworks: result.app.linkedFrameworks, size: 18)
                         }
                     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ reads the section matching the version being shipped and embeds it in the GitHub
 release and the Sparkle appcast, so this file is the single source of truth for
 "what's new" — keep each version's prose written for users, not commit-speak.
 
+## 0.3.78
+
+**The mark that says what an app is built with now needs proof for Tauri, not a resemblance.** Tauri leaves nothing in a bundle to find — no framework, no folder of its own — so that one mark was worked out from how the app was packaged plus the fact that it links Apple's web view. Longbridge matches all of that and is not Tauri: it draws its own windows with the same renderer Zed uses, and embeds a web view for one corner of its interface. Duo Updater now reads Tauri's own fingerprint out of the binary before claiming it, so an app is called Tauri when it is one — and Longbridge reads as the native Mac app it is.
+
 ## 0.3.77
 
 **Every app in the list now says what it is built with.** A row for an App Store app has always carried the store's badge, and every other row looked alike — a Sparkle app, an Electron app and a native one were indistinguishable. Each name now carries the technology's own mark — click it for a word and a sentence, or hover for the tooltip: Electron, Tauri, Flutter, Qt, Java, Chromium, Mac Catalyst, an iPhone app on Apple silicon, or a native Mac app — with the runtime's version where that can be read as a fact (Electron 42.4.1, Qt 6.2, the Chromium an app embeds, the Tauri it was built with). It is read from the bundle itself — the framework a packager had to ship, the runtime a launcher needs, the libraries the binary links — so it is a fact about what is installed rather than a guess from the app's name. Where a long name leaves no room, the symbol steps aside rather than pushing the name onto a second line: the name is the row. Turn the whole thing off in Settings → General.
--- a/DuoUpdaterCore/Sources/DuoUpdaterCore/Models/InstalledApp.swift
+++ b/DuoUpdaterCore/Sources/DuoUpdaterCore/Models/InstalledApp.swift
@@ -172,9 +172,13 @@ public struct InstalledApp: Sendable, Identifiable, Hashable {
     /// binary, … Nil when the bundle carries no evidence either way, which the UI
     /// must render as "no badge" rather than as "native".
     ///
-    /// Purely descriptive: nothing in the update engine reads it. It is a fact
-    /// about the bundle, read at scan time from the same directory listing and
-    /// Info.plist the scan is already paying for. See `AppRuntimeDetector`.
+    /// Purely descriptive: nothing in the update engine reads it. Mostly it is a
+    /// fact read from the directory listing and Info.plist the scan is already
+    /// paying for — with one exception that stopped being free in #206: proving an
+    /// app is Tauri means reading its executable through, which the scan does for
+    /// the handful of bundles the cheap markers admit (six of about a hundred and
+    /// fifty here, half a second between them, remembered per binary afterwards).
+    /// See `AppRuntimeDetector`.
     public let runtime: AppRuntime?
 
     /// Which of Apple's UI frameworks the executable links, and whether Swift is
--- a/DuoUpdaterCore/Sources/DuoUpdaterCore/Scan/AppRuntime.swift
+++ b/DuoUpdaterCore/Sources/DuoUpdaterCore/Scan/AppRuntime.swift
@@ -10,10 +10,10 @@ import Foundation
 /// the bundle does not say.
 ///
 /// Every case is decided from something the packager *wrote*: a framework it had
-/// to bundle, a directory its launcher needs, or the dylibs its executable links.
-/// Nothing here guesses from an app's name or vendor. See `AppRuntimeDetector` for
-/// the evidence behind each case, and note that `tauri` is the one heuristic in
-/// the set.
+/// to bundle, a directory its launcher needs, the dylibs its executable links, or
+/// a Cargo path baked into it. Nothing here guesses from an app's name or vendor,
+/// and nothing is inferred from the absence of a marker. See `AppRuntimeDetector`
+/// for the evidence behind each case.
 public enum AppRuntime: String, Sendable, Hashable, CaseIterable, Codable {
     /// A wrapped iPhone/iPad app running on Apple Silicon.
     case iOSApp = "ios"
@@ -21,7 +21,9 @@ public enum AppRuntime: String, Sendable, Hashable, CaseIterable, Codable {
     case catalyst
     /// Chromium + Node, bundled by Electron.
     case electron
-    /// A Rust binary drawing into the system WebView (Tauri / wry).
+    /// A Rust app built on the Tauri framework, drawing into the system WebView.
+    /// Requires the `tauri` crate itself: wry — the WebView layer Tauri sits on —
+    /// is embedded by apps that are not Tauri at all.
     case tauri
     /// Flutter's macOS embedder.
     case flutter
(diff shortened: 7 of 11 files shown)


jizhi0v0:
## Review round 1 — three findings fixed, plus a bug the fix itself introduced

All three turn on the same promotion: a nil from the Tauri reader used to be a missing subtitle and is now a **verdict**, so everything that could produce a nil by accident needed looking at again.

**1. A failed read was indistinguishable from a proven absence, and got cached as one.** `scan` swallowed every read error into the same nil as a completed walk, and treated any short read as EOF. Split into `Probe` with three outcomes; only a walk that reached the end may be remembered.

Writing that fix introduced a worse version of the same bug — and the new test caught it. `FileHandle.read(upToCount:)` returns nil **at EOF**, and `try?` flattens a thrown error into that same nil, so `guard let chunk = try? …` reported *every completed walk* as unreadable. Nothing was ever cached and Longbridge was re-read on every scan. The verdicts stayed correct, which is precisely why only a memoization test could see it:

```
second sweep, before: 0.72s
second sweep, after:  0.06s
```

**2. The cache key was a marketing version string.** This repo documents at length that many apps ship new builds under an unchanged one (Amp: ten in a day, all `1.0`), and it also cannot see a bundle rewritten under a running scan by an install. The key is now the executable's **size and modification date** — it changes when and only when the bytes the answer was read from change. The `appVersion` parameter goes away with it, and with it the requirement that two call sites agree on which version string to pass.

**3. `carriesTauriCrate` took an `infoPlist` it used only for that key** while resolving the executable by re-reading the same plist off disk. Both gone.

Also corrected: the cost comment counted candidates in `/Applications` alone. It is **six bundles and 471 MB**, not three — `~/Applications` holds three more (CC Switch, Clash Verge, Handy). And the two shapes the proof cannot see are now written down: a `tauri` taken as a git dependency (no version in the path), and the other slice of a fat binary (one of the six is universal).

Four tests now pin the cache itself, since no verdict would ever notice losing it.

## Verification

Real-library sweep unchanged, release build:

```
SWEEP total=145  elapsed=0.50s   second pass 0.06s
SWEEP changed=["Longbridge.app: tauri -> native"]
Conductor 2.11.1 · STCM Editor 2.10.1 · CC Switch 2.10.3 · Clash Verge 2.11.5 · Handy 2.11.5 · Longbridge nil
```

Those six versions were independently recomputed by porting the rule to Python and running both a windowed port of the Swift chunking **and** a whole-file regex over the real binaries — all three witnesses agree.

## Unrelated red in `make test`, pre-existing on `origin/main`

`installPipelineDryRun` fails on this machine with *"Could not read a Team Identifier from the installed app"*. It reproduces on unmodified `origin/main` (22.7s), and this branch touches no install-path file. Cause: `~/Applications/Eudic.app` is **ad-hoc signed** (`Signature=adhoc`, `TeamIdentifier=not set`), so the gate correctly refuses it and the test counts a correct refusal as a failure. Filed separately rather than papered over here.

jizhi0v0:
## Review round 2 — a false positive, and a coverage regression round 1 caused

### 1. The 4 MiB window edge was being read as the file edge

`firstVersion` gets one window of a larger file and applies two rules that are both claims about the **file**:

- **The identifier guard.** `index > 0` asks "is there a character before this match". In a continuation window index 0 is not the start of anything — its real predecessor is in the previous chunk — so the guard **silently switched off at exactly one offset per chunk**.
- **The digit run.** A run that stopped because the buffer stopped was taken as a finished version.

Reproduced against this code, then again against an independent Python port of the rule:

```
xtauri-2.11.5  at the boundary → 2.11.5     ← accepted; same bytes mid-chunk → None
tauri-2.11.50  at the boundary → 2.11.5     ← truncated; mid-chunk → 2.11.50
```

Both predate this branch. The second is cosmetic either way. **The first is not**: `xtauri-1.2.3` at the wrong offset in a non-Tauri binary yields exactly the label this PR exists to stop handing out, so it is fixed here rather than filed.

Fix: the carry grows to 128 bytes and continuation windows start at index 1, so every position examined has its predecessor in hand — and a match starting on that first byte had `overlap` bytes of room in the previous window, so it was already whole there. `probe` takes the file size up front so a window can tell "my buffer ended" from "the file ended".

Chromium was the reader most exposed (four-digit components, ~60 boundaries in a Chrome framework). After the fix: `Chrome 152.0.7977.65`, `Helium 152.0.7977.64`, `Spotify 146.0.7680.179`.

Both new tests were **confirmed red against the unfixed reader** before the fix landed.

### 2. Round 1's cache fix silently switched the cache off for seven tests

Keying on the executable meant `executableURL` had to resolve one — and every Electron, Qt, Java and Chromium fixture in `RuntimeVersionTests` is built from frameworks and files with **no `Contents/Info.plist` at all**. Key came out nil; seven tests began asserting their values against a reader with memoization disabled, and passed exactly as before.

That is the same failure shape commit `b9f1d14` was written to close, reopened one layer down by the fix itself. `VersionBundle` now always writes an executable and a plist naming it.

### 3. Smaller, all real

- **A nil produced only because `scanningBinaries` was false is no longer remembered.** It is an answer about the caller, not the app, and both callers now share a key. Latent, not live — `.tauri`/`.chromium` return before the write, but `.electron` cannot, so one `scanning: false` read would pin "no version" on a re-signed Electron framework for the process lifetime.
- **Three tests marked `.enabled(if: geteuid() != 0)`.** `chmod 000` does not stop root: under root one fails and two pass **vacuously**, leaving the suite green and no longer pinning the thing it exists to pin.
- **Four false comments**, one of which predates the branch and is why a second existed: the file header still listed Chromium among the runtimes with no readable version, which stopped being true when `chromiumVersion` began reading the user-agent instead of the plist. I copied that stale claim into `RuntimeTag` in `d8ed28a`; both corrected.
- Corrected counts and honesty: "five stop early" → measured first-match offsets are 34% / 52% / 63% / 66% / 86% of the way into each file; MB → MiB; the symlink and coarse-timestamp caveats on the cache key are now written down; `.unreadable` reading as `.native` is documented rather than implied.

## Verification

```
make test
━ Test run with 1777 tests in 139 suites passed after 453.648 seconds with 1 known issue.
✔ Test run with 184 tests in 24 suites passed
✓ localizable keys agree — 429 in the catalog, all reachable
✓ version comparisons discriminate — 199 files, 7 marketing-first picks, all display-only
✓ app audits consistent — 89 docs
```

Real-library sweep, release build, unchanged from round 1:

```
SWEEP total=145  elapsed=0.49s   second pass 0.05s
SWEEP changed=["Longbridge.app: tauri -> native"]
Conductor 2.11.1 · STCM Editor 2.10.1 · CC Switch 2.10.3 · Clash Verge 2.11.5 · Handy 2.11.5 · Longbridge nil
```

Note for whoever merges: this branch adds a paragraph to the **0.3.78** section of `CHANGELOG.md`, and another session is working in the same section right now.

jizhi0v0:
## Review round 3 — and a correction to what I claimed in round 2

**I have to withdraw a sentence from the round-2 comment.** I wrote that both boundary tests were "confirmed red against the unfixed reader". One of them was not.

`aCandidateAtAChunkBoundaryIsStillHeldToTheIdentifierRule` wrote the overlap as a literal `128` and placed its payload at `chunkSize - 128 - 1`. The reader it was meant to catch used an overlap of **64**, which puts that offset in the middle of a window — where the identifier guard works correctly and always did. My red-check reverted `from:` and `isFinal:` but kept the widened overlap, which is the one part of the fix that *moves the seam*. Verified just now against a full revert:

```
full revert (overlap 64, from: 0, isFinal: true)
  ✔ aCandidateAtAChunkBoundaryIsStillHeldToTheIdentifierRule   ← vacuous
  ✘ aVersionCutByAChunkBoundaryIsReadWhole
```

The overlap is now `RuntimeVersion.scanOverlap` and the test reads it, so the payload tracks the seam rather than a copy of a number — and a future change to the overlap cannot silently move the test off it. Confirmed red against a full revert this time, **both** tests:

```
✘ aCandidateAtAChunkBoundaryIsStillHeldToTheIdentifierRule
✘ aVersionCutByAChunkBoundaryIsReadWhole
```

## The other real finding: `isFinal` was computed from a stale size

Reading the file size once up front is wrong in both directions, and one of them is exactly the failure class `Probe` was added to prevent:

- **Truncated under the walk** → `consumed` never reaches `size` → *no* window is final → a version ending at the real EOF is discarded → for `.tauri` that nil is **remembered as a verdict**: a cached "not Tauri".
- **Appended to** → the loop ends on an empty read, not at `size`, so it reads past it with every later window wrongly calling itself final — reinstating the truncation bug for the appended region. The comment claiming the walk "ends where it was going to end anyway" was simply false.

The loop now reads **one chunk ahead**: a window is the last one when the next read comes back empty. That cannot go stale, and it restores the reader on unseekable files, which `seekToEnd` had quietly excluded.

## Also corrected

- The `scanningBinaries: false` memo guard is now described accurately — it applies to **every** runtime, not just Electron, and it closes a trap rather than a live bug (no production caller passes false).
- **Withdrawing a second claim:** round 2 said Chromium "was the reader most exposed" to the boundary bugs. Measured across all 148 main executables and all 40 Electron/CEF/Chrome frameworks on this machine, the fix changes **zero** answers. It was the most exposed *in principle* — four-digit components, ~60 boundaries per framework — and was not actually being bitten.
- A doc comment that had drifted over the wrong overload of `executable`.

## Known, documented, and unreachable

`from: 1` combined with `isFinal` can drop a candidate whose digit run is longer than `scanOverlap - needle.count` — about 122 characters for `tauri-`. Real versions are under 20. It fails toward **missing** a match, never toward inventing one, and the constant's doc comment now says so.

## Verification

```
make test
━ Test run with 1777 tests in 139 suites passed after 446.730 seconds with 1 known issue.
✔ Test run with 184 tests in 24 suites passed
✓ localizable keys · ✓ version comparisons · ✓ app audits
```

Real-library sweep, release build, unchanged across all three rounds:

```
SWEEP total=145  elapsed=0.50s   second pass 0.05s
SWEEP changed=["Longbridge.app: tauri -> native"]
Conductor 2.11.1 · STCM Editor 2.10.1 · CC Switch 2.10.3 · Clash Verge 2.11.5 · Handy 2.11.5 · Longbridge nil
Chrome 152.0.7977.65 · Helium 152.0.7977.64 · Spotify 146.0.7680.179
```

Plus 20,000 randomised differential cases of the windowed rule against a naive whole-file reference: **0 disagreements** (3 on the pre-fix rule).

jizhi0v0:
## Review round 4 — one regression of mine, and a comment that has been false since before this branch

### 1. A failed lookahead threw away a match already in hand *(my regression, from round 3)*

The `failed` check sat **before** the window was scanned. So if chunk N carried the crate path and read N+1 threw, the proof was discarded and a proven Tauri app read as native for that scan.

`Probe`'s own doctrine says why that's wrong: reading to the end is what **`.absent`** needs, because absence is a claim about the whole file. **`.found` is positive evidence and complete in itself** — bytes after it failing to read say nothing about it. The scan now runs first, and only a *miss* is handed to `.unreadable`. A read that threw also no longer claims finality, since it knows nothing about the end either.

Transient, not cached — so strictly milder than the cached-verdict class round 2 fixed, but the same family.

### 2. The chunked walk was not keeping the file out of memory

This doc comment has been shipping since before the branch:

> *"in chunks so a 261 MB executable never lands in memory at once"*

It was describing an intent. `read(upToCount:)` hands back a bridged `NSData` that is **autoreleased**; nothing drained until the walk returned, so the whole file landed anyway — one chunk at a time. Measured myself on Longbridge, peak `phys_footprint` for a single probe:

```
without the pool   +277.8 MiB
with the pool       +29.7 MiB
```

Pre-existing, and **this branch is what made it matter**, by moving the walk onto the library-wide scan. A pool per window makes the sentence true.

### 3. `from: 1` could blind file offset 0 for the rest of the walk

It keyed off `carry.isEmpty` — and if a window is no larger than `scanOverlap`, the carry *is* the whole window, so the next window still starts at offset 0. And the one after that. A match there would never be examined again. Unreachable on regular files (no mid-file short read in ~700 real 4 MiB reads), but the fix is to stop asking a proxy question and track the window's file offset directly.

### Also

The chunk size was still a literal on the test side — half of round 3's fix. Tuning it would move the seam just as surely as the overlap does, and both boundary tests would go quietly green against the reader they exist to catch. Both constants now come from `RuntimeVersion`.

## Verification

```
make test
━ Test run with 1777 tests in 139 suites passed after 514.557 seconds with 1 known issue.
✔ Test run with 184 tests in 24 suites passed
✓ localizable keys · ✓ version comparisons · ✓ app audits
```

Real-library sweep, release build — unchanged across all four rounds, now including the Electron readers that share this code:

```
SWEEP total=145  elapsed=0.49s
SWEEP changed=["Longbridge.app: tauri -> native"]
Conductor 2.11.1 · STCM Editor 2.10.1 · CC Switch 2.10.3 · Clash Verge 2.11.5 · Handy 2.11.5 · Longbridge nil
Chrome 152.0.7977.65 · Helium 152.0.7977.64 · Spotify 146.0.7680.179
VS Code Electron 42.8.1 · Notion Electron 42.4.1
```

## Where this stands

Four rounds; every one found something real, and three of the four found a defect introduced by the previous round's fix. This round's own fixes are smaller and more local than the last two (an ordering, a pool, an offset), and the loop's behaviour is pinned by 40,000 fuzzed layouts against a whole-file reference plus two red-verified boundary tests. I'd merge it here rather than keep spinning, unless a fifth round is wanted.

jizhi0v0:
## Review round 5 — clean on the code, one real gap in the tests

**No defect found in round 4's fixes.** The reviewer ported `probe`/`firstVersion` into standalone Swift with an injectable reader and fuzzed **480,000 random layouts** against a whole-file reference — three needles, chunk sizes above *and below* the overlap, short reads anywhere, throws at any call, both EOF shapes — with **zero disagreements**, plus 546 (binary, needle) probes across 183 real executables, also zero. The harness is not vacuous: the pre-round-4 body produces 10–11 mismatches per 60k on every seed.

### The gap: three fixes that nothing could see

Replacing `probe`'s body with the pre-round-4 version left **every test in the suite green**. Rounds 2 and 4 fixed three real defects here — a match discarded when the read *after* it failed, a short read taken for end-of-file, and the first bytes of a file going unexamined for the rest of the walk — and none was expressible as a test, because `probe` opened its own `FileHandle` and no real file can be made to throw halfway through or return a short read.

`probe` now has an overload taking a closure that yields one chunk at a time — the shape this file's neighbours already use (`LibraryReader`, `TauriProof`). Five tests drive it with scripted reads, **each verified red against a revert of the specific fix it pins**:

```
revert the failed-check ordering + windowStart:
  ✘ aMatchSurvivesAReadThatFailsAfterIt
  ✘ aWindowNoLongerThanTheCarryDoesNotBlindTheStartOfTheFile
revert short-read-is-EOF:
  ✘ aShortReadInTheMiddleIsNotTheEndOfTheFile
```

The offset-0 bug reduces to 18 bytes once the seam exists:

```swift
probe([.chunk("tauri-9.16.2"), .chunk("a9  ")])   // pre-fix: .absent
```

No behaviour change — the URL overload opens the handle and calls the new one.

### Four comments corrected, three of which said the opposite of what this branch made true

- `InstalledApp.runtime` and `Preferences.showRuntimeTags` both claimed the runtime fact is **free** because the scan already reads what it comes from. It stopped being free here. `showRuntimeTags` is also **display-only** — turning it off hides the mark without saving the work — which matters under a default-ON toggle.
- *"Returns nil when neither source recognizes anything"*, sitting under a list this branch grew to three sources.
- *"An unreadable binary is not [cached]"* is true only of `.tauri`. Electron and Chromium go through `scan`, which flattens `.unreadable` into the same nil as a real absence, and the cache key is the *main* executable's stat — which can succeed while a framework binary cannot be read. Reachable; costs a missing version label, not a wrong runtime. Now documented rather than implied.

### Merged `origin/main`

Both sides had written a **0.3.78** section. Resolved inside the conflict block, all four paragraphs kept — the three from main first, the Tauri one last, nothing dropped or reordered. Notably #209 fixed the blank-name bug and #212 the 欧路词典 changelog, which are the two things I reported from this branch's test run and did not touch.

## Verification

```
make test
━ Test run with 1786 tests in 139 suites passed after 434.744 seconds with 1 known issue.
✔ Test run with 184 tests in 24 suites passed
✓ localizable keys agree — 430 in the catalog
✓ version comparisons discriminate — 199 files, all display-only
✓ app audits consistent — 90 docs
```

Five rounds. Rounds 1–4 each found a real defect and three of them found one introduced by the previous round's fix; round 5 found the code clean and the tests thin, which is now fixed. Ready to merge and ship as 0.3.78.

jizhi0v0:
## Correction, post-merge

I claimed in the round-1 comment on this PR:

> Note that a debug build is ~200× slower on this path (`-Onone` byte loop, 104s for the same sweep) — the new tests use tiny stub executables, so `make test` is unaffected.

**The second half is wrong.** The new tests are indeed fine. The *pre-existing* scan tests are not — they scan real `/Applications`, so they now walk six binaries in a `-Onone` build:

```
scanFindsRealApps @ f771c6a (before this PR)      0.239 s
scanFindsRealApps @ main    (after  this PR)    107.155 s      448×
```

`extraLocationsAreScannedOnTopOfBuiltInRoots` and `listUnknownApps` pay the same. I verified the release-build cost before shipping and never measured a debug-build scan test, which is exactly the gap the claim papered over.

No user impact — the shipped app is a release build (0.50 s for the whole library, 0.05 s on a repeat scan). Filed as #214 with a proposed fix (`Data.range(of:)` for the needle, keeping the hand-rolled loop only for parsing at match positions), which wants its own review pass rather than a hot patch, given how many rounds the window-edge rules took to settle.

Issue #269, opened 2026-09-02
notFaithful under-reports: 8 committed tiles carry the ImageRenderer placeholder and say nothing


jizhi0v0:
`notFaithful` lists 3 tiles. Measured against the committed sheet, **11** contain
`ImageRenderer`'s yellow/red "unavailable" placeholder — so 8 tiles have been
lying with nothing saying so.

Scanned every committed PNG for the placeholder's signature yellow
(R>230, G 170–225, B<70), counting tiles with more than 40 such pixels:

| tile | placeholder px | in `notFaithful`? |
|---|---|---|
| popover/02-relaunching | 666 | **no** |
| popover/05-installing-queued | 666 | **no** |
| popover/06-installing-downloading | 2198 | **no** |
| popover/07-installing-extracting | 666 | **no** |
| popover/18-update-major-upgrade | 638 | yes |
| popover/21-update-app-store-region-locked | 658 | yes |
| popover/22-update-app-store-mac-incompatible | 638 | yes |
| workbench/02-relaunching | 666 | **no** |
| workbench/05-installing-queued | 666 | **no** |
| workbench/06-installing-downloading | 3638 | **no** |
| workbench/07-installing-extracting | 666 | **no** |

The three that are listed are the documented case: an SF Symbol inside a
`.buttonStyle(.borderless)` button. The eight that are not are a **second,
separate gap** — `ImageRenderer` does not draw a plain native `ProgressView()`
(spinner, and the determinate bar in `06`). Found while adding #265's cases.

Why this one is worse than the documented three:

- **It hits both surfaces.** `CLAUDE.md` tells the reader that when a popover tile
  is unfaithful they should "看工作台那张 (它用 `Label`,渲染是对的)". For these
  four states that escape hatch does not exist — the workbench tile shows the same
  placeholder. There is no faithful picture of an installing row anywhere in the
  committed sheet.
- **It covers the states most likely to be read.** 02/05/06/07 are relaunching and
  the three install stages — exactly what someone opens the sheet to check after
  touching install progress.
- **The blank gate cannot see it.** A placeholder is not blank, so the tile counts
  as drawn. The collision gate cannot see it either: the placeholders sit in
  different positions with different labels beside them, so the tiles differ.

Scope: #265 registered the two tiles it added (`popover/37-stage-label-hidden`,
`workbench/37-stage-label-hidden`) and flagged the pre-existing eight rather than
silently leaving its own case off the list. Widening `notFaithful` to cover them
[...]

Issue #265, opened 2026-09-02
Gallery draws one of three download readouts, and none of the explanation popovers


jizhi0v0:
Three branches of the popover that the gallery draws nothing for, so a regression
in any of them is caught by no check at all (there is no test target over
`App/Sources`):

**1. The explanation popovers.** `majorUpgradePopover`, `regionHintPopover` and
`macCompatHintPopover` (`App/Sources/PopoverRowAction.swift`) sit behind `@State`
flags the harness cannot set. They hold the entire licence-boundary warning, the
region explanation, and **two buttons that actually install** ("Update anyway" and
"Update in background", both calling `actions.install()`). #259's argument for
keeping the workbench's gated states as a pointer — "the explanation that makes
this safe lives in the popover" — rests on content nothing renders.

Would need the presented flag to be injectable, e.g. an internal init parameter
the gallery sets, or splitting the popover body into a view drawn on its own.

**2. `DownloadReadout` has three cases and the gallery draws one.**
`RowStateGalleryCases.surfaces` constructs `PopoverRowAction(state:result:)`, so
`downloadReadout` takes its `.barAndPercent` default. `ringAndPercent`, `ringOnly`
and the `.help("Downloading …")` fallback are drawn by nothing. The declaration
order of this enum is load-bearing — `AppRow` walks `allCases` and takes the first
fit — so a reorder degrades every downloading row silently.

**3. `showsStageLabel` has two branches and the gallery draws one** (the default
`{ _ in true }`), so the unlabelled-spinner path in the popover's stage progress
is never drawn.

Fix for 2 and 3 is cheap: add cases to `RowStateGalleryCases.all` that pass the
non-default values, the same way the App Store and MAS/TestFlight fixtures were
added. 1 needs a seam.

Found by the second adversarial review on #259.

PR #270, opened 2026-09-02
Row state: push the last three view-side opinions into the state

### LINKED ISSUE #260: Row state: push the last three view-side opinions into the state
`ui = f(state)` is true for the popover and one step short for the workbench.
Three places a view still forms its own opinion about a row, each of which can
drift between the two windows without any check noticing:

1. **`WorkbenchRowAction` re-reads `result.remote?.appStore`** to pick between the
   one-click button and the amber "not available" Label
   (`App/Sources/RowActionViews.swift`, `.appStore` case). The region /
   Mac-compatibility gate is therefore evaluated in a view rather than carried in
   the route, so the two windows evaluate it independently.
2. **`PopoverRowAction` reads `result.app.isMASApp` / `isTestFlightApp`** to pick
   between three pictures for `.upToDate`.
3. **`sourceHint(for:)`** (now shared, `App/Sources/RowActionViews.swift`) reads
   `isMASApp` / `sparkleFeedURL`. Sharing the function stops the two windows
   disagreeing, but the opinion is still formed in the view layer.

Proposed shape:

- `UpdateRoute.appStore(managedHere: Bool, gate: AppStoreGate)` with
  `.none / .region / .macIncompatible`.
- `RowActionState.upToDate(channel:)` and `.noSourceCovers(hint:)`.

Two things fall out for free. The workbench could then draw the two App Store
gates **distinctly** instead of collapsing them into one Label — which is what
currently forces the `mayLookAlike` exemption
`workbench/21-update-app-store-region-locked == workbench/22-...-mac-incompatible`.
And `RowStateGalleryCases` stops needing per-row fixtures for `.upToDate` and
`.noSourceCovers`, because the branch would key off the state again.

Context: #259 moved the ladder into Core and both windows now read
`AppListModel.rowState(for:)`. These three are what is left. Found by the second
adversarial review on that PR.

jizhi0v0:
Issue #260, the follow-up to #259: three places a view still formed its own
opinion about a row, all found by the second adversarial review on that PR.

## What moved into state

1. **`UpdateRoute.appStore(managedHere:)` → `.appStore(managedHere:, gate:)`.**
   The workbench used to re-read `result.remote?.appStore` and test
   `!info.isRegionMismatch, !info.isLatestMacIncompatible` to choose between a
   one-click button and an amber Label — the branch decision lived in a view,
   duplicated (or duplicable) between the two windows. `AppStoreGate` (`.none`
   / `.region` / `.macIncompatible`) is now resolved once, in
   `AppStoreGate.resolve(_:)`, with the same precedence the popover's
   explanation used when both are somehow true (Mac-incompatible over region).
2. **`RowActionState.upToDate` → `.upToDate(channel:)`.** The popover branched
   on `result.app.isMASApp` / `isTestFlightApp` to pick between three
   pictures; the workbench had no branch at all and drew `EmptyView` for all
   three. `UpToDateChannel` (`.none` / `.appStore` / `.testFlight`) is decided
   in `RowAction.state`, same priority (App Store first) the popover used to
   apply itself.
3. **`RowActionState.noSourceCovers` → `.noSourceCovers(hint:)`.**
   `sourceHint(for:)` took an `UpdateResult` and read `isMASApp` /
   `sparkleFeedURL` directly; sharing the function kept the two windows from
   disagreeing, but the opinion was still formed in the view layer. It now
   takes the state's `SourceHint`.

All three priorities are fed by three new plain-bool `RowActionFacts` fields
(`isMASApp`, `isTestFlightApp`, `hasSparkleFeed`) and one new `RouteInputs`
field (`appStoreGate`), filled once per row in `AppListModel`.
`RowActionFacts.route` stays the `@autoclosure` deferral point —
`routeIsDeferred` still passes, unchanged.

## What that let the workbench do

The workbench now draws the two App Store gates **distinctly** — `Region-locked`
with the globe badge, `Not supported on this Mac` with the triangle badge
(reusing the popover's own catalogue strings for both), instead of collapsing
both into one "App Store" Label. That retires the `mayLookAlike` exemption
`workbench/21-…-region-locked == workbench/22-…-mac-incompatible` — CLAUDE.md
is explicit that the person who makes a view stricter is the one who drops the
now-dead exemption, and the gallery's collision gate fails the build on a
stale one anyway.

`RowStateGalleryCases` also stopped needing per-row `UpdateResult` fixtures for
`.upToDate` and `.noSourceCovers` (`masInstalled` / `sparkleInstalled` /
`testFlightInstalled` are gone) — the branch keys off the state again, so the
plain `app` row reaches all of them.

The workbench's `.upToDate(channel: .appStore / .testFlight)` now draws the
same tile as `.managedElsewhere` for that channel (by design — a managed app
that happens to be current must not look updatable from here), so
`31-up-to-date-app-store` / `32-up-to-date-testflight` move out of
`mayBeBlank` and the workbench halves of those two pairs join `mayLookAlike`,
matching the popover halves already there.

## `ui = f(state)` — what's left

Neither view forms its own opinion about *which branch to draw* any more. One
thing is deliberately left alone: both `.appStore` cases keep an
`if let info = result.remote?.appStore { … } else { openButton }` nil-guard.
The route only resolves to `.appStore` when `RouteInputs.hasAppStoreAvailability`
was true (i.e. `result.remote?.appStore != nil` at the moment the route was
built), so the `else` is defensive against a data race that shouldn't occur
within one render, not a second opinion on the gate — `info` itself is read
only for display content below that guard (a deep link, a region name), never
to pick between pictures.

## Not renumbered

No gallery case was renamed or renumbered — cases 19-22, 26, 30-34 keep their
names and positions; only their `RowActionState` value and (for 31-34) their
fixture changed. Safe to merge in either order relative to #265 (which appends
new cases at the end of `RowStateGalleryCases.all`).

## Verification

```
cd DuoUpdaterCore && swift test               # 1926 tests, 151 suites, 1 known issue, all passing
swift test --package-path CLI                  # 184 tests, 24 suites, passing
python3 scripts/check_localizable_keys.py --derived-data /tmp/duo-loc-check-260   # 430 keys, all reachable
python3 scripts/check_staged_version_use.py    # unaffected, still clean
python3 scripts/check_app_audits.py            # unaffected, still clean
python3 scripts/test_appcast_edit.py           # 45 tests, passing
GALLERY_DD=/tmp/duo-gallery-260 bash scripts/row-state-gallery.sh
  # rendered 68 states
  # NOT FAITHFUL: popover/18, popover/21, popover/22 (pre-existing ImageRenderer/
  #   borderless-button SF Symbol issue, unrelated to this change — read the
  #   workbench tile for those three)
  # no state renders blank
  # no two states draw the same tile
```

Four PNGs changed (all workbench: 21, 22, 31, 32); the other 64 are
byte-identical.

### Mutants run

| mutant | caught by |
|---|---|
| `UpdateRoute.resolve` hardcodes `gate: .none` instead of reading `inputs.appStoreGate` | `UpdateRouteResolutionTests.appStoreCarriesGate` (2 of 3 cases fail) |
| Workbench's `.region` / `.macIncompatible` branches collapsed back to one shared `Label("App Store", …)` | `make gallery`'s collision gate: `IDENTICAL: workbench/21-update-app-store-region-locked == workbench/22-update-app-store-mac-incompatible` |

Both mutations were applied, run, confirmed to fail exactly as expected, then
reverted before committing (including re-rendering the gallery to restore the
correct PNGs — the mutant run had overwritten `verify/row-states` with the
broken pictures on its way to failing).

Closes #260

---

## Rebased onto main after #268 merged

`main` moved (#268, issue #265) while this was in review — 12 new gallery tiles
(cases 35-40, appended after mine, so no renumbering), a `tile(name, state,
result)` signature change in `main.swift`, and the three popover explanation
panels losing `private`. Rebased; the only real conflict was in
`RowStateGalleryCases.all` / `mayBeBlank` (both touched cases 31-34's
neighbourhood) — resolved by keeping this PR's state-driven 31-34 and main's
new 35-40 side by side, and adding `gate:` to the two new explanation cases
(39, 40) that construct `.appStore(managedHere:)` routes without it.

**Case numbers**: unchanged. Nothing in this PR was renumbered or renamed;
cases 19-22, 26, 30-34 keep the names and positions they had before the
rebase. Tile count: 68 → 80 (34→40 states × 2 surfaces), entirely from #268's
6 additions, not from anything here.

**Re-ran `make gallery`** after `git checkout origin/main -- verify/row-states`
(took main's side for the whole directory rather than trusting the binary
merge, per instructions) and a fresh render:

```
rendered 80 states → …/verify/row-states
NOT FAITHFUL (see RowStateGalleryCases.notFaithful for why each one): popover/18-update-major-upgrade, popover/21-update-app-store-region-locked, popover/22-update-app-store-mac-incompatible, popover/37-stage-label-hidden, workbench/37-stage-label-hidden
DownloadReadout order intact (barAndPercent, ringAndPercent, ringOnly)
no state renders blank
no two states draw the same tile
```

The regenerated 21/22/31/32 tiles were byte-identical to what was already
committed (`compare -metric AE` reported `0 (0)` against each). One unrelated
tile, `workbench/04-just-updated.png`, came back 5 bytes different from a
non-deterministic PNG re-encode of visually identical pixels (`compare`
reported `9.7e-08` — effectively a rounding artifact, not a content change);
reverted to main's committed bytes rather than carry unrelated noise. Net
image diff after the rebase: still exactly the same 4 files as before
(21, 22, 31, 32).

**The `mayLookAlike` removal still holds.** The collision gate's exact line
after the rebase: `no two states draw the same tile` — the region/Mac-incompatible
workbench pair draws distinctly now, same as before the rebase (confirmed by
re-running the mutation from the "Mutants run" section above wasn't needed
again since the gate output already answers it; if it hadn't held, the line
would instead have read `IDENTICAL: workbench/21-update-app-store-region-locked
== workbench/22-update-app-store-mac-incompatible`, which is what mutant (b)
above produced when I deliberately broke it).

**Issue #269 (11 tiles carry the `ImageRenderer` placeholder, `notFaithful`
only names 3)**: unaffected by this PR. My workbench changes (21, 22, 31, 32)
never wrap an SF Symbol in a `.buttonStyle(.borderless)` button — the
region/Mac-incompatible badges are plain `Label`s, and the up-to-date channel
tiles reuse `appStoreManagedTile`/`testFlightManagedTile` (an `Image(nsImage:)`
and a `Text`) — so none of them start or stop showing the placeholder. My
popover changes (`appStoreTrailing`) kept the same borderless-button
structure, just swapped what condition selects each branch; popover/21 and
popover/22 were already in `notFaithful` before this PR and still are, unchanged.

**Full verification, post-rebase, on a quiet machine:**

```
cd DuoUpdaterCore && swift test    # 1926 tests, 151 suites, 1 known issue, 39.7s (vs ~8min before — no sibling worktrees competing)
swift test --package-path CLI      # 184 tests, 24 suites, passing
python3 scripts/check_localizable_keys.py --derived-data /tmp/duo-loc-check-260   # 430 keys, all reachable
python3 scripts/check_staged_version_use.py    # clean
python3 scripts/check_app_audits.py            # clean
```


jizhi0v0:
Reviewed. Three things I verified independently rather than taking from the PR body:

**1. The gate precedence is a faithful transcription.** `AppStoreGate.resolve` checks `isLatestMacIncompatible` before `isRegionMismatch`; `git show main:App/Sources/PopoverRowAction.swift` has the same order at lines 500/513. Worth noting what this PR actually recovers: on `main` the workbench *did* distinguish the two gates — but only in `.help()` text (`RowActionViews.swift:298`), which no check can see and no screenshot contains. The visible Label was shared. So this is one of the tooltip-only agreements `CLAUDE.md` warns about being paid for, converted into something the collision gate can enforce.

**2. The `.upToDate` channel test bites.** I mutated `RowAction.state` to always return `.upToDate(channel: .none)`:

```
✘ "a current row's channel: App Store beats TestFlight beats neither" — 3 issues
✘ Test run with 19 tests in 1 suite failed
```

Restored → 19/19 green. It pins the priority, not just the presence of a payload.

**3. `ui = f(state)` now holds in the workbench `.appStore` branch.** The `if let info = result.remote?.appStore, !info.isRegionMismatch, !info.isLatestMacIncompatible` condition is gone; `info` survives only as an optional read for the deep link (`info?.deepLink`). That is display content, not branch selection, and the PR body says so accurately.

**On the two added `mayLookAlike` entries** (`workbench/27 == workbench/31`, `workbench/29 == workbench/32`): these are a net improvement, not a new cost. Those workbench tiles were previously *blank* and exempted as blank — which was the bug. They now draw the managed marker and are exempted as deliberate look-alikes. Blank-because-nobody-drew-it became identical-because-they-should-be.

**One gap this PR revealed, filed as #271:** removing `workbench/31` and `workbench/32` from `mayBeBlank` was done by hand and correctly — but nothing would have caught leaving them in. `mayLookAlike` has a dead-exemption gate; `mayBeBlank` does not, so a stale entry there silently disables the blank detector for that tile. Not this PR's job to fix.

PR #272, opened 2026-09-02
Chronic check failures: divergence between the banner and the row badge is intentional

### LINKED ISSUE #264: Chronic check failures are suppressed in one place and nagged about in two
`AppListModel.failedCheckResults` suppresses a row after ~3 consecutive failures,
deliberately: a permanently dead feed should not pin a banner with a Retry that
re-runs the same 404.

`RowAction.state` has no such notion. Every `.error` row draws a permanent orange
"Failed + ↻" badge. That was already true of the popover; #259 gave the workbench
the same badge, so the surface is now doubled — and the two mechanisms hold
opposite opinions about whether a chronically failing check is worth nagging
about.

Either carry `chronic` into `RowActionFacts` and let the ladder answer it (a
distinct quieter state, or `.checkFailed(message:rateLimited:chronic:)`), or
write down that the divergence is intended and why. Right now it is neither.

Found by the second adversarial review on #259, under "the `.checkFailed` payload".

jizhi0v0:
Issue #264: `AppListModel.failedCheckResults` excludes a row after 3 consecutive
`.error` rounds from the failed-check banner; `RowAction.state(for:)` has no such
notion, so the same row's badge nags forever. The issue asks: carry `chronic`
into `RowActionFacts` and let the ladder answer it, or write down that the split
is intended and why.

## Decision: the divergence is intentional. Documented it, did not change the ladder.

The banner and the per-row badge are not the same claim, and suppressing one is
not the same call as suppressing the other:

- **The banner is one global claim** — "your checks are healthy" — with **one
  Retry that reruns every failed check**, including whichever ones are
  permanently dead. Once a failure looks permanent (a vendor retired its feed —
  Alfred's appcast 404'd for weeks), that claim is false and that Retry is
  wasted work every time the popover opens. Excluding chronic rows from it is
  correct and unrelated to what any individual row says about itself.
- **The badge is a fact about one row**, with its **own per-row Retry that
  re-checks only that app** (`AppListModel.retry(_:)`, wired in both windows —
  see `MenuContentView.swift:969` / `WorkbenchWindowView.swift:1009`), not the
  bulk one the banner uses. "This check has failed" does not stop being true
  because it has been failing for a while; if anything a long streak is *more*
  worth knowing, not less.

**What a chronically-failing row looks like today, in each window — traced,
not assumed:**

- **Popover, default view.** Never shown, chronic or not: `MenuContentView
  .visible` filters on `model.needsAction(_:)`, which only counts actionable
  updates / restarts / staged relaunches — `.error` never qualifies, so an
  `.error` row (chronic or fresh) is invisible here regardless. Chronic-ness
  changes nothing about this window's default view.
- **Popover, "Show all".** Every row renders, including every `.error` row.
  `RowAction.state` maps it to `.checkFailed`, and `PopoverRowAction
  .errorBadge` draws the orange "Failed"/"Rate-limited" text + a `↻` Retry
  button — identically whether the row is one round or thirty rounds deep.
- **Workbench.** `WorkbenchWindowView.apps` is `model.results` filtered only by
  source (`!= "Homebrew"`) — **no `needsAction`/"Show all" gate at all**.
  Every result is a row, always. A chronic `.error` row draws the same
  Failed/Retry badge as popover's Show-all view (`RowActionViews.swift:176`,
  shared with the popover since #259).

`VisibilityRules` (Core) has nothing to do with any of this — it only answers
"is this app ignored" and "is this version skipped"; it has no branch for
check failures at all, chronic or not.

**Why the badge must not go quiet:** the workbench shows every row
unconditionally. If a chronic row's badge were blanked or demoted to
something silent, the row would render the same as an up-to-date row in the
one window that has no other cue — exactly the "空白被读成没事" failure mode
`CLAUDE.md` documents for this workbench (the reason #259 gave it real
branches for `.error`/`.unknown`/managed/ignored/skipped in the first place,
instead of `EmptyView`). A row that silently stops reporting is worse than
one that keeps nagging.

**Is `consecutiveCheckFailures` reachable and cheap at the point facts are
built?** Yes on both counts, which is why I didn't reject the "carry it into
`RowActionFacts`" option on feasibility grounds — `AppListModel.rowState(for:)`
is a method on the same object that owns `consecutiveCheckFailures`, and a
dictionary lookup by `result.id` costs the same as the half-dozen other
lookups (`relaunching.contains`, `installing[result.id]`,
`pendingBatchRestart[result.id]`, …) it already does inline per row per
redraw. Feasibility isn't what settles this — the blank-row risk above does.

**Would a `chronic:` flag on the existing `.checkFailed` payload (no rendering
change) be safer than nothing?** No — it would be worse than nothing: an
unused field is a fabricated fact nobody asked for, and `make gallery`'s
collision gate would immediately (and correctly) flag "chronic" and
"non-chronic" `.checkFailed` tiles as identical pixels — a real collision,
not a fixture problem, since nothing would read the flag to draw them
differently. Better to not carry information the state has nowhere to spend.

## What changed

Not the ladder — the divergence needed a home to be pinned down, and
`AppListModel`'s own threshold was a private `App`-target constant, which
`CLAUDE.md` already flags as the wrong place for anything that needs a test
(`App/project.yml` carries no test target; nothing in `App/Sources` runs).
So:

- **`DuoUpdaterCore/.../Engine/CheckFailureRules.swift`** (new): extracts the
  boundary AppListModel was comparing against inline
  (`consecutiveFailures >= chronicThreshold`, threshold `3`) into a small,
  testable Core type, with the "why doesn't `RowActionState` read this"
  reasoning written down where the next person will find it.
- **`AppListModel.swift`**: `failedCheckResults` and `recordCheckOutcomes` now
  call `CheckFailureRules` instead of a private constant. No behavior change —
  same threshold, same comparison, same log line. Doc comments on
  `consecutiveCheckFailures` and `failedCheckResults` cross-reference
  `RowActionState.checkFailed` and this issue.
- **`RowActionState.swift`**: `.checkFailed`'s doc comment cross-references
  `AppListModel.failedCheckResults` / `CheckFailureRules`, spelling out why the
  case deliberately carries no streak.
- **`CheckFailureRulesTests.swift`** (new): pins the boundary.

No new `RowActionState` case, no `RowActionFacts` field, no view change — so
no gallery tile changes either (confirmed below).

## Mutants (the one new guard: `CheckFailureRules.isChronic`)

| mutant | result |
|---|---|
| `>=` → `>` | `CheckFailureRulesTests.boundary` fails: `isChronic(consecutiveFailures: chronicThreshold)` now reads `false` |
| body hardcoded to `true` | fails: the two "not yet chronic" assertions (`0`, `threshold - 1`) go red |
| body hardcoded to `false` | fails: the two "chronic" assertions (`threshold`, `threshold + 10`) go red |

All three applied to the working tree, run with `swift test --filter
CheckFailureRules`, confirmed red for the expected reason, then reverted
(diffed back to zero against the pre-mutation file) before committing.

## Verification

```
  # 1927 tests, 152 suites, 1 known issue, all passing (40.3s)
swift test --package-path CLI
  # 184 tests, 24 suites, passing
python3 scripts/check_localizable_keys.py --derived-data /tmp/duo-loc-check-264
  # 431 keys, all reachable
  # unaffected, still clean
  # unaffected, still clean
  # 45 tests, passing
GALLERY_DD=/tmp/duo-gallery-264 bash scripts/row-state-gallery.sh
  # rendered 80 states
  # NOT FAITHFUL: popover/18, popover/21, popover/22, popover/37, workbench/37
  #   (pre-existing, unrelated to this change — see #269)
  # no state renders blank
  # no two states draw the same tile
```

Zero PNG diffs against `main`, except one that isn't real: re-running the
gallery twice, byte-for-byte identical code both times, reproducibly
re-encoded `workbench/04-just-updated.png` 5 bytes different
(`compare -metric AE` / RMSE: well under one pixel-equivalent — not a content
change). #270's own PR body hit and documented this exact same tile the same
way during its rebase. Reverted rather than carry unrelated noise; not
something this issue touches or fixes.

Closes #264

### DIFF
--- a/App/Sources/AppListModel.swift
+++ b/App/Sources/AppListModel.swift
@@ -5835,12 +5835,12 @@ final class AppListModel {
     /// by `UpdateResult.id` (the install path). In memory only — a fresh launch
     /// starts everyone at zero, which is the right bias: after a relaunch we have no
     /// evidence a failure is chronic and should say so once more.
+    ///
+    /// This is the ONLY place a chronic streak is tracked. `RowAction.state(for:)`
+    /// — the ladder both the popover and the workbench draw a row from — has no
+    /// notion of it and is not meant to: see `failedCheckResults` below and
+    /// `CheckFailureRules` for why (issue #264).
     @ObservationIgnored private var consecutiveCheckFailures: [String: Int] = [:]
-
-    /// Rows this many rounds deep stop driving the banner and the header count.
-    /// Three rounds of a check interval (an hour by default) is long past the point
-    /// where "retry" is the useful advice.
-    private static let chronicFailureThreshold = 3
 
     /// Update the streaks from one completed round, and drop rows that are gone.
     /// Called only from `performRefresh` — deliberately NOT from `retryFailedChecks`,
@@ -5854,7 +5854,7 @@ final class AppListModel {
             }
         }
         let newlyChronic = next.filter {
-            $0.value == Self.chronicFailureThreshold
+            $0.value == CheckFailureRules.chronicThreshold
         }.keys.sorted()
         if !newlyChronic.isEmpty {
             Log.app.info(
@@ -5863,16 +5863,30 @@ final class AppListModel {
         consecutiveCheckFailures = next
     }
 
+    /// Rows this many rounds deep stop driving the *aggregate* surfaces below —
+    /// the banner, the header's "N not checked" line, and the bulk-retry target
+    /// list. They do NOT stop being `.checkFailed` rows: `RowAction.state(for:)`
+    /// never reads `consecutiveCheckFailures` (it has no parameter for it), so a
+    /// chronic row keeps its own orange Failed badge — with its own per-row
+    /// Retry, which re-checks only this app rather than every failed row the way
+    /// the banner's Retry does — for as long as it keeps failing.
+    ///
+    /// That split is deliberate, not an oversight (issue #264): the banner is one
+    /// global claim ("your checks are healthy") that becomes false advice once a
+    /// failure is permanent — a vendor that retired its feed fails identically
+    /// forever, and Alfred's appcast did exactly that for weeks. Left counted, it
+    /// pins the banner up with a Retry that just re-runs the same 404, and stops
+    /// the header from ever reading "up to date" again. A row's own badge makes
+    /// no such claim; it states one fact about one app, and that fact stays true.
+    /// Silencing it would blank the row in the workbench, which shows every
+    /// result unconditionally (no "Show all" gate) — and a blank row there reads
+    /// as "up to date", which is the specific failure mode `RowActionState`
+    /// exists to rule out for every non-quiet state. See `CheckFailureRules`.
     var failedCheckResults: [UpdateResult] {
         results.filter {
             guard case .error = $0.status else { return false }
-            // A vendor that retired its feed fails identically forever — Alfred's
-            // appcast 404'd for weeks. Left in, that pins the banner permanently with
-            // a Retry button that re-runs the same 404, and stops the header from
-            // ever reading "up to date" again. Those rows fall back to the old
-            // behaviour: visible under "Show all", reported by `RecipeHealth` and
-            // `duo verify`, and silent here.
-            return (consecutiveCheckFailures[$0.id] ?? 0) < Self.chronicFailureThreshold
+            return !CheckFailureRules.isChronic(
+                consecutiveFailures: consecutiveCheckFailures[$0.id] ?? 0)
         }
     }
 
--- a/DuoUpdaterCore/Sources/DuoUpdaterCore/Engine/CheckFailureRules.swift
+++ b/DuoUpdaterCore/Sources/DuoUpdaterCore/Engine/CheckFailureRules.swift
@@ -1,0 +1,33 @@
+import Foundation
+
+/// How many consecutive check failures make a row "chronic" — worth suppressing
+/// from `AppListModel`'s *aggregate* opinions about the whole list, never from a
+/// row's own account of itself.
+///
+/// `AppListModel.failedCheckResults` uses this to drop a row from the failed-check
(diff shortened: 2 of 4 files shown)


jizhi0v0:
Reviewed, and I checked the load-bearing half of the argument myself rather than taking it from the body — because "the divergence is intentional" is the kind of conclusion that is only worth anything if the visibility claims under it are true.

All three hold:

- **`needsAction` has no `.error` branch** (`AppListModel.swift:566`) — it returns true only for an actionable update, a pending restart, or a nudgeable staged update. So the popover's default view hides every failed row, chronic or not.
- **`VisibilityRules` has no notion of check failure at all** — it lives in `DuoUpdaterCore/Sources/DuoUpdaterCore/Engine/UpdateSettings.swift`, and grepping it for error/failed returns nothing.
- **The workbench really does show every row.** `WorkbenchWindowView.apps` filters exactly one thing — `sourceName != "Homebrew"` — then sorts. `filteredApps` adds only the search box. There is no "Show all" gate to hide behind.

So the consequence you name is real: silence the badge and a chronically failing app renders as nothing in a window that reserves nothing for "up to date". That is the failure #259 existed to remove, and re-introducing it for the rows that have been broken longest would be the worst possible subset to pick.

I also verified the distinction the whole decision rests on, since it is asserted in three comments: the badge's button is `actions.retry()`, bound at `MenuContentView.swift:969` and `WorkbenchWindowView.swift:1009` to `model.retry(result)` — one app. The banner's is `retryFailedChecks()`, which re-checks every failed row. They are genuinely different affordances, so "suppress the global nag, keep the per-row fact" is a coherent position and not a rationalisation of an accident.

**Mutant, run independently:** flipped `>=` to `>` in `isChronic` → `"chronic starts exactly at the threshold, not one round early or late"` failed. Restored → green. The boundary is genuinely pinned, which is what makes this a decision rather than a shrug.

Merging.

PR #281, opened 2026-09-02
Row gallery: measure translated strings, verify tooltip-differentiator claims

### LINKED ISSUE #263: The row gallery renders English only, so string overflow is invisible to it
`make gallery` pins `AppleLanguages='(en)'` and `AppleLocale='en_US'`
(`scripts/row-state-gallery.sh`), because `ImageRenderer` follows the host's
language and a run in another locale would rewrite all 68 tiles with a diff that
says nothing about the change under review.

The cost is that **the one failure this area has actually shipped is structurally
invisible to the sheet**: translated strings overflowing their slot. See memory
`duo-updater-status-line-slot-width` — the status line left 9pt in Spanish. The
2026-09 language harness existed precisely to catch this class and was torn down
each time.

Two candidate shapes:

1. Render a second pass in a language known to be long (ru or de) into
   `verify/row-states-ru/`, committed the same way. Doubles the tile count and the
   diff, but the sheet then covers the thing it currently looks like it covers.
2. Measure instead of render: assert each tile's drawn content stays inside the
   slot across languages, and commit the measurements rather than the images.

Also unresolved and related: `.help()` text is invisible in a PNG by construction.
Seven of the `mayLookAlike` exemptions are justified as "only the tooltip differs",
covering 13 of 34 states — for those, the evidence that the two windows agree is
"the pixels match and the invisible part is promised to differ". A text-level
check over the two views' help strings would close that.

Found by the second adversarial review on #259.

jizhi0v0:
## Shape chosen: measure, not render (option 2)

`make gallery` pins `AppleLanguages='(en)'` on purpose — `ImageRenderer` follows
the host language, so a run in any other locale rewrites all 80 committed PNGs
with a diff that says nothing about the change under review. That's also why the
sheet structurally can't see the one failure this area has actually shipped: a
translated string overflowing its slot (`CLAUDE.md`: the status line left 9pt in
Spanish; the workbench once drew a bigger font with no
`lineLimit`/`minimumScaleFactor` guard, and Russian's `Rate-limited` would have
pushed the app name out).

I took option 2 over option 1 (a second rendered sheet in ru/de) because a second
picture still needs a human to *notice* overflow — the same weakness that let the
original bug ship — and it doubles the committed diff on every future row change.
A measurement asserts the actual failure mode and fails the build by itself,
which is the whole value of the deliverable.

## What the check actually measures, precisely

`App/RowStateWidthCheck` reads `App/Resources/Localizable.xcstrings` directly and
measures each row-action string's **natural (unconstrained) width**, with the
real font/control each call site uses — `NSAttributedString` for a `Text` tag, a
real `NSButton` at the call site's `controlSize` for a button (same technique
memory `duo-updater-status-line-slot-width` used, and the one PR #267's own width
harness used) — against `RowStateGalleryCases.tileWidth` (320pt, the box every
committed PNG tile is drawn into).

This is **not a simulation of either window's real layout**. `PopoverRowAction`'s
real neighbour in `MenuContentView` is a `minWidth` reservation that concedes
width to a wide button rather than clipping it; `WorkbenchRowAction`'s only
production home (`WorkbenchWindowView.DetailHeader`) sits after a `Spacer()` with
~900pt of window behind it. Nothing in this codebase hard-clips at 320pt today. A
pass proves: this string, in this language, isn't asking for more room than the
English baseline every committed screenshot already treats as normal — which is
exactly the previously-invisible half of the failure class. `App/RowStateWidthCheck/main.swift`'s
header spells this out in full, including why the report also prints growth vs.
English even when nothing fails on it (there's no non-arbitrary "acceptable
growth" number to gate on, but the magnitude is worth a human's eyes).

**Why it reads the string catalogue directly instead of rendering the real views
with `AppleLanguages` pinned** (which I tried first, following `RowStateGallery`'s
own technique): it doesn't work. A `type: tool` product is a bare Mach-O, not an
`.app` bundle, and never gets `Localizable.xcstrings` compiled into it — confirmed
by giving the target an explicit `Resources` build phase + `GENERATE_INFOPLIST_FILE`
and finding no `CompileXCStrings`/`.lproj` step anywhere in the build log, and
separately by a self-check inside that first version (comparing
`Bundle.main.preferredLocalizations.first` against the requested language) that
caught its own strings resolving to English regardless of `AppleLanguages`. Full
story in `main.swift`'s header comment.

## Headroom, disclosed (post-review update)

Coordinator review found the OVERFLOW gate has generous headroom against the
widest real string and asked me to state the number plainly, since the original
PR body didn't. It does — and the number the review cited (136.1pt) wasn't
actually the widest, so I re-measured before responding: the true widest string
across en/ru/fr is `workbench/Not supported on this Mac [label:callout]` at
**194.5pt (ru)**, against the 320pt budget — **125.5pt of headroom, ~1.65x**.
That's the number now stated in `App/RowStateWidthCheck/main.swift`'s header,
and in every run's console output.

A gate with that much slack can't fire on anything realistic. I considered
tightening it (tie the budget to production layout, or fail on a relative growth
threshold) and rejected both — full reasoning and the redone mutation test are in the PR
comments below. Landed instead on: the committed
`verify/row-state-widths/*.txt` files are the actual deliverable (a translation
change is now a reviewable numeric diff, the same role the PNGs play for
`make gallery`), a new `growth_ratio` column makes that diff legible, and
anything growing 2.0x or more over English is echoed to console as
`HIGH GROWTH` — informational, not gating. It currently surfaces 8 real,
already-shipped strings (ru `Rate-limited`/`Verifying`, fr `Verifying`/
`Installing`, each on both surfaces) — not false positives to explain away, just
surfaced instead of hidden behind a 320pt pass. `no string exceeds the 320.0pt
tile width` no longer stands alone as if it meant "overflow is guarded" — every
run now also states the widest string and remaining headroom.

## Languages: ru and fr, not all six

Measured 2026-09-02 against the real `Localizable.xcstrings` values for every
row-action catalogue key, at both fonts these views use (`.caption2`
popover, `.callout` workbench), via `NSAttributedString`. Average width delta vs.
English:

| lang | avg Δ | notes |
|---|---|---|
| **ru** | **+41.0pt** | worst average AND worst single case: `Rate-limited` +119.3pt — the exact string `CLAUDE.md` already names |
| **fr** | **+25.5pt** | second-worst average, narrowly ahead of de; its own parenthetical already caused a real bug (fixed in #267) |
| de | +24.5pt | |
| es | +16.1pt | |
| ja | +6.7pt | net *narrower* on several keys (compact CJK glyphs); a few multi-kana words still run wide |
| zh-Hans | −17.8pt | narrower than English on every measured key — no signal |

ru and fr are the two that actually threaten this layout; de trails fr by 1pt on
average but neither de nor es tops ru or fr on any key measured. Including ja/zh-Hans
would inflate the report for two languages that have never once been the wide one.

## Injected-overflow demonstration (mutation test)

Inflated ru's `Rate-limited` value in `Localizable.xcstrings` to 3× its length,
reran `scripts/row-state-width-check.sh`:

```
ru: OVERFLOW — 2 string(s) exceed the tile width:
  popover/Rate-limited [tag:caption2] = "Ограничение частоты запросов Ограничение частоты запросов Ограничение частоты запросов" (497.4pt > 320.0pt)
  workbench/Rate-limited [tag:callout] = "Ограничение частоты запросов Ограничение частоты запросов Ограничение частоты запросов" (572.1pt > 320.0pt)
```
exit 1. Reverted (`git checkout -- App/Resources/Localizable.xcstrings`), reran:
```
ru: no string exceeds the 320.0pt tile width
```
exit 0.

## Second half of the issue: `.help()` comparison — done, not filed

The issue also flags that seven of `mayLookAlike`'s pairs are justified as "only
the tooltip differs", a claim nothing verifies since `.help()` text doesn't exist
in a PNG. I assessed the cost before building anything, per the issue's own "don't
half-build it": a throwaway probe found `.help(...)` text recoverable via plain
`Mirror` reflection — SwiftUI wraps a `.help()` call in a private
`HelpView<Content>` type carrying the help text as a `Text` — with **no
accessibility tree and no window**, as long as `.body` is called explicitly on the
concrete view type first (`Mirror` only sees stored properties, not a computed
`body` — an early version reflected the unevaluated struct and always found
nothing). That's small enough to just do, so I did: `RowStateGallery/main.swift`
gained `collectHelpTexts` (the reflection walk) and
`tooltipsDifferentiateExemptedPairs`.

**Re-reading the seven pairs' own comments first mattered**: only two actually
claim a tooltip differentiator ("the help text says which one" / "the tooltip is
what separates them" — the 10-vs-11 and 13-vs-19 pairs). The other five claim the
opposite — deliberately identical end-to-end ("one button", "a tile cannot show
which explanation appears", "the same marker … never reads like something we
could update ourselves"). Running the check against all seven first confirmed
exactly that for the five — a real discovery, not a bug, and folding them into the
same gate would have turned a correct design decision into a permanent build
failure. `tooltipDifferentiatedPairs` scopes to just the two real claims, with a
subset check against `mayLookAlike` so it can't silently drift if a pair is ever
renamed there.

Mutation-tested: gave the 10-vs-11 popover pair identical `.help()` text —
```
TOOLTIP DOES NOT DIFFERENTIATE (pixels match AND every collected .help() string matches too — the exemption's own justification is false): popover: 10-relaunch-to-apply-staged vs 11-restart-to-apply (tooltip set: ["MUTATION TEST: identical tooltip"])
```
exit 1. Reverted:
```
4 mayLookAlike pair(s) claiming a tooltip differentiator checked — every one actually has a different tooltip on each side
```
exit 0.

Fragile by construction — `HelpView`'s internal shape is an undocumented SwiftUI
implementation detail Apple can reshape on any OS/Swift update with no compiler
error. `.extractorBroken` distinguishes "nothing has a tooltip anymore" from "the
reflection technique itself broke", so that would fail loudly instead of as a wall
of unrelated-looking false positives.

## What remains invisible

- The five `mayLookAlike` pairs that are deliberately identical (including
  tooltip) still rest on the same "pixels match" pixel gate as before — correctly,
  since they're not claiming a tooltip differentiator to verify.
- `.help()` text on states OUTSIDE `mayLookAlike` (i.e. states that already render
  different pixels) is still unchecked — this PR only closes the gap for pairs
  whose safety claim specifically depends on the tooltip.
- Dynamic/interpolated help strings (`"Update \(version)"`, the App Store
  deep-link help, restart/relaunch help text) are out of scope for the width
  check — no single stable English catalogue value to measure. Named, not silently
  dropped, in `MeasuredString.all`'s doc comment.
- The width check's 320pt budget is honest about not being a literal production
  hard-clip (spelled out in `main.swift`) — a pass means "not wider than any
  already-reviewed tile", not "guaranteed to never look cramped".

## Verification

```
  # 1927 tests, 152 suites, 1 known issue, all passing (46.6s)
swift test --package-path CLI
  # 184 tests, 24 suites, passing
python3 scripts/check_localizable_keys.py --derived-data /tmp/duo-loc-check-263
  # ✓ 431 keys, all reachable
  # ✓ clean
  # ✓ clean
  # 45 tests, passing
GALLERY_DD=/tmp/duo-gallery-263 bash scripts/row-state-gallery.sh
  # rendered 80 states, no blank, no collision,
  # 4 mayLookAlike pair(s) claiming a tooltip differentiator checked — every one differs
GALLERY_DD=/tmp/duo-gallery-263 bash scripts/row-state-width-check.sh
  # en/ru/fr: no string exceeds the 320.0pt tile width
```

`git status --short verify/` is clean except one pre-existing, unrelated
nondeterminism: re-running the (untouched-by-this-PR) render code twice re-encodes
`verify/row-states/workbench/04-just-updated.png` a few bytes different, byte-for-byte
identical source both times — the same tile and cause PR #272's own body documented
and reverted rather than carry. Reverted here too; not something this PR touches or fixes.

Closes #263

### DIFF
--- a/App/RowStateGallery/Cases.swift
+++ b/App/RowStateGallery/Cases.swift
@@ -94,9 +94,24 @@ enum RowStateGalleryCases {
                     state: state, result: result,
                     downloadReadout: popoverDownloadReadoutOverrides[name] ?? .barAndPercent,
                     showsStageLabel: popoverShowsStageLabelOverrides[name] ?? { _ in true })
-                .frame(width: 320, height: 44, alignment: .trailing))
+                .frame(width: tileWidth, height: tileHeight, alignment: .trailing))
         }
     }
+
+    /// The box every ordinary row tile is drawn into — named rather than left as
+    /// a bare `320`/`44` so `RowStateWidthCheck` (#263) can measure a state's
+    /// UNCONSTRAINED width against the exact same number instead of inventing its
+    /// own budget. Not a literal SwiftUI hard clip anywhere else in the app:
+    /// `PopoverRowAction`'s real neighbour in `MenuContentView` is a `minWidth`
+    /// reservation that concedes width to a wide button rather than clipping it,
+    /// and `WorkbenchRowAction`'s only home (`WorkbenchWindowView.DetailHeader`)
+    /// sits after a `Spacer()` with roughly 900pt of window behind it. What DOES
+    /// hold at this size: every tile in the committed English sheet was drawn at
+    /// exactly this width, so it is the widest any state's content has ever
+    /// actually been reviewed at — see `RowStateWidthCheck/main.swift`'s doc
+    /// comment for what exceeding it does and doesn't prove.
+    static let tileWidth: CGFloat = 320
+    static let tileHeight: CGFloat = 44
 
     /// The workbench has no equivalent of "open the popover's explanation panel" —
     /// it points at the popover for that (see the file-level doc comment on
@@ -111,7 +126,7 @@ enum RowStateGalleryCases {
         }
         return AnyView(
             WorkbenchRowAction(state: state, result: result)
-                .frame(width: 320, height: 44, alignment: .trailing))
+                .frame(width: tileWidth, height: tileHeight, alignment: .trailing))
     }
 
     /// Each case names the row it is drawn against, because several states only
--- a/App/RowStateGallery/main.swift
+++ b/App/RowStateGallery/main.swift
@@ -153,7 +153,209 @@ func render() {
               + identical.joined(separator: ", "))
         failed = true
     }
+
+    // #263: TWO of `mayLookAlike`'s seven pairs are justified in their own comment
+    // as differentiated BY the tooltip ("the help text says which one" for
+    // 10-vs-11, "the tooltip is what separates them" for 13-vs-19) — a claim
+    // nothing checked, because `.help()` text is invisible in a PNG by
+    // construction. This verifies those two rather than trusting them: collect
+    // every `.help()` string reachable from each side (via `collectHelpTexts`'s
+    // Mirror-reflection walk — see its own doc comment for the technique and its
+    // risk) and require the two sets to differ.
+    //
+    // Deliberately NOT applied to the other five pairs (15-vs-28, 18-vs-22,
+    // 17-vs-23, 27-vs-31, 29-vs-32): their own comments claim the OPPOSITE —
+    // "one button", "a tile cannot show which explanation appears", "the same
+    // marker … never reads like something we could update ourselves" — i.e. those
+    // are deliberately identical end-to-end, tooltip included, not "differentiated
+    // by a tooltip nothing here checks". Running this check against them found
+    // exactly that (identical tooltips) on a first pass — a real discovery, but not
+    // a bug: verified against `RowActionViews.swift`, all three ARE the same
+    // control/wording by design. Folding them into `tooltipDifferentiatedPairs`
+    // would have turned a correct design decision into a build failure.
+    switch tooltipsDifferentiateExemptedPairs() {
+    case .ok(let checked):
+        print("\(checked) mayLookAlike pair(s) claiming a tooltip differentiator"
+              + " checked — every one actually has a different tooltip on each side")
+    case .undifferentiated(let names):
+        print("TOOLTIP DOES NOT DIFFERENTIATE (pixels match AND every collected"
+              + " .help() string matches too — the exemption's own justification is"
+              + " false): " + names.joined(separator: ", "))
+        failed = true
+    case .extractorBroken:
+        // Distinguished from `.undifferentiated` on purpose: if Mirror found NO
+        // `.help()` text anywhere in the whole run, that is far more likely
+        // SwiftUI's private internal shape having changed out from under
+        // `collectHelpTexts` (undocumented by Apple, not guaranteed stable across
+        // an OS/Xcode update) than every single row in the app suddenly losing
+        // its tooltips. Reporting it as its own failure, rather than as N
+        // `.undifferentiated` pairs, keeps a real extractor break from reading
+        // like a wall of unrelated view regressions.
+        print("TOOLTIP EXTRACTOR FOUND NOTHING ANYWHERE: either every `.help()` in"
+              + " the app was removed, or collectHelpTexts's Mirror-reflection"
+              + " technique no longer matches this SwiftUI version's private"
+              + " HelpView<Content> shape — see that function's doc comment.")
+        failed = true
+    case .driftedFromMayLookAlike:
+        print("TOOLTIP CHECK SCOPE DRIFTED: tooltipDifferentiatedPairs names a pair"
+              + " that is no longer in RowStateGalleryCases.mayLookAlike — update"
+              + " (or remove) it to match.")
+        failed = true
+    }
+
     if failed { exit(1) }
+}
+
+/// Result of checking every `mayLookAlike` pair's tooltip text against its pixel
+/// exemption. `.extractorBroken` is kept distinct from `.undifferentiated([...])`
+/// — see the call site for why that distinction matters.
+enum TooltipCheckResult {
+    case ok(checked: Int)
+    case undifferentiated([String])
+    case extractorBroken
+    case driftedFromMayLookAlike
+}
+
+/// `.body` for `PopoverRowAction`/`WorkbenchRowAction` at a state's DEFAULT
+/// `downloadReadout`/`showsStageLabel` — the same defaults `popoverTile`'s
+/// non-explanation branch uses. Deliberately NOT `RowStateGalleryCases`'
+/// `popoverTile`/`workbenchTile`: those wrap the constructed view in `.frame(...)`
+/// and `AnyView` for the PNG pass, and `Mirror` only sees a view's STORED
+/// properties — it cannot see through an unevaluated `body` computed property, so
+/// reflecting the wrapped-but-unevaluated struct finds nothing at all (confirmed:
+/// an earlier version of this function did exactly that and always reported
+/// `.extractorBroken`). Calling `.body` explicitly, on the concrete type, forces
+/// SwiftUI to evaluate the `switch state { … }` into its actual primitive tree
+/// (`Button`, `HelpView`, `_ConditionalContent`, …) BEFORE `Mirror` ever sees it —
+/// which is also why this can't be generalized to walk an arbitrary `AnyView`:
+/// it relies on knowing the concrete type at the call site.
+@MainActor
+private func rowActionHelpTexts(surface: String, state: RowActionState, result: UpdateResult) -> [String] {
+    switch surface {
+    case "popover": return collectHelpTexts(PopoverRowAction(state: state, result: result).body)
+    case "workbench": return collectHelpTexts(WorkbenchRowAction(state: state, result: result).body)
+    default: return []
+    }
+}
+
+/// The subset of `RowStateGalleryCases.mayLookAlike` whose OWN comment claims a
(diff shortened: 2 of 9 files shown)


jizhi0v0:
Addressed. Went with your option 3, with one correction and one addition.

**Correction to the headroom number first — I re-measured before acting on it.**
The widest string is not `workbench/Verifying [tag:callout]` at 136.1pt. It's
`workbench/Not supported on this Mac [label:callout]` at **194.5pt (ru)** —
already present in the `ru.txt` you reviewed, just not the one that got read as
the max. Real headroom against the 320pt budget is **125.5pt, ~1.65x**, not 2.4x.
Still real slack, still couldn't have caught a 9pt Spanish near-miss, but it's
the right number now — `App/RowStateWidthCheck/main.swift`'s header states
194.5pt/125.5pt explicitly, and I want the correction on the record rather than
silently fixing it, since it means the original review claim shouldn't be
repeated elsewhere.

**Why option 3 over 1 or 2, given the corrected number:**

- **Option 1** (derive a budget from `MenuContentView`'s real layout) doesn't
  resolve to a real number for these two tiles specifically. `nameFits`/
  `trailingSlot` trade the trailing control's width against the NAME column's
  width, and the name has no fixed minimum — it's an arbitrary app name. There's
  no fixed budget to derive without inventing one.
- **Option 2 as a hard gate** doesn't survive contact with the real data: at any
  round, non-cherry-picked relative threshold worth using (I checked 2.0x
  before looking at what it would catch), **8 already-shipped strings today**
  exceed it — ru `Rate-limited`/`Verifying` and fr `Verifying`/`Installing`, each
  on both surfaces. Gating on that would mean either fixing translations this
  issue has no business touching, or building an exemption-list mechanism under
  time pressure — the scope-expansion you told me not to do.

So: the committed `.txt` files are the primary deliverable (reviewable numeric
diff, same role the PNGs play for the gallery), `growth_ratio` is now a column
in every row, and anything ≥2.0x is echoed to console as `HIGH GROWTH` —
**informational, not gating**. It currently surfaces exactly the 8 strings
above, including ru `Rate-limited` at 2.7x — the string CLAUDE.md already names
as risky. The OVERFLOW gate's threshold is unchanged (320pt); its pass message
now states the widest string and remaining headroom every run instead of
implying a tight bound.

**Mutation test redone against the updated code** (threshold itself didn't
change, but the report format and messaging did, so re-verified both):

Inflated ru's `Rate-limited` to 3x length:
```
ru: OVERFLOW — 2 string(s) exceed the tile width:
  popover/Rate-limited [tag:caption2] = "...×3..." (497.4pt > 320.0pt)
  workbench/Rate-limited [tag:callout] = "...×3..." (572.1pt > 320.0pt)
ru: HIGH GROWTH (...): popover/Rate-limited ... (8.18x English's 60.8pt → 497.4pt), workbench/Rate-limited ... (8.27x ...)
```
exit 1. Reverted:
```
ru: no string exceeds the 320.0pt tile width (widest: workbench/Not supported on this Mac [label:callout] at 194.5pt — 125.5pt of headroom left in that budget)
ru: HIGH GROWTH (...): [the same 4 real entries as before]
```
exit 0.

Commit `3349171`. `make gallery` and `make width-check` both still exit 0 with
zero unexplained diff (same pre-existing `workbench/04-just-updated.png`
nondeterminism as before, reverted).

jizhi0v0:
You are right and I was wrong: I sorted a tab-separated file with `sort -t'|' -k2`, which put the whole line in field 1 and made `-k2` empty, so the ordering was arbitrary and I read 136.1pt off whatever line happened to land first. Re-measured properly:

```
sort -t$'\t' -k2 -rn verify/row-state-widths/ru.txt | head -3
workbench/Not supported on this Mac [label:callout]  194.5  320.0  ok  +43.9  1.29
workbench/Rate-limited [tag:callout]                 188.4  320.0  ok  +119.3 2.73
workbench/Relaunch now [button:regular]              168.0  320.0  ok  +59.0  1.54
```

194.5pt widest, 125.5pt headroom, ~1.65x — your numbers. Checking the figure before building on it was the right call, and it is the correct instinct in general: a finding handed down by a reviewer is still just a claim.

Option 3 with the evidence you gave is the right landing. The part that convinces me is not the reasoning but the data: **eight already-shipped strings exceed any round relative threshold today**, so a hard growth gate would have shipped red, and the only ways out are fixing translations (out of scope) or inventing an exemption mechanism (scope creep I explicitly told you not to do). Picking 2.0x before looking at what it would catch is the right order to do it in.

What makes this a real deliverable rather than a green light nobody has seen fail: the `growth_ratio` column is committed, so a translation ballooning shows up as a reviewable numeric diff exactly the way a rendering change shows up as a PNG diff — and ru `Rate-limited` at 2.73x is already sitting there flagged, which is the exact string `CLAUDE.md` names as the one that would push the app name out. The pass message stating widest-observed and headroom every run is what keeps it from reading as a tight bound.

I also note `NO TRANSLATION FOUND (measured English as a stand-in…)` — that is the right shape for the failure you discovered mid-task. A tool that silently measures English in every language is exactly the kind of green check this whole issue exists to prevent.

Merging.

PR #284, opened 2026-09-02
Gallery: audit the gallery's own exemptions

### LINKED ISSUE #269: notFaithful under-reports: 8 committed tiles carry the ImageRenderer placeholder and say nothing

### LINKED ISSUE #271: mayBeBlank has no dead-exemption gate, so a stale entry silently disables the blank detector
`make gallery` fails the build on a dead `mayLookAlike` exemption. `mayBeBlank`
has no equivalent check, so an exemption there that stopped applying is silently
kept — and it disables the blank detector for that tile forever.

Where the asymmetry is, in `App/RowStateGallery/main.swift`:

```swift
// blank gate — an exemption is consulted, never audited
if isBlank, !RowStateGalleryCases.mayBeBlank.contains("\(surface)/\(name)") { … fail … }

// look-alike gate — exemptions ARE audited
let deadExemptions = RowStateGalleryCases.mayLookAlike.subtracting(drawnPairs)
if !deadExemptions.isEmpty { … fail … }
```

The reasoning `CLAUDE.md` already gives for the `mayLookAlike` gate transfers
word for word: a hand-maintained exemption that no longer matches anything is a
free pass issued to future drift, and the person who made the view stricter is
the one who should retire it.

This is not hypothetical — #260 just walked into exactly the situation. The
workbench used to draw `EmptyView` for `.upToDate`, so `workbench/30`, `/31` and
`/32` were all in `mayBeBlank`. #260 made `.upToDate(channel:)` draw the App
Store icon and the TestFlight tag, so `/31` and `/32` stopped being blank. They
were removed by hand and correctly — but had they been left in, **the build would
have stayed green** and two tiles would have been permanently exempt from the one
check that exists to catch a state drawing nothing.

Fix is the same shape as the existing gate: collect the surface-qualified names
that actually rendered blank during the run, subtract them from `mayBeBlank`,
fail on the remainder with the same message style.

One wrinkle worth deciding while implementing: three of the six current entries
(`workbench/38`/`39`/`40`, the explanation panels from #265) are blank because
`workbenchTile` deliberately returns `EmptyView()` for them — they will always
match, so they are never at risk of going dead. `workbench/30-up-to-date` is the
one that could. That does not change the fix; it just means the gate's value is
concentrated in the entries that describe a *view's* behaviour rather than the
harness's.

Found while reviewing #270.

jizhi0v0:
Same bug in two places: `App/RowStateGallery/` keeps three hand-maintained exemption lists, and only `mayLookAlike` was ever audited for a dead entry. `mayBeBlank` (#271) and `notFaithful` (#269) were only ever *consulted* — a stale or missing entry in either one stayed silently in effect. `CLAUDE.md`'s justification for the `mayLookAlike` audit — a hand-maintained exemption nobody checks is a free pass for future drift, and whoever tightened the view is the one who should retire it — transfers to both word for word.

**Edit:** the last commit (`Gallery: widen the placeholder scan's green floor from 170 to 185`) fixes a margin an independent review found the first version of this PR didn't measure — see "Threshold" below. Everything else described here is unchanged by it.

## #271 — `mayBeBlank` dead-exemption gate

Mirrors the existing `mayLookAlike` audit exactly: collect every surface-qualified name that actually rendered blank this run (exempted or not), subtract `mayBeBlank` from that set, fail on the remainder with the same message shape.

One entry can never go dead by construction: `workbench/38`/`39`/`40` are blank because `workbenchTile` deliberately returns `EmptyView()` for the three explanation-content names — no view change can make those draw something. `workbench/30-up-to-date` is the one entry whose blankness depends on how `WorkbenchRowAction` draws `.upToDate(channel: .none)`, which is exactly the case #270 changed for `/31` and `/32` (removed by hand, correctly — but nothing would have failed the build if they'd been left in).

## #269 — `notFaithful` under-reporting

Implemented the issue's own first proposal: scan every rendered tile for `ImageRenderer`'s placeholder signature and fail the build when a tile carries it and isn't registered.

Re-measured rather than trusting either number in play: the issue's table (11 of 3) predates #268's two `ProgressView` registrations and #265's ring-readout cases. Scanning the current 80-tile committed sheet found **15** tiles carrying the placeholder, only **5** registered — 10 unregistered, none of which showed up in the issue's original table (`02/05/06/07` on both surfaces predate #265 entirely; `workbench/35`/`36` didn't exist yet when the issue was filed).

Registered each with a reason, split into three groups rather than one comment, because two different things distinguish them — not just the two known causes, but whether the *other* surface has a faithful twin to read instead:

1. **SF Symbol in `.buttonStyle(.borderless)`** (`popover/18`, `/21`, `/22`, already registered) — workbench draws the same states correctly via `Label`. Escape hatch: read the workbench tile.
2. **Plain `ProgressView()`/`ProgressView(value:)`, no escape hatch anywhere** — `popover` and `workbench` `02/05/06/07` plus the already-registered `37`. Both windows reach for a real `ProgressView` for every non-ring install stage, so there is no faithful picture of an installing row on either surface.
3. **`ProgressView(value:)`, workbench-only** — `workbench/35`/`36`. `PopoverRowAction.downloadProgress` draws the ring readouts with a custom `ProgressRing`, not a `ProgressView`, so the popover half of these two states is genuinely correct. Escape hatch exists, just on the opposite surface from group 1.

### Threshold: two margins, both measured

There are two independent margins here, and an earlier revision of this PR only measured one of them.

**Pixel-count margin.** Scanned all 80 committed tiles for R>230, G∈[window], B<70 pixels. At the shipped window (green floor 185) the result is an exact bimodal split, nothing in between:

| | count |
|---|---|
| tiles at exactly 0 matching pixels | 65 |
| tiles at ≥630 matching pixels | 15 |
| tiles anywhere from 1–629 | **0** |

`pixelsMatching > 40` has ~15–16× headroom on both sides of that gap.

**Colour margin — the one the first pass missed.** Review caught this and I reproduced it independently before changing anything: with the original green floor of 170, the pixel-count margin above was real but the *colour* margin protecting it was not. Every tile this app legitimately paints orange (`.tint(.orange)` Relaunch/Update buttons, amber `exclamationmark.triangle.fill`/`globe.badge.chevron.backward` warning `Label`s, the rate-limited badge) scored 0 in the window, but only because system orange's green channel tops out at ≈167–169 — 1–3 units below a floor of 170, on a colour Apple controls, not this codebase. A future system-orange nudge above that floor would flip hundreds of legitimate pixels into the window at once (single tiles already carry 465–1761 orange-ish pixels), failing unrelated tiles by the hundreds — the pixel-count headroom does nothing to stop a false positive that lands that far over 40, not near it.

Fix: moved the floor from 170 to **185**, roughly halfway between real orange's ceiling (≈167–169) and the placeholder's own centre (≈204, where the bulk of matching pixels sit exactly — the rest are anti-aliased edges spread down to ≈174). That's ~18 units of margin on each side instead of 3 and ~34. Cost: 1.5% of the placeholder's own matching pixels — the smallest affected tile drops from 638 to 630, still nowhere near the 40-pixel cutoff.

## Mutants (red → green, pasted)

**#271** — made `WorkbenchRowAction`'s `.upToDate(channel: .none)` draw `Text("MUTATION-TEST")` instead of `EmptyView()`, leaving `workbench/30-up-to-date` in `mayBeBlank` but no longer blank:

```
RED:
DEAD EXEMPTION (these no longer render blank — drop them from mayBeBlank): workbench/30-up-to-date
(exit 1)

GREEN (reverted):
no state renders blank
(exit 0)
```

**#269** — commented out `"popover/18-update-major-upgrade"` from `notFaithful`, re-run against the shipped 185 floor:

```
RED:
UNFAITHFUL TILE NOT REGISTERED (contains ImageRenderer's placeholder signature — see
RowStateGalleryCases.placeholderPixelCount — but is missing from notFaithful; register
it there with a reason, don't just silence this): popover/18-update-major-upgrade (630px)
(exit 1)

GREEN (reverted):
no unregistered ImageRenderer placeholder pixels
(exit 0)
```

That red run is also the negative-control: all 79 other tiles — including every orange-badge/button tile — passed the scan in the same run, so the same evidence covers both "the gate catches a removed entry" and "the gate doesn't fire on ordinary orange UI." (630px matches the independent re-count at the new floor exactly.)

## Verdict on the `ProgressView()` stand-in

No — leave it as a documented placeholder, don't build a renderable stand-in.

The two causes are not symmetric. The borderless-button SF Symbol has a real escape hatch: the workbench draws the *identical* state correctly via `Label`, so reading that tile costs nothing. A hand-drawn arc for `ProgressView()` would have no such backing — it would be a shape this harness invented, existing nowhere in the shipped app, standing in for a control that *is* real but that `ImageRenderer` specifically cannot draw. That picture would look more trustworthy than the yellow placeholder while being strictly less connected to what ships: the placeholder at least announces "something is wrong here," a drawn stand-in announces nothing. The sheet's whole value, per its own file header, is "a change ... shows up as an image diff instead of being noticed in use — or not"; a fabricated spinner optimizes for the sheet looking complete over the sheet staying honest about what it can and can't show. `notFaithful` plus the new gate already gets the actually-useful outcome — the state coverage is verifiably real (no relevant branch renders blank, nothing collides) and the one thing that isn't real is named, tracked, and enforced.

## Verified

```
━ Test run with 1927 tests in 152 suites passed after 43.287 seconds with 1 known issue.

swift test --package-path CLI
✔ Test run with 184 tests in 24 suites passed after 0.075 seconds.

python3 scripts/check_localizable_keys.py --derived-data /tmp/duo-loc-269
✓ localizable keys agree — 431 in the catalog, all reachable

✓ version comparisons discriminate — 210 files, 2 ledger call(s) keyed on the build, 7 marketing-first pick(s), all display-only

✓ app audits consistent — 90 docs, all indexed, no machine inventory, no pointers into untracked evidence

Ran 45 tests in 0.005s — OK

GALLERY_DD=/tmp/duo-gallery-269 bash scripts/row-state-gallery.sh
rendered 80 states
no state renders blank
no unregistered ImageRenderer placeholder pixels
no two states draw the same tile
(exit 0)
```

`git status --short verify/` is empty — no view changed, so no committed PNG moves. (One flake surfaced repeatedly across every run in this branch, on both the mutated and unmutated code: `workbench/04-just-updated.png` differing by 4 pixels at a 1/255 channel delta, inside the "Updated" checkmark glyph — reproduced identically on plain `main` with none of this PR's code, so it's `ImageRenderer` anti-aliasing jitter, not a regression, and was reverted each time rather than committed.)

Closes #269
Closes #271

### DIFF
--- a/App/RowStateGallery/Cases.swift
+++ b/App/RowStateGallery/Cases.swift
@@ -356,6 +356,46 @@ enum RowStateGalleryCases {
         return true
     }
 
+    /// Counts pixels matching `ImageRenderer`'s "unavailable" placeholder — the
+    /// signature yellow it substitutes for content it cannot draw (see
+    /// `notFaithful`'s doc comment for the two things that triggers it): red
+    /// channel over 230, green between 170 and 225, blue under 70, all in 0...255
+    /// (this compares against `NSColor`'s 0...1 components instead).
+    ///
+    /// Threshold and window are #269's, re-measured directly against the
+    /// committed sheet rather than trusted from the issue: scanning all 80
+    /// committed tiles found an exact bimodal split with nothing in between — 65
+    /// tiles at 0 matching pixels, 15 at 638 or more, no tile anywhere from 1 to
+    /// 637. The 0-pixel group includes every tile this app legitimately paints
+    /// orange (the `.tint(.orange)` Relaunch/Update buttons, the amber
+    /// `exclamationmark.triangle.fill`/`globe.badge.chevron.backward` warning
+    /// `Label`s, the rate-limited badge) — none of them land in the window
+    /// because system orange's green channel (~141–153 across the samples
+    /// checked) sits below it; the placeholder's is ~204. So the separation is
+    /// structural, not a threshold picked to happen to clear this sheet: a
+    /// `pixelsMatching > 40` cutoff has roughly 16× headroom on both sides of the
+    /// actual gap (638 down to 40, 40 up from the observed max-non-match of 0).
+    static func placeholderPixelCount(_ rep: NSBitmapImageRep) -> Int {
+        var count = 0
+        for x in 0..<rep.pixelsWide {
+            for y in 0..<rep.pixelsHigh {
+                guard let c = rep.colorAt(x: x, y: y) else { continue }
+                if c.redComponent > 230.0 / 255.0,
+                   c.greenComponent >= 170.0 / 255.0, c.greenComponent <= 225.0 / 255.0,
+                   c.blueComponent < 70.0 / 255.0 {
+                    count += 1
+                }
+            }
+        }
+        return count
+    }
+
+    /// A tile with more matching pixels than this is presumed to be carrying the
+    /// placeholder, not a few incidental orange pixels — see
+    /// `placeholderPixelCount`'s doc comment for the measurement behind the
+    /// number.
+    static let placeholderPixelThreshold = 40
+
     /// `DownloadReadout`'s declaration order IS the algorithm `AppRow.downloadReadout`
     /// runs — widest first, first fit wins (see that property's doc comment). The
     /// gallery tiles added for #265 make the three cases look different, but a diff
--- a/App/RowStateGallery/Cases.swift
+++ b/App/RowStateGallery/Cases.swift
@@ -386,30 +386,43 @@ enum RowStateGalleryCases {
 
     /// Counts pixels matching `ImageRenderer`'s "unavailable" placeholder — the
     /// signature yellow it substitutes for content it cannot draw (see
-    /// `notFaithful`'s doc comment for the two things that triggers it): red
-    /// channel over 230, green between 170 and 225, blue under 70, all in 0...255
+    /// `notFaithful`'s doc comment for the two things that trigger it): red
+    /// channel over 230, green between 185 and 225, blue under 70, all in 0...255
     /// (this compares against `NSColor`'s 0...1 components instead).
     ///
-    /// Threshold and window are #269's, re-measured directly against the
-    /// committed sheet rather than trusted from the issue: scanning all 80
-    /// committed tiles found an exact bimodal split with nothing in between — 65
-    /// tiles at 0 matching pixels, 15 at 638 or more, no tile anywhere from 1 to
-    /// 637. The 0-pixel group includes every tile this app legitimately paints
-    /// orange (the `.tint(.orange)` Relaunch/Update buttons, the amber
-    /// `exclamationmark.triangle.fill`/`globe.badge.chevron.backward` warning
-    /// `Label`s, the rate-limited badge) — none of them land in the window
-    /// because system orange's green channel (~141–153 across the samples
-    /// checked) sits below it; the placeholder's is ~204. So the separation is
-    /// structural, not a threshold picked to happen to clear this sheet: a
-    /// `pixelsMatching > 40` cutoff has roughly 16× headroom on both sides of the
-    /// actual gap (638 down to 40, 40 up from the observed max-non-match of 0).
+    /// There are TWO margins here, not one, and the window's green floor exists
+    /// to protect the thinner of them:
+    ///
+    /// - **Pixel-count margin**: scanning all 80 committed tiles at this window
+    ///   found an exact bimodal split with nothing in between — 65 tiles at 0
+    ///   matching pixels, 15 at 630 or more (the smallest, `21-…-region-locked`,
+    ///   drops from 638 to 630 once the green floor below is raised — still true
+    ///   for every tile). A `pixelsMatching > 40` cutoff has roughly 15–16×
+    ///   headroom on both sides of that gap.
+    /// - **Colour margin**: real orange this app legitimately paints — the
+    ///   `.tint(.orange)` Relaunch/Update buttons, the amber
+    ///   `exclamationmark.triangle.fill`/`globe.badge.chevron.backward` warning
+    ///   `Label`s, the rate-limited badge — tops out at green ≈167–169 across
+    ///   every sampled pixel in the sheet; the placeholder's is ≈204, with the
+    ///   bulk of it (a strong majority of matching pixels) sitting at exactly
+    ///   204 and the rest (anti-aliased edges) spread down to ≈174. A floor of
+    ///   170 put only 1–3 units between real orange's ceiling and the window —
+    ///   thin enough that a future system-orange nudge could push those hundreds
+    ///   of legitimate pixels straight into the window and fail unrelated tiles
+    ///   by the hundreds, which the pixel-count margin above does nothing to
+    ///   stop (a false positive of that kind arrives WAY over 40). 185 sits
+    ///   roughly halfway between the two clusters — ~18 units above real
+    ///   orange's ceiling, ~19 below the placeholder's centre — trading 1.5% of
+    ///   the placeholder's own matching pixels (still nowhere near the 40-pixel
+    ///   cutoff on the smallest affected tile) for a margin that isn't the one
+    ///   thing standing between this gate and a false-positive storm.
     static func placeholderPixelCount(_ rep: NSBitmapImageRep) -> Int {
         var count = 0
         for x in 0..<rep.pixelsWide {
             for y in 0..<rep.pixelsHigh {
                 guard let c = rep.colorAt(x: x, y: y) else { continue }
                 if c.redComponent > 230.0 / 255.0,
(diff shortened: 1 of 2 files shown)


jizhi0v0:
Reviewed. I reproduced the central measurement with my own scanner rather than taking it from the body, and it holds: 80 tiles, 65 at zero, 15 carrying the placeholder, **nothing in between** — the split is genuinely bimodal, not a threshold someone picked.

**Independently run mutant:** dropped `workbench/06-installing-downloading` from `notFaithful` and re-ran the gallery →

```
UNFAITHFUL TILE NOT REGISTERED (…) workbench/06-installing-downloading (3638px)
```

— the same pixel count my own scanner returned for that tile — and green after restoring. The gate is wired, not just written.

**On the colour margin:** confirmed at the new floor. Real orange still scores exactly 0 on tiles that carry 465–1761 genuinely orange pixels (`popover/10`, `popover/11`, `workbench/21`, `workbench/22`), and all 15 placeholder tiles still register. Good call re-measuring your own ceiling (169) instead of taking my 167 — either way the old floor of 170 had 1–3 units of margin against a system colour, which is the part neither of us had looked at.

**One correction pushed as `e8e938a`** — two numbers in the doc comment, not the logic. It read "the smallest, `21-…-region-locked`, drops from 638 to 630", which conflates two tiles: 638 was 18/22's count at the *old* floor, and 21's was 658. Re-scanned at both floors: **21 goes 658 → 616** and is the minimum at the new floor; **18 and 22 go 638 → 630**. The 15–16× claim survives (616/40 = 15.4) — but a threshold comment is exactly what someone re-calibrating this window will trust, so the numbers should be the ones the scan returns.

Re-ran the full gallery after the correction: 80 states, all gates green, no PNG diff.

Merging. Both lists are now audited by the gate they exempt from, which is what `mayLookAlike` already had and the reason it kept catching things this week.

PR #311, opened 2026-09-03
Structured changelogs for Xcode, Antigravity (hub + IDE), AnyDesk, Headlamp and Helium

jizhi0v0:
Six changelog recipes for five apps that had nothing structured to show, plus the
one piece of core the last of them needed.

| app | before | now |
|---|---|---|
| AnyDesk | embedded page → Cloudflare challenge (403) | 80 macOS entries from the plain-text file its probe already reads |
| Antigravity | embedded page | 18 entries, newest 2.12.0 |
| Antigravity IDE | **no changelog link at all** | 30 entries, newest 2.5.5 |
| Headlamp | embedded page (`GitHubMarkdownParser` refuses its tables) | 18 entries, table era and bullet era |
| Helium | embedded page (feed has no notes, vendor has no changelog site) | 33 entries from the commit logs in each release |
| Xcode | embedded page → a 17 KB SPA shell with no note text | the whole 27 train, one entry per beta |

### Antigravity IDE: a comment that was wrong for a measurable reason

`VendorProbeRecipe` declines to link the changelog page from the IDE, recording it
as "JS-rendered — 88 KB with zero version strings in the served HTML". That was a
reading of a gzip stream: the host answers gzip even for
`Accept-Encoding: identity`, and 88 KB is the compressed size. Decoded, the page
is 401 KB of server-rendered markup with both products' releases in it. The two
recipes anchor on the product token in each row's release link, so neither can
read the other's versions — and they are 2.12.0 and 2.5.5 apart.

### Xcode: why this one needed core

A regex cannot read Apple's document. A note's visible text is an *array* of runs,
and 83 of the 335 notes on the live Xcode 27 page have more than one — so
"capture the text" captures the first run, and a quarter of the notes would end
mid-sentence at the first inline code span. The decoder joins the runs, resolves
a link to the words the document's own reference table gives it, and keeps each
note's section: "Known Issues" versus "New Features" is a level-4 heading and
nothing else, and flattened away every known issue reads as a shipped feature.

The URL needed its own template token because Apple spells the version its own
way (`26.6` → `26_6`, `27.0` → `27`). `{major}`, the closest existing token, maps
every 26.x onto the Xcode 26.0 page — and Apple ships betas for nearly every
minor (16.1–16.4, 26.1–26.5 all had them), so that is the common case, not a
corner. Every mapping is pinned against the real `links.notes.url` values in
`xcodereleases.com/data.json`.

### What `duo verify` caught that the fixture tests did not

Both GitHub-sourced recipes went green on fixtures and **red against the live
endpoints**, for the same reason twice: the fixture was one sample of a format
that has more than one form.

1. `api.github.com` serves the same document compact *and* pretty-printed, and
   which one arrives varied by request. The registry's other GitHub recipes never
   met this because they go through `Decodable`, which cannot see whitespace.
2. Helium ends its stable release bodies `\r\n` and its prerelease bodies `\n`.
   This one never failed loudly — a pattern that knows only the first yields an
   entry with *no changes*, which is invisible. It surfaced by mutating the
   prerelease gate and finding the test still green: the decoy was being skipped
   for being unreadable, not for being a prerelease. That test now relaxes the
   registered pattern's one gate and asserts the entry comes back.

### Verification

- `make test` — green (2019 core + 188 CLI tests, plus the four Python gates).
  One flake seen on an earlier run: `vendorResolvesInstallPlans`, a live-network
  test, during a stretch where 11 unrelated hosts were also unreachable from this
  machine; it passes on re-run and touches nothing in this change.
- `duo verify --only` on all five bundle ids: **6 changelogs ✓, 0 broken**.
- Full `duo verify` (306 checks): 0 broken. The one WARN (LibreOffice) and the
  INFRA lines are pre-existing / this machine's network.
- Mutation-checked: dropping AnyDesk's `(macOS)` anchor, widening Antigravity's
  IDE tab anchor, and reverting Helium's line-ending handling each turn a test
  red.

`verify/baseline.json` is deliberately untouched — it is the nightly sweep's to
write, and a local run would have recorded this machine's network.

Four of the five apps have no `docs/app-audits` entry to update; Helium's is
updated. Writing the other four is `/app-audit` work, not this change's.


jizhi0v0:
## Adversarial review round — 9 findings, all closed

Reviewed with a numbered attack-surface brief ("find what I missed", no builds).
I re-verified every "verified" finding against live bytes myself before acting.
**Three were already losing content on today's endpoints:**

| finding | measured | consequence |
|---|---|---|
| Headlamp bullet marker | 8 of 18 releases use `* `, not `- ` | those 8 rendered **nothing** and the recipe still reported success |
| Helium item capture | 5 commit subjects in the 40-release window contain `\"` | each cut in half at the backslash — the exact trap `postmanReleaseNotes` documents |
| Xcode 26.1.1 | 103 rows checked, **1 mismatch** | 404 → falls back to embedding Apple's page |

On the last one: Apple files 26.1.1 under `xcode-26_1` (a page whose title reads
"Xcode 26.1.1 Release Notes") while filing 26.0.1 and 26.4.1 under their own
pages, so no mapping satisfies both. My doc comment had claimed verification
against 30 entries; it now states the real population, names the exception, and
says the consequence is the pre-PR behaviour for that one release. Pinned in a
test so it's a recorded measurement, not something to rediscover.

Four latent ones, each now pinned by a test: Antigravity's unfenced run to the
`<h3>` (a heading-less row would have taken the next row's notes, and the last
hub row would have reached into the IDE panel); the Apple decoder returning half
a changelog when a page has no title (now fails closed); DocC key-name collisions
turning one unknown block into a nil changelog (per-field decode now, so the
"unrecognized blocks are skipped" comment is true); and the heading that opens an
entry being reused as that entry's section label.

Two more: the helm-chart decoy in Headlamp's fixture had an empty body, so it was
excluded for having no items rather than by the tag anchor the test credits — it
carries a real bullet now; and `codeListing` inside a list item is dropped, which
was silent and is now stated where the decision lives.

**Every fix is mutation-checked** — reverting it turns a named test red. Verified
after: `swift test` green, the four Python gates green, `duo verify --only` green
on all five bundle ids.

The review also cleared four items I asked it to attack and found nothing:
`parserGeneration` (correct not to bump — no cached entry can exist for these
bundle ids), opt-in safety of the new template token and `StructuredFormat` case,
GitHub field-order (stable, and fails to *dropped* entries rather than merged
ones if it changes), and the widened `markdownSource` guard.

### Two things deliberately not done

- **No `CHANGELOG.md` entry.** v0.3.81 is tagged, so there is no unreleased
  section to write into, and creating one would conflict with every other PR in
  flight — the recipe PR merged earlier today did the same. This is a rollup at
  release time; flagging it here so it isn't missed.
- **No audit docs for AnyDesk, Antigravity (×2), Headlamp or Xcode.** None exist
  for those apps, and writing one is `/app-audit` work — a partial doc that looks
  like a completed audit is worse than none. Helium's, which does exist, is
  updated.

PR #316, opened 2026-09-03
Turn the download ledger into a queryable request log

jizhi0v0:
Duo Updater is a background updater, which makes it a plausible candidate
for the heaviest network consumer on the machine — and it owed the person
running it an answer. This turns the download ledger into a general event
log and puts both halves of that answer in one window.

## What you get

**A Network window with two tabs.** *Downloads* is the per-app ledger,
unchanged: what each update cost as a file on disk. *Requests* is the log
itself — every fetch the updater made, with the figures above it answering
whatever the field is currently asking rather than showing a fixed
lifetime total.

They share a window because they are one question at two altitudes: the
same transfer counted as a file and as bytes across a socket. Met in two
different places, a reader takes one of the two figures for a bug.

**A filter language**, one grammar for the window, its chips and
`duo events --filter`: `host:` `app:` `purpose:` `status:` `client:`
`size>` `took>`, bare words matching host and path, a repeated key meaning
either. Filters commit to removable capsules; a bare word stays in the
field and narrows as you type.

**`app:` is new information.** `objects.githubusercontent.com/…/asset/12345`
names no app and neither does a vendor's `/api/v1/latest`, so the log could
say what was fetched and never what for. Requests now carry the app they
were made for, by way of a task-local set where the app is actually known.

## What it found on the way

**Spotify's update check was re-downloading a 1.87 MB installer every
time.** `URLCache` silently declines to store a response larger than about
5% of its capacity, so at 16 MB that body never fit; with nothing cached,
`reloadRevalidatingCacheData` had no validator to send. 92 checks on one
machine came to 160 MB — the largest update-check cost by an order of
magnitude, more than every other app's checks put together. The same check
is now a 304 of 238 wire bytes, confirmed in production.

**The Network Activity pane leaves the workbench**, which was the wrong
home for it: every other row there is one app, and nothing in the log is.
It also left the sidebar with a selection no list row could match.

## Storage

Events live in SQLite with a `kind` discriminator and a JSON payload, so
an older `duo` meeting a newer app's rows degrades to "unknown kind, kept
verbatim" rather than taking the log down. `traffic.json` is migrated in
on first launch, keyed by content hash so a re-import replaces its own
rows instead of doubling a lifetime total — verified across a downgrade
and back on a real machine with 187 real installs.

Two schema migrations backfill from payload fields that were always
there, so history reads the same way new traffic does.

## Verification

`make test` — 2099 tests. Two pre-existing failures are unrelated and
reproduce on the base commit: a Postman endpoint that times out through
the local proxy, and CapCut's beta recipe resolving no installer URL.

Reviewed at high effort; seven defects found and fixed, two of which were
introduced by the fixes themselves and caught on the second pass.


jizhi0v0:
## Review pass: six defects found and fixed

Reviewed the branch and found six real issues; three were confirmed with measurements rather than reasoning. All six are fixed in `a63f0e0`.

| # | Issue | Evidence |
|---|-------|----------|
| 1 | Events lost at process exit — the delegate filed them via `Task { await … }`, but `duo` reads `hasRecorded` and exits the moment its command returns | structural race; now a synchronous hand-off |
| 2 | A busy database lost the whole batch — buffer cleared before `BEGIN IMMEDIATE` was known to succeed, then inserts ran in autocommit, breaking event/rollup atomicity | reproduced with a second connection holding the write lock |
| 3 | `make test` wrote into the developer's real store | **1643 `cli` events**, incl. 252 to `127.0.0.1` and 48 to `example.com` — a host nothing contacted — in the never-pruned totals |
| 4 | `requests recent --json` truncated all 11 phase timestamps to whole seconds | `fetchStart`/`domainLookupStart`/`connectStart` all printed `2026-09-03T13:12:21Z`; stored payload has a 1 ms gap |
| 5 | Retention ceiling and `--status` size ignored the `-wal` file | 2 686 976 B database vs 951 752 B WAL — 36% low |
| 6 | `reset` cleared the two tables non-atomically | not test-covered, and says so in the comment |

Four of the six have a guard that fails when the fix is reverted (1–4 above map to the first four; 5 has its own). The fifth mutation initially passed because it was too weak — it left the COMMIT-failure requeue intact — and was redone against the true pre-fix shape. The reset fix has no guard because reaching the bad state needs a crash between two deletes, and a test that cannot fail is worse than none.

After the fix, one `make test` adds **zero** loopback or fixture rows to the real store (833 events go to a scratch store instead). The rows that do appear are the running menu-bar app's own checks against real vendor hosts.

`make test` green: 2049 + 188.

**Local cleanup for anyone who ran the suite before this commit:** the pre-fix rows are still in your store, and they are in `totals` permanently.

```
duo requests reset --yes --all-clients
```

jizhi0v0:
## A seventh, found by actually running the command

`duo requests reset` cleared every row — and left the file at **14 897 152 bytes** with a `page_count` of 126, i.e. 516 KB of real content.

`incremental_vacuum` had done its job. In WAL mode the truncation it performs is written to the log and only reaches the database file when something checkpoints, and I had the checkpoint running *before* the vacuum instead of after. One checkpoint afterwards brought the file to exactly **516 096 bytes**.

The same inversion was in the size-retention loop, where it did more than waste space: the loop steers by the file size, so it was reading a figure one round stale and giving up rows it did not need to.

Fixed in `4e8bcb0`, with a guard that grows the store past a megabyte, resets it, and fails if the file does not shrink. Swapping the two statements back makes it fail.

Worth noting how this one surfaced: it was invisible to every test I had written, because they all asserted on row counts and on `databaseBytes()` *after* a prune had happened to checkpoint. It took someone running the command on a real store and me looking at `ls`.

`make test` green: 2050 + 188.

jizhi0v0:
## The download ledger moved in too

`traffic.json` and the event store were two accounts of the same download at different altitudes — one counting the file that landed, the other counting wire bytes per redirect hop. Two stores that have to agree eventually don't. There is one now, with `install` as a second event kind.

### Verified against the real file, not a fixture

| | before | after |
|---|---|---|
| apps | 81 | 81 |
| events | 586 | 586 |
| bytes | 115.367 GB | 115.367 GB |
| month buckets | 4 | 4 |
| source buckets | 3 | 3 |
| per-app histories identical | — | 81/81 |

A forced repeat still yields 586. `traffic.json` is untouched (138 767 bytes, same as before) and stays as the user's own copy.

### What makes it safe

- **Stable ids.** Every migrated event's id is derived from its own content, and the per-app totals are `SUM`ed from those rows rather than incremented — so a repeat replaces its own rows instead of doubling a lifetime total. The `meta` marker stops it running twice at all, but a flag is a promise and the id is a proof.
- **Installs are exempt from both retention passes.** Retention bounds a diagnostic log; this is the answer to what keeping a machine up to date has cost. 586 rows across the app's whole life, against thousands of request events a day.
- **`reset()` spares them** unless a caller explicitly says otherwise — `duo requests reset` would have taken someone's entire download history with the network log.
- **They stay out of the network rollups.** `Downloader` already files request events with purpose `.install` for the same transfer; counting the ledger entry too would report one download twice, at two different byte counts.

### Two defects a fresh-database test structurally could not find

Both surfaced only when the migration ran against this machine's actual store:

1. `CREATE TABLE IF NOT EXISTS` does nothing to a table that already exists, so `app_id` never arrived on a database created by the earlier commits, and every insert failed on the unknown column.
2. **The import marked itself done having written nothing.** The marker is what makes the migration one-shot, so a marker after a failed write does not lose a launch — it loses the history on every launch after it, permanently.

Columns are now added explicitly, and an import counts what actually landed before recording that it happened. Both have a guard; reverting either fix makes it fail.

`TrafficStore` keeps only its reader. Its eleven rules moved to the new store rather than being deleted with it — a rule that stops being executed because its subject was renamed is a rule that stops holding.

`make test` green: 2067 + 188.

PR #323, opened 2026-09-04
Read PDF Expert's live Sparkle feed instead of the frozen one its bundle names

jizhi0v0:
PDF Expert ships Sparkle, and the generic source has always resolved it — at the address the bundle names, which Readdle froze on **2022-07-12**.

```
$ curl -sI https://downloads.pdfexpert.com/release/appcast.xml
last-modified: Tue, 12 Jul 2022 14:49:50 GMT
```

Four items, newest build `764` / `2.5.22`. Three cap out at `maximumSystemVersion` 10.10–10.12 — but **the fourth does not**. So on a current Mac this is not a feed that fails; it is a feed that succeeds with a 2022 answer: `usableItems` keeps 764, `2.5.22` loses to the installed `3.13.2`, and the row reads "up to date". Forever, at HTTP 200, with no miss for `RecipeHealth` to show. The `pdf-expert` cask is `auto_updates: true`, so `HomebrewCaskSource` skips it and nothing else picks the app up either.

The live feed is `/pem3/release/appcast.xml`.

## The design decision

`SparkleFeedCatalog` could only ever fill a blank, and its doc comment argues at length that an address a bundle *states* is the app speaking for itself. That argument still holds, so the new table matches on **the dead address, not the bundle id**:

```swift
"com.readdle.pdfexpert-mac": SupersededFeed(
    declared: URL(string: "https://downloads.pdfexpert.com/release/appcast.xml")!,
    live:     URL(string: "https://downloads.pdfexpert.com/pem3/release/appcast.xml")!),
```

`replacement(forBundleID:declaredFeed:)` returns the live feed only on a byte-identical match. The day Readdle edits their plist — to pem3, or to a fourth address — the entry stops applying and the bundle's own word wins again. A bundle-id lookup would keep asserting our 2026 reading of their infrastructure over theirs, and would look identical while doing it.

Applied in `AppScanner.readApp` after the fill-in and **before** `ChannelBinding`, whose `feedOverride` is a channel decision and still outranks it. It does not touch `channelIsAuthoritative` — the pem3 feed carries no `<sparkle:channel>` tags, so inference is unchanged.

## Why pem3 and not merely the newer-looking address

Stated with its strength labelled, because the first version of this was circular:

1. **Apple's attestation, independent of the CDN path** — the payload's `.app` is Developer ID-signed by Team `3L68KQB4HG` and notarized (`spctl` accepts it). Whoever operates that path cannot forge it.
2. **A third party points at the same build** — Homebrew's `pdf-expert` cask, maintained by nobody here, links `pem3/versions/1172/PDFExpert.zip` at `3.13.2,1172`.
3. ⚠️ **Internally consistent, not independent** — the payload's `SUPublicEDKey` does verify that item's `edSignature` over the exact 128 MB zip (Ed25519). But key and payload both come from the pem3 tree.

(3) **cannot** be made independent: builds 764, 936 and 964 — everything the *dead* feed serves — carry no `SUPublicEDKey` at all, while all four builds declare the same dead address. Which has a real consequence: every user this change rescues is below 3.13 and so has no key, meaning one-click runs under the Developer ID / Team gate and **EdDSA never applies to them**. Written into the audit.

## Red → green, on a real bundle

Extracted from the vendor zip, driven through the production chain (`channel-verify` runs `UpdateChecker.check`, not an approximation):

| | before | after |
|---|---|---|
| `SUFeedURL` | `/release/appcast.xml` | `/pem3/release/appcast.xml` |
| latest | `2.5.22` | `3.13.2` |
| deltas | 0 | 2 |
| a `3.13.1` copy | `up to date` | `UPDATE → 3.13.2` |

## Release notes

The appcast's `releaseNotesLink` is a 3.5 KB page holding only the newest release. The history lives behind `sparkle:fullReleaseNotesLink` — an element `SparkleAppcastParser` does not read at all, hence a recipe.

The heading is matched as `<p[^>]*>` rather than the bare `<p>` the page uses today, because this recipe cannot fail the usual way: a lost terminator yields **one** correctly-versioned entry carrying the whole page, and `duo verify` records only the newest version, so that failure sweeps green. One added class attribute would have been enough. Pinned by a test.

88 headings but **85 distinct versions** — three numbers appear twice, two of those pairs with entirely different text, and the extractor's version+title dedupe drops the second of each. Left alone: which of two same-numbered paragraphs is real is the vendor's question.

## Both harnesses were lying

Neither goes through `AppScanner.readApp`. `channel-verify` reproduced the bug it exists to catch, and `feed-discover` printed *"SparkleAppcastSource already resolves this app"* beside an address production refuses to use. Both taught the new step; `FeedDiscovery` gained a `.superseded` verdict, since `.declared` used to mean two things that moved together and no longer do.

## Verification

- `make test` — green, 0 failures
- `duo verify --only pdfexpert` — `changelog ✓ 1`, `version 3.13.2`, against the live page
- Mutations checked: the scan test fails with the `AppScanner` call removed, the `FeedDiscovery` test with its branch removed, the restyled-paragraph test with the terminator reverted

## Not done, deliberately

- **No end-to-end install.** That needs an installed copy; no evidence was taken from one. `needs-verify` in the audit.
- **`SparkleFeedCatalog` is outside `duo verify`** (it sweeps vendor / GitHub / changelog only). Worth flagging that this PR changes the table's risk shape: it could previously only fail to fill a blank, and can now assert an address a bundle contradicts — so the half that can be *wrong* is the half nothing watches. A fourth registry in `verify` is the fix, out of scope here.
- **`phasedRolloutInterval`** (86400 on the current item) is read nowhere in the repo, so we offer a build as soon as the feed carries it. Global behaviour, not this app's.

Adversarial review run per `CLAUDE.md`; nine findings, all addressed or documented.


jizhi0v0:
**Update: one-click is verified end to end now** — a real install turned up, so the `needs-verify` item in the description is closed.

An installed **3.13.1** copy (Readdle's own `pem3/versions/1171` build, notarized, Team `3L68KQB4HG`) was updated back to 3.13.2 through the engine:

```
$ duo check --all --json | grep pdfexpert
  installed 3.13.1/1171   latest 3.13.2/1172   route "in-place"   source "Sparkle"   hasUpdate true

$ duo install com.readdle.PDFExpert-Mac --yes --json
  {"app":"PDF Expert","applied":true,"bytesDownloaded":781078,"route":"sparkle"}
```

After: `3.13.2 / 1172` on disk, `codesign --verify --deep --strict` clean, `spctl` still `Notarized Developer ID / Readdle Technologies Limited`, the app relaunches, and `duo backups` holds a 3.13.1 rollback point.

**`bytesDownloaded: 781078` is the delta**, not the 128 MB full zip — exactly the size of `PDF Expert1172-1171.delta` in the feed. So the delta path is exercised too, through `SparkleAppcastParser`'s own parsing (not `VendorAppcastDeltas`, which this app cannot reach).

Before the downgrade, the untouched 3.13.2 install also confirmed the swap on the production CLI path: `source "Sparkle"`, `latestVersion 3.13.2` — a number the feed its bundle names cannot produce, since that one tops out at 2.5.22.

**And a claim narrowed.** The audit said every copy below 3.13 lacks `SUPublicEDKey` and so takes the unsigned-feed branch. What was actually measured is 764/936/964 without a key and 1171/1172 with one — the whole 3.4.1→3.13 range was never looked at, and I had extrapolated from three samples. Corrected to "old copies without a key exist; which versions is unmeasured". Consequence: this run exercised the **EdDSA** branch, so the unsigned-feed branch remains unverified.

jizhi0v0:
**The unsigned-feed branch is verified now too** — the last gap in the description is closed.

The concern was that the copy used for the first end-to-end run (3.13.1) carries `SUPublicEDKey`, so it exercised the EdDSA path — while the users this change actually rescues are stuck on older builds that have no key and take the other branch.

So I fetched **3.4.1 / 964 from the dead feed's own enclosure** (`/release/10_15/PDFExpert.zip`) — exactly what a stuck user has. Same bundle id, Team `3L68KQB4HG`, notarized, universal, and **neither `SUPublicEDKey` nor Sparkle 1.x's `SUPublicDSAKey`**: no signing key of any kind, not an older signing scheme. (DuoUpdater reads no DSA key anywhere either.)

```
$ duo check   →  3.4.1/964 → 3.13.2/1172   route in-place   source Sparkle
$ duo install →  {"applied":true,"bytesDownloaded":128745812,"route":"sparkle"}
```

Full 128 MB archive rather than a delta, as expected — the feed publishes deltas only from 1170/1171. Afterwards: `3.13.2 / 1172`, `codesign --verify --deep --strict` clean, `spctl` still `Notarized Developer ID / Readdle Technologies Limited`.

| starting point | `SUPublicEDKey` | gate taken | downloaded | result |
|---|---|---|---|---|
| 3.13.1 / 1171 | yes | EdDSA + code signature / Team / bundle id | 781,078 B (delta) | ✓ 3.13.2 |
| 3.4.1 / 964 | **none at all** | EdDSA skipped; code signature + Team + bundle id | 128,745,812 B (full) | ✓ 3.13.2 |

Skipping EdDSA is not an observed coincidence: `verifiesWithEdDSA` is computed straight from `sparkleEdPublicKey`, so the bundle decides it.

**Still not covered**, stated plainly: the branch's *fail-closed* direction on this app — a payload signed by a different team should be refused. That behaviour lives in `InstallPipelineSmokeTests` / `VendorInstallTests`, not here.

Incidental: the 3.4.1 copy was placed at a second scan location rather than replacing the first, and once both copies were at 3.13.2 `dedupeIdenticalInstalls` folded them into one row — matching its doc comment.

PR #325, opened 2026-09-04
Let a closed vendor track be closed, not a failed check

jizhi0v0:
CapCut's beta track closed between releases and the row went red for it.
It should not have.

## What happened

9.4.0 graduated to stable and no new beta has opened, so `lastest_url` —
the field the beta recipe reads — now carries the **stable** artifact. The
recipe pins the beta token inside the filename, so it correctly refuses to
read a stable build as a beta one, and the probe finds no version.

That arrived as a check failure: a red row, a line in the "could not be
checked" banner, and a Retry that cannot work until ByteDance opens the
next beta.

**The recipe was not broken.** Days earlier the same recipe resolved
`9.4.0-beta7 → 9.4.0-beta8` with the Update button live — the "before"
picture rules out rot and leaves only "the vendor closed the track".

## The fix

A recipe can declare `trackClosedPattern`: the vendor saying, in their own
data, that this track has nothing published. CapCut empties
`lastest_beta_number` between betas — `"4"` while beta4 was current, `""`
once 9.4.0 shipped. A match reports `notApplicable`, which returns nil
instead of throwing, so the source simply has no answer.

Three constraints the mechanism is wrong without:

- **consulted only after the version pattern already missed** — a track
  that is publishing can never be talked into looking closed;
- **anchored to its key** — a bare `""` matches in a thousand places in a
  436 KB body and would swallow a genuinely broken recipe;
- **beta only** — stable's field is always populated, and giving it a
  closed state would let a real outage pass as "nothing published".

The vendor sweep's exemption is derived from the registry rather than
hand-listed, and asserts the track is *still* closed, so the day the
vendor reopens it the build fails until the declaration is removed.

## Also

- Corrects a measurement recorded on 2026-08-27 and stale since:
  `capcutpc_beta` does now return an `update_reminder`, field-for-field
  identical to `capcutpc_0`. The comment says so rather than being quietly
  edited, because the next reader should re-measure rather than trust
  either version of it.
- The updater's own release check, changelog fetch and Sparkle download
  showed a blank app column, filed beside the Homebrew catalog — which is
  blank for the opposite reason. The catalog belongs to no app; a
  self-update belongs to exactly one.
- The audit doc carries the re-measurement, the vendor-side defects found
  along the way (their What's new panel and their `1970-01-01` release
  date, neither of which touches us), and why falling back to the stable
  track — which CapCut's own updater does — was **not** implemented: it
  needs `ChannelProofRegistry` to accept a `capcutpc_0` artifact for a
  beta row, and that gate exists to stop exactly that.

## Verification

`make test` — 2222 + 188, green. The new mechanism is covered at the
recipe level and by the live vendor sweep; the `VendorProbeSource` branch
itself has no unit test, which would need a probe harness that can serve a
canned body.

### DIFF
--- a/App/Sources/AppUpdater.swift
+++ b/App/Sources/AppUpdater.swift
@@ -117,6 +117,11 @@ private final class SelfUpdateInstaller: NSObject, SPUUpdaterDelegate {
         let event = RequestEvent(
             purpose: .selfUpdate, method: "GET", scheme: item.fileURL?.scheme,
             host: host, port: item.fileURL?.port, path: item.fileURL?.path ?? "",
+            // Ours. The self-update is the one fetch that is unambiguously for a
+            // single app, and leaving it blank put our own downloads in the same
+            // column as the Homebrew catalog — which is blank because it is
+            // shared by every app, the opposite reason.
+            appID: Bundle.main.bundlePath,
             // A synthetic task with a single hop: Sparkle's own redirects are as
             // invisible to us as its byte counts, and inventing hops we did not
             // observe would be worse than recording the one we can attest to.
--- a/App/Sources/SelfChangelogView.swift
+++ b/App/Sources/SelfChangelogView.swift
@@ -142,8 +142,10 @@ struct SelfChangelogView: View {
         request.cachePolicy = force ? .reloadIgnoringLocalCacheData : URLRequest.versionFeedCachePolicy
         request.timeoutInterval = 15
         do {
-            let (data, response) = try await URLSession.updates.countedData(
-                for: request, purpose: .selfUpdate)
+            let (data, response) = try await RequestAttribution.withApp(Bundle.main.bundlePath) {
+                try await URLSession.updates.countedData(
+                    for: request, purpose: .selfUpdate)
+            }
             if let http = response as? HTTPURLResponse, !(200..<300).contains(http.statusCode) {
                 state = .failed(String(localized: "GitHub answered HTTP \(http.statusCode)."))
                 return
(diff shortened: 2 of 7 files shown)


jizhi0v0:
Reviewed at high effort. One finding, and it was mine — introduced by the fix itself:

**`VendorInstallTests.swift:124` — the sweep would have failed the day the vendor reopens the track.** The exemption keyed on "declares `trackClosedPattern`" as if that meant "is dormant right now". It does not: the pattern is a permanent property of the recipe, saying how this vendor *signals* dormancy. So when ByteDance opens beta9 the recipe resolves again — correct behaviour — and the sweep would have failed telling the reader to delete the declaration, which is precisely what keeps the row calm the next time the track closes. While the track was open it also skipped the installer-kind, install-routing and cross-channel-artifact assertions, so a working channel silently lost coverage.

Fixed in 0113017: the exemption now applies only when the probe actually resolved nothing.

Verified in passing:
- the new branch sits inside `guard let version = extractor(…) else`, so "consulted only after the pattern already missed" is structural rather than a claim in a comment;
- `duo verify` maps `.notApplicable` to `.skipped`, not `.broken` — the nightly sweep will not report this;
- `RecipeHealth` still records a miss, which the existing code documents as wanted for `.notApplicable`, and the detail reads "no current build on the beta track".

Mutation-tested: removing the declaration puts `resolved no installer URL` back.

Known and deliberate, recorded in the audit rather than fixed here: the row goes from red-with-a-reason to a silent "—". The vendor states the track is closed and we drop that on the floor; keeping it needs a new row state, which is its own change.

`make test` — 2222 + 188 green.

PR #386, opened 2026-09-06
verify: update recipe baseline

jizhi0v0:
Nightly recipe sweep. Carries the next run's failure streaks and issue numbers.


jizhi0v0:
Closing: this branch was cut from a feature branch instead of main, so it carries three files that belong to #383 (now merged separately, and squashed — so the branch has diverged and re-applying it would revert a review fix). Auto-merge was disabled for that reason. The baseline update itself is fine and will be regenerated by the next scheduled run against a clean main.

PR #392, opened 2026-09-06
Sweep SparkleFeedCatalog's addresses on the same schedule as the recipes

### LINKED ISSUE #324: duo verify does not sweep SparkleFeedCatalog, and that table can now be wrong
`duo verify` sweeps three registries — `Registry` in `CLI/Sources/DuoKit/Finding.swift:7` is `vendor`, `github`, `changelog`. `SparkleFeedCatalog` is in none of them, so no scheduled check ever fetches the addresses it hands out.

That was fine while the table could only *fill a blank*. #323 changed its risk shape: `supersededFeeds` lets it **assert an address against one the bundle explicitly states**. The half that can now be wrong is the half nothing watches.

## What goes unnoticed today

For `com.readdle.pdfexpert-mac` → `downloads.pdfexpert.com/pem3/release/appcast.xml`:

- **the live feed dies or moves** — every PDF Expert install silently stops seeing updates. Same failure the entry was added to fix, one address further along.
- **the vendor edits `SUFeedURL`** — the entry correctly stops applying (it matches on the dead address, by design), but nothing tells us the entry is now dead weight. It just sits there matching nothing.
- **the item shape changes** so `usableItems` filters everything — `SparkleAppcastSource` logs an `.info` line and returns nil, and `.info` for a third-party subsystem is not persisted.

None of these produce a miss anywhere a person looks. The original bug took an unknown number of years to notice for exactly this reason: the dead feed answered 200 and parsed fine.

## Sketch

A fourth `Registry` case, whose unit is one catalog entry:

- fetch the feed, parse it, assert ≥1 item survives `usableItems` on a current OS, and record the newest version the way the other sweeps record `version` — that is what `Baseline.reconcile` needs to catch a regression.
- for a `supersededFeeds` entry, also assert the recorded `declared` address is **still the one the vendor ships**, and report it when it is not. That is the entry's own expiry condition and it is cheap to check: fetch nothing, just compare against the bundle if one is installed, or leave it to the "no longer matches anything" report.
- `recipeID` shape: the existing ones are `changelog:<bundleID>:<channel>`; `feed:<bundleID>` would fit. Note `Baseline.prune(keeping:)` is keyed on the registries, so the new ids have to be added to the `live` set in `Verify.run` or every sweep will drop them again.

Two entries today (`net.imput.helium`, `com.readdle.pdfexpert-mac`), so this is ~2 requests, not a meaningful cost to the nightly run.

## Related, separable

`sweepChangelog` records only `newest.version`, never an entry count. The PDF Expert changelog recipe's realistic failure mode is not zero entries — a lost terminator yields **one** correctly-versioned entry carrying the whole page, which sweeps green. An entry count in `Finding` (and a baseline comparison on it) would catch that class. Worth its own issue if this one gets split.

jizhi0v0:
Closes #324.

`duo verify` swept four registries; `SparkleFeedCatalog` was in none of them, so no scheduled check had ever fetched the addresses it hands out. #323 changed that table's risk shape — `supersededFeeds` lets it assert an address *against* one the bundle explicitly states — and the half that can now be wrong was the half nothing watched.

Every way it fails is quiet. `SparkleAppcastSource` answers a dead, moved, or reshaped feed with nil, and every caller renders nil as "up to date". That is the same silence the PDF Expert entry was added to escape, one address further along.

## What lands

A fifth registry, `--feed`, one finding per catalog entry, keyed `feed:<bundleID>`. Two entries today: two requests for the superseding one (it reads both addresses), one for the fill-in.

**Core** — `SparkleAppcastSource.readFeed` is the production fetch with no installed copy behind it. `usableItems` needs an app to decide which channels are allowed, and a runner with nothing in `/Applications` has none, so it asks the question the catalog's promise is actually about: what would a *fresh default-channel install* see today? The request moves into `feedRequest`, shared with `latestVersion` — a hand-rolled copy that lost `versionFeedCachePolicy` would be a sweep reporting on a request the app never makes.

**CLI** — `FeedVerify.swift`, split network half / pure `classifyFeed`, the same split `AppStoreVerify` uses. In order:

| check | verdict | why it matters |
|---|---|---|
| the address answered | `infra` | a bad minute is not a dead feed; `Baseline.isInfraReportable`'s week-long window decides |
| the body carried items | `broken` | a CDN error page parses to zero items and 200s |
| ≥1 item survives the filters | `broken` | **the silent one** — this is exactly what `latestVersion` turns into "up to date" |
| the head item names a marketing version | `warn` | it is what the baseline records, and the only field this source may compare against a bundle's own |
| its enclosure resolved | `warn` | detection survives, installing does not — same call as `installURLUnresolved` |
| *(superseded only)* the dead address is still behind | `warn` | the entry's own expiry condition |

That last one is the part the issue left open. It sketched comparing against an installed bundle, which neither entry has here and CI never will. Fetching the address we redirect *away from* answers the same question for one request and answers it everywhere.

Never downloads an enclosure — shape only. Whether the artifact is still served is a separate question with a separate cost; `installURLNotFound` is the precedent if it is ever worth the request.

## Measured

Live, today:

```
Sparkle feed  ✓ 2  ⚠ 0  ✗ 0  ~ 0  - 0
feed:com.readdle.pdfexpert-mac  3.13.2  (5 items, 1 usable — the other 4 cap at macOS 11.999)
feed:net.imput.helium           0.16.5.1 (10 items, 9 usable — 1 is beta-tagged)
```

PDF Expert's live feed has **one** usable item on a current Mac. Its margin over "everything filtered, reads as up to date" is that single entry, which is a good argument for the check existing.

Negative controls against the same live endpoints — swap the superseded entry's two addresses, point the fill-in at an HTML page:

```
⚠ supersededAddressIsNoLongerBehind — … (…/pem3/release/appcast.xml at 3.13.2 (1172), live 2.5.22 (764))
✗ feedParsedToZeroItems — the address answered, with a body carrying no appcast items at all (38824 bytes).
```

`make test` green (`MAKE_TEST_RC=0`). 12 new CLI cases, 11 mutations applied and run: all compile, all go red, each hits the case it was written for. `everyCatalogEntryIsSwept` pins the derivation — dropping the superseded half from `verificationCases` fails four expectations.

## Left out, on purpose

The issue's "related, separable" half — `sweepChangelog` recording no entry count, so a lost terminator yields one correctly-versioned entry carrying the whole page and sweeps green. Different registry, different field, worth its own issue.

### DIFF
--- a/CLI/Sources/DuoKit/FeedVerify.swift
+++ b/CLI/Sources/DuoKit/FeedVerify.swift
@@ -36,9 +36,10 @@ extension Verify {
         let source = SparkleAppcastSource()
 
         // Sequential with the same inter-request delay every other sweep uses.
-        // Today's entries sit on a host each, so the delay buys nothing — but a
-        // table this small has no parallelism worth having either, and the two
-        // halves of a superseding entry are on one host by construction.
+        // Today's entries sit on a host each, so the delay buys nothing between
+        // them — but a table this small has no parallelism worth having either,
+        // and a superseding entry's two reads go to the same host back to back,
+        // which is the pair the delay is actually for.
         var findings: [Finding] = []
         for (index, entry) in cases.enumerated() {
             if index > 0 {
@@ -47,15 +48,16 @@ extension Verify {
             }
             // Requests, not probes: a retried entry has cost the endpoint every
             // one of them, and `Finding.attempts` is documented as the number
-            // that cannot understate what the sweep spent.
-            let perProbe = entry.declared == nil ? 1 : 2
+            // that cannot understate what the sweep spent. Carried forward from
+            // each probe rather than multiplied out, because a probe that gave
+            // up after an unreachable live read spent one request, not two.
             var attempt = 0
             var finding = await probe(entry, source: source, requestsAlready: 0)
             while attempt < options.infraRetries, finding.status == .infra {
                 attempt += 1
                 try? await Task.sleep(for: .seconds(attempt))
                 finding = await probe(
-                    entry, source: source, requestsAlready: attempt * perProbe)
+                    entry, source: source, requestsAlready: finding.attempts)
             }
             findings.append(finding)
         }
@@ -75,9 +77,16 @@ extension Verify {
         // what makes the entry's own expiry condition mechanical rather than a
         // note in a comment: it is only justified for as long as the address it
         // overrides is genuinely behind.
+        //
+        // Skipped when the live read already failed: `classifyFeed` returns from
+        // its `.unreachable` branch without ever looking at this one, and the
+        // retry loop above would otherwise spend three extra requests on a host
+        // that is down — the one moment it should be asking for less, not more.
         var declared: FeedObservation.Fetch?
-        if let deadAddress = entry.declared {
+        var requestsSpent = 1
+        if let deadAddress = entry.declared, case .read = live {
             declared = await read(deadAddress, bundleID: entry.bundleID, source: source)
+            requestsSpent += 1
         }
 
         let verdict = classifyFeed(entry, FeedObservation(live: live, declared: declared))
@@ -85,10 +94,11 @@ extension Verify {
             recipeID: entry.recipeID, registry: .feed, bundleID: entry.bundleID,
             channel: entry.kind.rawValue, status: verdict.status, version: verdict.version,
             // The LIVE address's host: it is the one an app is pointed at, and
-            // the one a nightly triage should route by. A superseding entry's
-            // dead address is named in the warning that complains about it.
+            // the one a nightly triage should route by. The dead address is
+            // named in full inside the warning that complains about it, which
+            // is what a reader needs when the two share a host.
             warnings: verdict.warnings, endpointHost: entry.feed.host ?? "-",
-            attempts: requestsAlready + (entry.declared == nil ? 1 : 2),
+            attempts: requestsAlready + requestsSpent,
             elapsedMs: Int(Date().timeIntervalSince(started) * 1000))
     }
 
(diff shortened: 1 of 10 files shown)


jizhi0v0:
`/code-review 392 high` — 4 findings, all mine, all fixed in 3309290.

1. **`FeedVerify.swift` — a dead address offering nothing was a complaint.** `VersionComparator.isNewer` fails closed, so an empty `deadSide` answered "not behind". A superseding entry whose dead address had stopped offering anything usable — a retired path answering with a landing page, or the vendor capping its last uncapped build the way that feed already caps three of four — would have warned every night and opened an issue against the one entry behaving perfectly. That state is the entry's justification at its *strongest*. Two non-empty sides that cannot be ranked still complain, which is what the fail-closed reading was actually for.

2. **`SparkleFeedProbe.swift` — `usableCount` moved with the host.** `usableItems` defaults `hostArch`/`allowingIntelTranslation` to the running Mac, and Rosetta is absent here and present on a GitHub runner, so the same feed could read `broken` on one machine and `ok` on another while `Baseline` keyed the streak on the recipe id alone. Pinned to arm64 (a product decision, `App/project.yml`) with translation allowed: the question is whether the *address* can serve a supported install, not what one Mac can run. `osVersion` is now the only host fact left in the answer, and it is reported.

3. **`FeedVerifyClassificationTests.swift` — the enclosure case fixtured a state the sweep cannot produce.** The parser resolves every enclosure against the feed's own address, so Helium's relative `assets/….dmg` arrives absolute. A base-less relative URL was a fixture wider than production, and it left the branch that *can* fire — a non-http scheme — untested. Rewritten around `ftp://`, with the resolution it used to claim to cover pointed at where it is actually pinned.

4. **`FeedVerify.swift` — the dead address was fetched even when the live read had failed.** `classifyFeed` returns from `.unreachable` without looking at it, so an outage on a superseding entry's host cost six requests where three carry the same information. Measured after the fix: an unresolvable live address reports `attempts=3`.

Two unenforced invariants I had written as "by construction" (both halves of an entry on one host) were softened to what is actually true.

Mutations re-run on the changed logic — dropping the empty-side guard, reading the fail-closed `false` as fine, dropping the scheme check — each goes red on the case written for it. `make test` green (`MAKE_TEST_RC=0`), live sweep still `✓ 2`.

PR #395, opened 2026-09-06
Count changelog entries, and notice when a page collapses into one

### LINKED ISSUE #393: sweepChangelog records no entry count, so a lost terminator sweeps green
Split off from #324, which raised it as separable.

`Verify.sweepChangelog` records `newest.version` and nothing else. That catches a recipe whose entry pattern stopped matching altogether — zero entries, `noEntriesExtracted`, loud. It does not catch the failure mode these recipes actually have.

## The shape

A changelog entry pattern is `<start> … <lookahead for the next start, or the end of the document>`. When the *terminator* half stops matching — the vendor adds a footer, moves the compare-link block, changes the heading markup one level down — the pattern still matches, the version still parses correctly, and the winner swallows everything after it. One entry, right version, whole page as its body.

#378 shipped exactly this: the Keep a Changelog entry pattern had no boundary before the trailing link block, so the oldest entry's last item absorbed 17 compare URLs — 1709 characters. The sweep was green throughout. It was found by looking at the rendered notes, not by the job built to find this class of thing.

## What would catch it

`Finding` carries an entry count, and `Baseline` compares it the way it compares `version`. The realistic signature is a **collapse** — N entries yesterday, 1 today — which is as mechanical as a version going backwards and about as cheap.

Worth thinking about before writing it:

- **A drop is not always a break.** `maxEntries` caps the list, and a vendor pruning their own page legitimately shrinks it. A comparison that complains about any decrease will complain about ordinary vendor housekeeping; one that only complains about a collapse to 1 (or to some small fraction of the baseline) is narrower and probably right.
- **The count alone still misses the two-entry version** of the same bug, where the terminator matches once and fails once. A body-length check per entry would catch more, but "how long is too long" is a threshold nobody can defend, whereas "this went from 20 to 1" needs no threshold at all. Start with the count.
- **Two-stage recipes** (`indexLinkPattern`) count entries on the second page, not the index. `classifyChangelogFailure` already distinguishes which request failed; the count has to be attributed the same way or the report will point at the wrong page.

Not urgent — the class is real but slow-moving, and the version check already covers the loud half.

jizhi0v0:
Closes #393.

`sweepChangelog` recorded `newest.version` and nothing else. That catches a recipe whose entry pattern stopped matching altogether — zero entries, loud. It cannot see the failure where the pattern still matches and the *page* has collapsed.

An entry pattern is a start plus a lookahead for the next start. When a vendor's restyle breaks only the **lookahead**, the first entry's body runs to the end of the document and every later start is swallowed inside it. One entry, correct version off the first heading, green sweep.

## Correcting the issue I filed

#393 named #378 as the motivating example. **That was wrong, and I should have checked it before filing.** `bf5db16` measured the fix as "the same 17 entries parse with the same item counts" — the bug there was an *item* pattern running past the end of the last entry, not an entry pattern losing its terminator. The entry count never moved.

Worse for the issue's framing: that bug was wrong from the recipe's first sweep, so **no history-based check could ever have caught it**. There was never a healthier count to fall from. What catches that class is a review, which is in fact what caught it.

So this lands the check the *original* observation in #324 asked for — an entry-level regression — and the test suite says so at the top rather than repeating the wrong example.

## The rule, and why it is this narrow

Complain when a recipe that previously read **more than one** entry now reads **exactly one**. Measured across the 80 changelog recipes that answered on 2026-09-07:

| entries | recipes |
|---|---|
| 1 | 6 |
| 2–19 | 27 |
| 20 (a `maxEntries` cap) | 22 |
| 21–39 | 13 |
| 40 (a `maxEntries` cap) | 12 |

- **Six recipes sit at one entry as their normal state** — VS Code, Ghostty, Blender, VLC, QQ Music, Doubao IME each publish one release per page. One entry is not suspicious; *becoming* one is. Hence `previous > 1`.
- **A fractional rule has nothing to calibrate against.** 22 recipes sit exactly at their cap of 20 and 12 more at 40, where the count is insensitive to a vendor adding releases and moves only when they prune — which they are entitled to do. A partial merge is therefore **not** covered, deliberately, rather than covered by a threshold nobody could defend. Said out loud in the code and in a test.

## `reconcile` now returns every complaint

The two history checks fire together for one cause: when entries merge, the surviving version is whichever heading the merged entry kept — the *oldest* on the page for a `newestLast` recipe. Reporting the version regression and dropping the collapse that explains it sends whoever picks it up after a version pattern that never changed.

The fold moved out of `run` into `foldingBaselineComplaints` so it can be tested. Without that it is one `.first` away from silently throwing half the complaints out, and no test would have noticed — which was true of the first version of this branch.

## The bug this nearly shipped with

⚠️ `Baseline.Entry` **decodes by hand and encodes by synthesis.** The first version declared `lastGoodEntryCount` and left it out of `init(from:)`.

Nothing failed. The count was written to `baseline.json` every sweep — it was right there in the file, looking correct to anyone who opened it — and read back as nil every sweep, so the check never once had a previous value to compare against. All seven unit tests passed throughout, because none of them touch a file.

A two-sweep run against a real page is what showed it. There is now a case that goes through the disk, and a warning on that initializer.

## Measured

- Two consecutive full changelog sweeps against a fresh baseline: `✓ 80  ⚠ 0  ✗ 0` both times. No recipe legitimately collapses between runs.
- Planting a previous count of 20 on a page that reads 1:

```
⚠ WARN  changelog:com.mitchellh.ghostty:-
    warning   entry count COLLAPSED to one since the last sweep (20 → 1) — an entry
              pattern whose terminator stopped matching leaves the first entry
              carrying the whole page, with its version still parsing correctly off
              the heading
```

- `make test` green (`MAKE_TEST_RC=0`).
- 8 new cases, 7 mutations applied and run: all compile, all go red, each on the case written for it.

The count also reaches the places a human reads: the Markdown report's per-finding detail, and the issue body's table as `entries parsed | 1 (was 20)`.

### DIFF
--- a/CLI/Sources/DuoKit/Baseline.swift
+++ b/CLI/Sources/DuoKit/Baseline.swift
@@ -25,6 +25,12 @@ public struct Baseline: Codable, Sendable {
     public struct Entry: Codable, Sendable {
         public var lastGoodVersion: String?
         public var lastGoodAt: Date?
+        /// Entries the last sweep that got an answer extracted from this
+        /// recipe's page. Changelog recipes only — nil everywhere else, and nil
+        /// on a row written before this field existed, which is why the
+        /// collapse check needs a previous value and cannot fire on the first
+        /// sweep after an upgrade.
+        public var lastGoodEntryCount: Int?
         /// Consecutive sweeps in which this recipe was broken *or* degraded.
         /// Warnings count: `installURLUnresolved` is how both of the real
         /// one-click failures were found, and neither ever produced a `broken`.
@@ -126,10 +132,21 @@ public struct Baseline: Codable, Sendable {
         /// empty baseline on any decode error, and an empty baseline means every
         /// issue number is forgotten — so the next run reopens a duplicate of
         /// every issue that is already open. One renamed field would do it.
+        ///
+        /// ⚠️ **A new field must be added HERE as well as declared above.** Only
+        /// the decoder is hand-written; the encoder is synthesized, so a field
+        /// left out of this list is written to the file every sweep and read
+        /// back as nil every sweep. Nothing fails: the value is on disk, it
+        /// looks right to anyone opening the file, and the check that depends on
+        /// it simply never has a previous value to compare against.
+        /// `lastGoodEntryCount` shipped that way for the length of one live
+        /// test — unit tests never touch the disk, so only a real two-sweep run
+        /// showed it.
         public init(from decoder: Decoder) throws {
             let c = try decoder.container(keyedBy: CodingKeys.self)
             lastGoodVersion = try c.decodeIfPresent(String.self, forKey: .lastGoodVersion)
             lastGoodAt = try c.decodeIfPresent(Date.self, forKey: .lastGoodAt)
+            lastGoodEntryCount = try c.decodeIfPresent(Int.self, forKey: .lastGoodEntryCount)
             consecutiveActionable =
                 try c.decodeIfPresent(Int.self, forKey: .consecutiveActionable) ?? 0
             consecutiveInfra = try c.decodeIfPresent(Int.self, forKey: .consecutiveInfra) ?? 0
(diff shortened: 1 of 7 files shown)


jizhi0v0:
`/code-review 395 high` — 2 findings, both mine, fixed in af08dfd.

1. **`Reconcile.swift` — `(was N)` could never print.** `duo reconcile` loads the baseline `duo verify` wrote minutes earlier, and that write already folded this sweep's count into it, so `entry.lastGoodEntryCount` always equals the number printed beside it. A page that collapsed 20 → 1 would have filed an issue whose table read `entries parsed | 1` with no sign anything had moved — the opposite of what the row was added to say. Now just the count; the transition is in the warning below it, written by the sweep that saw both numbers. (`| last good |` has had the same property since long before this PR, and now says so.)

2. **`EntryCountCollapseTests.swift` — the nil-count case pinned an input `reconcile` cannot receive.** It fed a `.vendor` finding under a `changelog:` recipe id, but ids are namespaced by registry and `Baseline.entries` is keyed on the id, so two registries never meet on one row. The production sequence that actually depends on that guard is healthy → a night the page did not parse (`sweepChangelog`'s `guard let changelog` branch, no count) → the collapse. Rewritten as that; the `?? 0` mutation still kills it.

`make test` green (`MAKE_TEST_RC=0`).

PR #375, opened 2026-09-06
Refuse a Sparkle update that walks the marketing version backwards

jizhi0v0:
Closes the guard half of #368. A Sparkle feed could offer a copy a package that was
newer by build and older by name, and every gate downstream passed it — the package
is the vendor's, signed, notarized, and the only downgrade guard in the repo is
about architecture.

## Reproduced first, against the real feed

`coteditor.com/appcast.xml` (2026-09-06, 13 items, exactly one `prerelease`):

```
installed 7.1.0-beta.3/840  detect=beta  allowed=[nil]  ->  best=7.0.9/843  updateAvailable("7.0.9")
```

The vendor keeps only the newest prerelease, so `channel(ofInstalled:)` misses on both
passes, `allowedChannels` falls back to `{nil}`, `usableItems` ranks by build, and
`evaluate` compares builds. Nothing in that chain reads the marketing version.

## The fix is two halves, and they answer different questions

**Which item to name** — `SparkleAppcastSource.offerableItem`. When the head would walk
this copy backwards, step over it and take the first installable, provably-not-older
entry below. That is what lets a beta copy find the next beta sitting *under* a stable
patch in build order, which taking the head unconditionally hides.

**Whether it is an update** — `UpdateChecker.evaluate`. When the builds say newer, ask the
marketing version whether that means forward, and refuse if not. One-way: it can only
settle a row to up-to-date, never raise an update.

Gated on a new **stated** fact, `RemoteVersion.marketingMatchesBundle`, default false
meaning *not established*. Only `SparkleAppcastSource` sets it, because
`sparkle:shortVersionString` is the bundle's own `CFBundleShortVersionString` by
construction. Two live sources prove the gate is needed rather than defensive:
`XcodeReleasesSource` puts the display label "27.0 beta 6" in that field against an
installed "27.0", and Mozilla's probes report `155.0b5` against a bundle that says
`155.0` — a prerelease tag sorts below the release it leads, so both read as downgrades.

## ⚠️ The withheld release is still named

`latestVersion` never returns nil for a feed that had usable items. Returning nil lands
the row on `.unknown`, whose own documentation is "no source *covers* this app, nothing
was tried", rendered as a dead "—" with no retry — a false statement about an app whose
feed we had just read, and it takes the release notes and the release timeline with it.
Simulated over all 13 Sparkle feeds this Mac reads, with the installed build trimmed out
of each in turn, withholding the answer hit **12 (app, version) pairs**: Bartender 6.6.2
[...]

PR #396, opened 2026-09-06
Release 0.3.87

jizhi0v0:
Version bump plus the six translations of the 0.3.87 notes.

## What ships in 0.3.87

The notes were written by the PRs that landed in them; nothing is added here. In the order the section already has them:

- **The swallowed Update click** (#375 / #380) — a pre-install re-check that came back older than what the row offered reported the app as up to date and dropped it from the list.
- **Fork on its Develop channel** (#391) — the channel setting was read backwards, so a Develop copy followed the Stable feed and looked current.
- **Mac Performance Monitor's release notes** (#378, #381) — published in its repository, not in the feed we read.
- **CotEditor, on both lines** (#377, #379) — which line a copy follows comes from its version and from CotEditor's own prerelease setting.
- **App Store apps are never offered a download from elsewhere** (#383, #389) — a failed or empty store lookup now says the store is managing it, instead of falling through to another distribution's build.
- **A long error no longer pushes the list down** (#380) — two lines, full text on hover.

## Deliberately not in the notes

#392 (the `SparkleFeedCatalog` sweep) and #395 (the changelog entry count) are maintenance checks on our own nightly job. A user cannot perceive either, and "Under the hood" is for changes they could — so they get no line rather than a line that means nothing to a reader.

Same for #382, #384, #387, #390: release tooling, tests, and a lint.

## Translations are in this PR, not a follow-up

0.3.85 and 0.3.86 both shipped English-only and had to be back-translated afterwards (#372). In between, every non-English window showed two English entries stacked above translated ones. `publish-release.sh` warns about untranslated languages on every run; it had been warned twice and shipped anyway both times.

Paragraph counts match English (6 each), every paragraph opens with its bold lead sentence, and the covered run stays hole-free. Verified by mutation: dropping one `ja` paragraph fails `everyShippedLanguageHasANotesFileMatchingEnglishParagraphForParagraph` on `5 == 6`.

`make test` green (`MAKE_TEST_RC=0`).


jizhi0v0:
`/code-review 396 high` — 1 finding, mine, fixed in b48a62b.

**Every user-visible change in this release had a note except #375's downgrade guard**, and that omission was not a decision — the PR body above accounts for #392/#395 and the tooling PRs as imperceptible and never mentions #375 at all.

It earns one. `b35b57f` records a user-visible consequence in the app audit: "a copy on an abandoned prerelease train is no longer carried across to the maintained line, and no row state says so." A reader can meet that across the upgrade with nothing in the notes accounting for it.

Written as the benefit rather than the cost, since the benefit is what the reader gets — an offer that would have walked their app backwards is no longer made — and kept distinct from the section's first note, which is about the re-check at the moment of the click rather than about the offer itself. Seven paragraphs now, in all seven files; the localization test still passes.

PR #420, opened 2026-09-07
Stop two concurrent smoke-test runs from deleting each other's download

jizhi0v0:
`installPipelineDryRun` fails intermittently on machines with real Sparkle apps — measured 3 failures in 5 runs — always with the same shape on the Helium candidate:

```
".duo-download-<uuid>.partial" couldn't be moved to
"DuoUpdaterTest-net.imput.helium-b233bd2839efd54c" because either the former
doesn't exist, or the folder containing the latter doesn't exist.
```

## Cause

The scratch directory name is process-independent in **both** halves. `scratchSlug` is deliberately stable across launches (a SHA-256 of the installed app's path, so a crashed install can reclaim its dir by name), and `FileManager.temporaryDirectory` is per-*user*, not per-process. So every test process on the machine running this test for the same app named the **same** directory — and the test's `try? removeItem(at: workDir)` on the way in and its `defer { try? removeItem(at: workDir) }` on the way out were aimed at each other. Each run deletes the other's in-flight download; the loser's `Downloader.finalizePartial()` renames into a directory that is gone.

Concurrent runs are routine here — this repo is worked in several worktrees at once, the same hazard `scripts/derived_data_path.py` exists to solve for xcodebuild.

## Not a timeout, and not the vendor

Every deleter of the `.partial` was traced: the pre-transfer sweep, `startFresh()` (which recreates it immediately), and `download`'s `defer` (which runs after it returns, by which point a successful transfer has already moved the file). `finalizePartial` runs inside `didCompleteWithError` *before* the continuation resumes. No in-process path leaves the partial missing at rename time — the deletion comes from outside the process. The observed correlation with slow runs is a co-symptom: the competing run both halved the bandwidth and destroyed the directory.

Worth calling out separately: the failure is reported as `install path broken for: Helium`. A collision arrives dressed as vendor breakage, in the one test whose job is to report vendor breakage.

## Fix

Per-run UUID in the directory name, so it is nobody else's path, and the destructive pre-clean is gone. A unique name can no longer be reclaimed by name, so a run killed mid-download would leak ~120 MB per crash — stale siblings are swept instead, **age-gated at six hours**, because "not mine" is exactly the judgement that caused the collision, and a live download can legitimately run long on a slow link.

## Verification

- **Red first.** Two concurrent `swift test --filter installPipelineDryRun`, given separate `--scratch-path`s so SwiftPM's build lock doesn't serialise them: both failed with the exact message above, on that exact directory name. (Without separate scratch paths they serialise and both pass — that masked the bug on two earlier attempts.)
- Solo runs before the fix: 4/4 pass, ~20s each.
- **Green after.** Same concurrent pair, run twice: 4/4 pass, 0 leftover scratch dirs in `$TMPDIR`.
- `scripts/run-with-hang-report.sh 1800 make test` → exit 0, 2433 tests in 201 suites + 223 app tests, all python gates green.

## Not verified

Which process was competing during the originally reported 5 runs. Nothing periodic is registered for it (no `DUO_INSTALL_SMOKE` launchd job on this machine), so another worktree's `make test` is the likely partner — inference, not measurement.

This stays invisible to CI either way: the test guards on `!targets.isEmpty` and a GitHub runner has no Sparkle apps installed, so it returns early there.

### DIFF
--- a/DuoUpdaterCore/Tests/DuoUpdaterCoreTests/InstallPipelineSmokeTests.swift
+++ b/DuoUpdaterCore/Tests/DuoUpdaterCoreTests/InstallPipelineSmokeTests.swift
@@ -80,11 +80,33 @@ private func verifyGates(
     // built a nested `DuoUpdaterTest-/Users/.../Foo.app` tree whose empty parent
     // directories outlived the run. Harmless once a day by hand; not something
     // to leave behind on a runner that now does this for every app, nightly.
+    //
+    // The UUID is what keeps the name from being a SHARED path. `scratchSlug` is
+    // deliberately stable across launches (it is a SHA-256 of the installed app's
+    // path, so a crashed install can reclaim its dir by name), and
+    // `temporaryDirectory` is per-user, not per-process — so without it, every
+    // process on this Mac running this test for the same app names the SAME
+    // directory, and the two lines below are then aimed at each other: each run
+    // deletes the other's in-flight download, and the loser's `Downloader` fails
+    // its final rename with "either the former doesn't exist, or the folder
+    // containing the latter doesn't exist". Measured 2026-09-07: two concurrent
+    // `swift test --filter installPipelineDryRun` (separate `--scratch-path`s, so
+    // SwiftPM's build lock doesn't serialise them) both failed that way, on
+    // `DuoUpdaterTest-net.imput.helium-b233bd2839efd54c`. That is a collision, not
+    // a vendor problem — but it arrives dressed as one, in a test whose whole job
+    // is to report vendor breakage. Concurrent runs are routine here: this repo is
+    // worked in several worktrees at once, which is the same hazard
+    // `scripts/derived_data_path.py` exists to solve for xcodebuild.
     let workDir = FileManager.default.temporaryDirectory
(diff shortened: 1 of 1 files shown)


jizhi0v0:
Review pass found one issue in the age gate I added, now fixed in 00b2eeb8.

**`sweepStaleScratchDirs` dated a live download from when it started, not from its last progress.** A directory's mtime records changes to its *entries*, not writes into them — measured: creating a `.partial` stamps the dir, appending 5 MB to that file three seconds later moves the file's mtime and leaves the dir's untouched. So a transfer running past the six-hour gate (a large payload on a throttled link, across up to `maxAttempts` resumes) looked abandoned to a run starting beside it and would be deleted mid-flight — the exact collision this PR removes. The comment justifying the gate named that case explicitly, so the guard was right and the discriminator was wrong.

Now keyed on the newest modification one level inside the directory. Verified with a staged pair in `TMPDIR` — a dir whose own mtime is 10h old with a child written just now, and one where both are 10h old: the live one survives, the stale one is deleted. Mutated back to reading only the dir's own mtime and the live one is deleted, so the check measures this code rather than the fixture.

PR #421, opened 2026-09-07
CLAUDE.md: the install path is machine-exclusive, and what a harness loses by copying its naming

jizhi0v0:
The production scratch dirs are named process-stably — `DuoUpdater-<scratchSlug>-<version>` ([SparkleInstaller.swift:99](DuoUpdaterCore/Sources/DuoUpdaterCore/Install/SparkleInstaller.swift:99)), `DuoUpdater-vendor-…` ([VendorInstaller.swift:128](DuoUpdaterCore/Sources/DuoUpdaterCore/Install/VendorInstaller.swift:128)) — each followed by a `try? removeItem(at: workDir)`. **That shape is not safe on its own**: two processes installing the same app delete each other's download, and the version in the name doesn't help, because concurrent installs of one app carry the same version.

What makes it safe is `InstallLock` — a machine-wide `flock` shared by the menu-bar app and the `duo` CLI. Three properties are load-bearing: it covers the **download**, not just the swap; it **refuses rather than queues** (the CLI exits naming the holder's pid); and it is **reference-counted in-process**, so the app's own `InstallPermits(applies: 2)` concurrency shares one claim and only another *process* is turned away.

Worth writing down because #420 was exactly that invariant being lost by a copy: `installPipelineDryRun` reused the production naming scheme **without** the mutex, and two worktrees running it at once deleted each other's in-flight package — surfacing as `install path broken for: <app>`, a collision wearing vendor breakage's clothes in the one test whose job is to report vendor breakage.

The section also records the two measurements from that investigation a reader would otherwise have to redo:

- SwiftPM's build lock serialises two `swift test` runs in one checkout (two ~20s runs, 43s total, both green), so reproducing a cross-process bug needs a second `--scratch-path` — otherwise you're measuring a queue. This produced two false-green repro attempts.
- A directory's own mtime freezes while bytes stream into a file inside it (creating `.partial` stamps the dir; appending 5 MB three seconds later moves the file's mtime to …307 and leaves the dir's at …304), so it cannot answer "how long since this run made progress".

Per the repo's own "改一处说法之前,先数它有几份" rule: only one copy of this claim exists. README's mention of the lock describes the CLI's user-facing refusal — a different statement — and no skill under `.claude/skills` or `.agents` mentions it.

`make test` green (exit 0, 2433 + 223 tests, all prose/skill gates).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -381,6 +381,41 @@ App 只留接线;接线本身要可测,就放进 `ScanRowAssembly` 这类无 UI 
   `$DD/app-tests.log` 并在失败信息里报出路径(第一版让人「重跑一遍看完整日志」,而重跑
   给出的是同样被过滤的 40 行——那条提示是假的)。
 
+## 装 app 的路径是全机独占的,靠 `InstallLock`
+
+生产的 scratch 目录名**跨进程稳定**:`DuoUpdater-<scratchSlug>-<版本>`(Sparkle)、
+`DuoUpdater-vendor-<scratchSlug>-<版本>`(Vendor),而且紧跟着一句
+`try? removeItem(at: workDir)`。这个形状**自己不安全**——两个进程同时装同一个 app
+就会互删(名字里那个版本号不解决问题:同时装的必然是同一个版本)。
+
+让它安全的是 `InstallLock`(`Install/InstallLock.swift`):一把全机 `flock`,菜单栏 app 和
+`duo` CLI 共用。三个性质都承重:
+
+- **覆盖下载,不只是 swap**——持有者可能正下到一半的 400 MB;
+- **拒绝而不是排队**——CLI 直接退出并报出持有者的 pid,「看起来卡住」比一句实话更糟;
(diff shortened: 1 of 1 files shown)


jizhi0v0:
Review pass found one issue, fixed in 9717126b.

The section justified "the version token in the scratch dir name doesn't separate the two processes" with an absolute — that concurrent installs are *necessarily* the same version. The code doesn't guarantee that: the menu-bar app and `duo install` resolve the feed independently, so a vendor rolling forward between the two resolutions does give them different names. The conclusion stands and is all the safety argument needs (the version token can't be *relied on* to separate them); only the reasoning was overstated — in a file whose own first rule is about absolutes asserted without checking.

Also re-verified the load-bearing claim while reviewing, since the whole section rests on it: [AppListModel.swift:2736](App/Sources/AppListModel.swift:2736) takes `ProcessInstallLock.shared.claim()` **before** `performInstall`, and releases after — so the lock covers the download on the app side too, not just the CLI's.

Issue #546, opened 2026-09-12
No OS-floor judge on either App Store install route


jizhi0v0:
`SignatureVerifier` gate 6 refuses a downloaded bundle whose `LSMinimumSystemVersion` is above this Mac's OS. It runs from `verifyInstallArtifact`, which only the Sparkle and vendor installers call. **Neither App Store route reaches it**: `mas` hands the install to the store, and the AX route drives the store's own product page.

So for a store app whose newest build raises its macOS floor past this Mac:

- `latestMacCompatible` is `nil` (the compat scrape answers "installable on a Mac", not "installable on *this* Mac)
- → `AppStoreGate.resolve` returns `.none`
- → the row offers **Update**
- → the App Store refuses, and nothing in the row ever explains why

This is a missing judge, not a duplicated one — there is no second opinion to reconcile, the question is simply never asked on these two routes.

## The data is already in hand

The floor is on the same HTML `extractMacCompatibility` already fetches, in the `information` shelf's Compatibility annotation, per platform:

```json
{"$kind": "AnnotationItem", "text": "Requires macOS 15.6 or later.", "heading": "Mac"}
```

(Real, from Nowdex's page, measured 2026-09-12.) So this costs no extra request — it is another field off a body we download anyway, the same way `appPlatforms` was added in #<this PR>.

## Sketch

Extend the reading `extractMacCompatibility` returns with `minimumMacOS: String?`, and add an `AppStoreGate` case for "needs a newer macOS". Keep it `String?` and fail open on nil, for the same reason `latestMacCompatible` is `Bool?`: a scrape failure must never hide a real update.

Compare through `VersionComparator` against `HostOS.numericVersion()` — the shared definition that exists precisely so the detection-time gate and the install-time gate cannot disagree. Note that would make this a **third** consumer of `HostOS`, which is the point: today the App Store routes are the only ones with no OS opinion at all.

## Not currently reachable on the two machines measured

Every Sparkle feed both machines read was pulled and every `minimumSystemVersion` / `maximumSystemVersion` extracted; the highest floor anywhere is `26.3`, against hosts on 27.0 and 26.6. No App Store app on either machine is currently blocked this way. Latent, not live.

jizhi0v0:
PR #552 已合并（squash `d349a761`）。

## 落地的东西

- `MacCompatibilityReading.minimumMacOS` —— 从**同一张已经在抓的产品页**读下限，不多发请求。
[...]

Issue #545, opened 2026-09-12
MacAppStoreSource.resolve() routes on a disk fact when the listing already answers it


jizhi0v0:
`MacAppStoreSource.resolve()` picks which release line to read from a fact about **this Mac's disk** (`app.isiOSAppOnMac`), while the fact that actually settles the question — does this listing publish a Mac binary — is already fetched on one of the branches and then thrown away.

```swift
// MacAppStoreSource.swift
if result.isNativeMac { return try await nativeMacVersion(...) }          // network: kind
if app.isiOSAppOnMac  { return try await remoteVersion(..., checkMacCompat: true) }  // DISK
if let trackId = result.trackId { return try await iosOnMacVersion(...) }
```

Same shape as the bug fixed in #<this PR>: a narrow fact standing in for a wide question. That fix moved `appPlatforms` into the compatibility **gate**; the **routing** still can't see it.

## Discriminating input

Nowdex, `org.randynamic.nowdex`, trackId `6791450777`. Measured 2026-09-12, `us` storefront:

- lookup `kind == "software"`, `version 1.1.2`
- plain product page: `appPlatforms == ["mac","phone","pad"]`, `isIOSBinaryMacOSCompatible == false`, `mostRecentVersion == "Version 1.1.2"`
- `?platform=mac` page: **no `mostRecentVersion` shelf at all** (HTTP 200, 29075 bytes, stable across 3 fetches, `isIncomplete: false`, and the page's own `shelfOrderings` still lists the shelf — Apple returned no data for it on the mac platform view)

Two users, same listing, same minute, same storefront:

- copy has `Wrapper/` → branch 2 → `RemoteVersion(shortVersion: "1.1.2")`. Row shows the update.
- copy is the native Mac build (no `Wrapper/`, has `_MASReceipt`) → branch 3 → `iosOnMacVersion` scrapes `?platform=mac`, the parse finds nothing, **returns nil** → the row falls through to `.appStoreManaged` with no version at all.

## Rarity of the trigger

Widened to 42 distinct listings (both page variants each, `us`, browser UA): **41 of 42 carry a parseable `mostRecentVersion` shelf on `?platform=mac`. Nowdex is the only one that does not** (~2.4%).

## Second consequence at the same site

[...]

PR #548, opened 2026-09-12
App Store and TestFlight: stop answering wide questions with narrow facts

jizhi0v0:
Two apps on two of my Macs were being told the wrong thing about themselves, in the same shape: **a narrow fact standing in for a wide question, where the two opposite truths look identical in that fact.**

## 1. Nowdex showed "Not supported on this Mac" the day it gained Mac support

`isIOSBinaryMacOSCompatible` answers "does the *iOS binary* run on macOS", not "does this app support Mac". A listing that ships its own Mac binary answers `false` and is fully supported — so the flag flips the day a developer **adds** Mac support natively, identically to the day one drops it.

Measured 2026-09-12 across 18 live listings (`us`, Aqara Home also `cn`): the page's own Compatibility section names Mac exactly when `isIOSBinaryMacOSCompatible == true || appPlatforms.contains("mac")`, no exception either way.

- flag `true`, no `"mac"` platform, Mac listed: Overcast, Aqara Home
- flag `false`, `"mac"` platform, Mac listed: nowdex, GoodNotes, WhatsApp, Kindle, CARROT Weather, Home Assistant, Bear, Things 3, Day One, Todoist, Microsoft To Do
- flag `false`, no `"mac"` platform, Mac NOT listed: Instagram, Discord, Facebook, Scriptable, ChatGPT, Telegram, Fantastical, Microsoft Edge

A `false` verdict now needs **both** readings. A lone `isIOSBinaryMacOSCompatible == false` is the majority shape among Mac-supported listings, so it is unsound on its own: if Apple renames `appPlatforms`, every app with a native Mac build lands on `nil` (offer the update, maybe wrongly) rather than `false` (refuse it, definitely wrongly).

Red→green on the machine that reported it, binary as the only change:

```
old: {"name":"Nowdex","installedVersion":"1.1.0","latestVersion":"1.1.2","route":"manual"}
new: {"name":"Nowdex","installedVersion":"1.1.0","latestVersion":"1.1.2","route":"in-place"}
```

## 2. An App Store purchase was handed to TestFlight

`isManaged` answered a membership question out of the table of every mac row the DB holds — the *availability* table. The iOS branch has always filtered on `ZINSTALLSTATUSRAW = 1`; the mac branch never did.

`cam.thescreen` is an App Store purchase (`_MASReceipt` present, `kMDItemAppStoreReceiptType = Production`, `Authority=Apple Mac OS Application Signing`) while the DB held `cam.thescreen | 1.3.1 | 20260903024310 | platform 3 | ZINSTALLSTATUSRAW 0` — a beta the user has access to carrying the same `CFBundleVersion` as the shipping build, which is what happens whenever a beta is promoted unchanged. Build match alone said "TestFlight install"; `AppScanner`'s `isTestFlight` is an `||` so the Production receipt could not veto it; `UpdateChecker` then returned from the TestFlight branch before any source ran, and the App Store could never answer for that row.

Fails **open** when the status is unknown (a schema without the column, or a fixture that does not say): losing the column must cost the precision, not the signal. `macInstalledRows` is `[Row]?` for exactly that — nil is "unknown", not "installed none".

Red→green, and verified against the live database rather than through `AppScanner`'s `||`, because all four genuine TestFlight installs on this Mac are already tagged by an earlier term and would have masked a broken `isManaged`:

```
lettera 4266 → true    mithka 1155 → true    notability 7133 → true
paste 29818681 → true (and still offered 29819872)
cam.thescreen 20260903024310 → false, while latest() still sees the row
```

## 3. "Is it a store copy" is a fact about the copy, not the app

Seven places argued from Keka. First "Keka is a store copy carrying a `SUFeedURL`", then a commit "corrected" it to "Developer ID-signed with no `_MASReceipt`, and the `isMAS` derivation is byte-identical then and now". **The correction is the second error, and it is the same error.** Two Macs, Keka 1.6.7 on both: one has a 2026-07-01 `_MASReceipt` with `Authority=Apple Mac OS Application Signing` *and* `SUFeedURL = https://u.keka.io`; the other is Developer ID with no receipt.

That matters beyond wording — "a store copy carrying its own feed" is a real arrangement, so the hole the original sentence described is real, and the correction argued it away on the strength of a different laptop. `CLAUDE.md` used this correction as the worked example for 「转引一条实测也是断言」, so it is updated rather than deleted, and gains the rule it was missing.

`grep -rn Keka` first: seven sites, not the five a first pass found.

## Verification

- `make test` green (`2682 tests / 226 suites` + `280 tests / 34 suites`, exit 0)
- Five mutations, each reddening only its own case: reverting to the flag alone; weakening `readBothSignals` to `macSupported != nil`; searching the whole payload instead of anchoring at `data[0].data`; taking the unfiltered mac rows; treating unknown install status as none
- `duo verify --only com.hammerandchisel.discord` green live against the stricter sweep gate
- Both fixes deployed and confirmed on the two machines they were found on

## Deliberately out of scope

[#545](https://github.com/jizhi0v0/duo-updater/issues/545) — `resolve()` still routes on `Wrapper/` when the listing already answers it. Note the fix must carry a lookup fallback or it turns working rows blank.
[#546](https://github.com/jizhi0v0/duo-updater/issues/546) — neither App Store install route has an OS-floor judge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@ Versions before 0.3.80 are the old long-form style; leave them as shipped.
 
 ## 0.3.93
 
+**Apps that added a Mac version no longer show as "not supported on this Mac."** An iPhone or iPad app you run on Apple silicon was mistakenly flagged the moment its developer shipped a real Mac build — the one change that makes the update more available, not less.
+
+**An App Store app you also beta-test is no longer mistaken for a TestFlight build.** When a developer promoted a beta unchanged, the two carried the same build number and your purchased copy was handed to TestFlight — so the App Store could never offer it an update.
+
 **The Requests window now shows how far back its log actually reaches, and marks date ranges it can't fully cover.** Before, picking "Last 30 days" on a log that only went back a few hours looked exactly like picking "Last 24 hours," with nothing on screen explaining why.
 
 ## 0.3.92
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -214,14 +214,28 @@ xcodebuild 的 SQLite 锁**,症状是 `database is locked` 或者干脆卡住—
 
 - **「转引一条实测」也是断言,而且是最容易漏的那种。** 引用一句带日期、带数字的实测,
   等于你自己也说了一遍——现有规矩只管「写新断言前先量」,不管「引用旧断言算不算」。
-  2026-09-06 实测过一次代价:`UpdateChecker` 里一句「measured 2026-08-29, Keka is a store
-  copy carrying `SUFeedURL`, 1 of the 22 store apps on this machine」写下来那天就是错的
-  (Keka 是 Developer ID 签的、没有 `_MASReceipt`,而 `isMAS` 的推导那天和今天逐字相同),
-  它却是「不给 `SparkleAppcastSource` 加商店闸」的**唯一记录理由**,撑了一周,并且被
-  转抄进另外三个文件——其中一次是我,而我当时的行为在旧规矩下完全合规。
+  `UpdateChecker` 里一句「measured 2026-08-29, Keka is a store copy carrying
+  `SUFeedURL`, 1 of the 22 store apps on this machine」是「不给 `SparkleAppcastSource`
+  加商店闸」的**唯一记录理由**,撑了一周,并且被转抄进另外三个文件。
   要么复测,要么写明「转引自 X,未复测」。
+
+- ⚠️ **这一条自己也栽在下一条上,2026-09-12 实测。** 上面这段原先接着写:
+  「(Keka 是 Developer ID 签的、没有 `_MASReceipt`,而 `isMAS` 的推导那天和今天逐字相同)」
+  ——那个**更正本身是错的**,而且错法和它要纠正的那句一模一样。
+  两台机器同为 Keka 1.6.7:一台 `/Applications/Keka.app/Contents/_MASReceipt/receipt`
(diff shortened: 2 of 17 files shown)


jizhi0v0:
**Blast radius, measured rather than argued.**

Cross-checked every installed bundle on two Macs against their TestFlight databases, looking for the one thing that can regress: an app whose mac row carries `ZINSTALLSTATUSRAW = 0` *and* whose on-disk `CFBundleVersion` equals that row's build, i.e. an app the old code tagged and the new code does not.

Exactly one turns up — `cam.thescreen`, the store purchase this PR is about. No genuine TestFlight install loses its tag: every mac row the second machine marks `= 1` corresponds to an app that really is TestFlight-signed there, and the four on the first machine answer `isManaged == true` when probed against the live database directly (which matters, because all four are also caught by `AppScanner`'s receipt term and would have masked a broken `isManaged` if checked only end-to-end).

**The near-miss is the more interesting result.** DingTalk (`5ZSL2CJU2T.com.dingtalk.mac`) is the same shape as ScreenCam — App Store purchase, `_MASReceipt` present, `Authority=Apple Mac OS Application Signing`, and a mac TestFlight row for a beta the account has access to. The old code did not mistag it, for one reason: the installed build is `58012001` and the row's is `30112376`, so the build match happened to miss.

That is luck, not a guard. The two numbers coincide exactly when a vendor promotes a beta unchanged, which is the normal case rather than the exotic one — so the defect was not a single-app curiosity, it was one coincidence away from a second instance on a different machine.

Issue #351, opened 2026-09-05
Download gate wedges the whole test process on a hosted runner


jizhi0v0:
`vendorDownloadPassesSignatureGate` wedges the whole test process on a hosted runner

## What happens

With `DUO_DOWNLOAD_GATE=1` set on `macos-latest`, the gate picks Firefox (the
only candidate installed on the image), prints `· resolving version`, and the
**entire `swift test` process stops emitting anything** ~2.4s later. Two runs,
same shape:

| run | gate starts | last test finishes | silence until |
|---|---|---|---|
| [33950064978](https://github.com/jizhi0v0/duo-updater/actions/runs/33950064978) | 06:35:10 | 06:35:19 | 07:21 (cancelled, 46 min) |
| [33952614948](https://github.com/jizhi0v0/duo-updater/actions/runs/33952614948) | 07:32:05 | 07:32:07 | 07:48 (cancelled, 16 min) |

It never reaches `· downloading`, so nothing is being downloaded — it wedges
inside `LiveProbe.remote` → `VendorProbeSource.probeDiagnostic(for:)` for
`org.mozilla.firefox`.

Not reproducible locally: the same bundle id resolves in seconds on the
author's Mac (`v=155.0.1 kind=dmg`), and the full suite is green there.

## Why the timeout did not save it

A 5-minute `AppRestarter.firstToFinish` was added around the gate and **did not
fire**. Its timer is itself `Task { try? await Task.sleep(...) }`, so it runs on
the same cooperative thread pool that has stopped scheduling. An async timeout
cannot bound a starved async runtime; a watchdog for this has to be out of band
(a dedicated `Thread`, or a process-level timeout in the workflow).

That the whole process — not just this test — goes quiet within 2.4s is the
main clue: a single hung network call would leave the other tests finishing.
`macos-latest` is 3-core, so the pool is narrow.

## Why it matters

`test` is a required status check. A wedge here blocks every merge for the full
60-minute job budget with no message.

## Current state

`DUO_DOWNLOAD_GATE` is **not** set in `ci.yml`, so the gate runs nowhere by
default. That is a real coverage hole and it is deliberate for now: this is the
only test that puts a downloaded artifact through sha512, extraction, code
signature and the Team ID match over real bytes. Run it on demand with
`DUO_DOWNLOAD_GATE=1 make test` (~161 MB on a machine with ChatWise and VLC
installed).

## Next steps

[...]

PR #565, opened 2026-09-12
Backup sealing and the interrupted-swap sweep's install lock

jizhi0v0:
Fixes two findings from the 2026-09-13 review. Both are in the install/backup path.

## L10 — `Install/BackupManifest.swift:47-76, 83-94`

`unreadableFiles` classified an unreadable file as "sealed" only if it was a key in
the bundle's `CodeResources` `files`/`files2`. Those enumerate **resources**; the
bundle's own `Info.plist`, its main executable and the `_CodeSignature/` envelope are
hashed into the CodeDirectory instead and appear in neither. An unreadable main
executable was therefore read as runtime droppings — `ditto` skipped it, the manifest
certified the result, and `BackupStore.save` returned `.savedWithoutRuntimeState`
(success) for a rollback point that is a bundle with no executable.

### The claim, verified before fixing it

Step 1 — are those keys really absent from `files`/`files2`? (plistlib, read-only):

```
--- /Applications/Safari.app  (files+files2 = 1575 keys)
    'MacOS/Safari'          in seal? False
    'Info.plist'            in seal? False
    'PkgInfo'               in seal? False
    '_CodeSignature/CodeResources'  in seal? False
--- /Applications/DuoUpdater.app  (files+files2 = 22 keys)   … all four False
--- /Applications/Keka.app  (files+files2 = 478 keys)        … all four False
```

(Nested bundles' `Info.plist`/`MacOS/<exe>` *are* listed — e.g. Safari's appex — so
the rule is specifically about the outer bundle's own copies.)

Step 2 — absence from the list is not the same as "unsealed", so ask `codesign`.
Each row is a fresh `ditto` copy with one file mutated:

```
--- /Applications/Keka.app
[baseline, untouched]                            verify rc=0
[mutated Contents/Info.plist]                    verify rc=1  invalid Info.plist (plist or signature have been modified)
[mutated Contents/MacOS/Keka]                    verify rc=1
[mutated Contents/_CodeSignature/CodeResources]  verify rc=1  invalid resource directory (directory or signature have been modified)
[mutated Contents/PkgInfo]                       verify rc=0
[deleted Contents/PkgInfo]                       verify rc=0
--- /Applications/DuoUpdater.app   (identical results, same six rows)
```

So three of the four are sealed and **`PkgInfo` is not** — it is in neither place and
stays skippable. That is what keeps the fix from widening into "every top-level file".

### The fix

A new `directlySealedKeys(of:interior:fileManager:)` contributes `Info.plist` and
`MacOS/<CFBundleExecutable>` (or the flat `<exe>` for a wrapped iPhone/iPad bundle,
decided by what is on disk), and `unreadableFiles` also treats anything under
`_CodeSignature/` as sealed. The `UnreadableFiles` doc comment at :30-39 now says
that "the seal covers it" ≠ "`CodeResources` lists it".

### Red → green

Three new tests in `BackupStoreTests`, fixtures named `ZZFixture-*` with a
`!fileExists` guard. Before the fix:

```
✘ theMainExecutableCountsAsSealedThoughCodeResourcesOmitsIt() failed … with 3 issues
    classified.sealed → []          (expected ["Contents/MacOS/App"])
    an error was expected but none was thrown   ← save happily stored the broken copy
✘ theInfoPlistAndSignatureDirectoryCountAsSealed() failed … with 2 issues
    classified.unsealed → ["Contents/Info.plist", "Contents/_CodeSignature/CodeDirectory"]
✔ pkgInfoIsNotSealedSoAnUnreadableOneStillAllowsTheBackup() passed   ← negative control
```

After: `✔ Test run with 35 tests in 1 suite passed`.

### Mutation checks (each reverted afterwards)

| mutation | result |
|---|---|
| drop `\|\| directlySealed.contains(sealKey)` | ✘ `theMainExecutableCountsAsSealed…` (3 issues), ✘ `theInfoPlistAndSignatureDirectory…` (2) |
| drop `\|\| sealKey.hasPrefix("_CodeSignature/")` | ✘ `theInfoPlistAndSignatureDirectory…` (2 issues) |
| over-widen: add `"PkgInfo"` to the set | ✘ `pkgInfoIsNotSealedSoAnUnreadableOne…` (2 issues) |

The third is the one that matters for scope: the negative control goes red if someone
later decides the whole top level is sealed.

## L13 — `Install/InPlaceSwap.swift:63-99`

`recoverInterruptedSwaps` deletes `*.duoupdater-new` / `.duoupdater-staged-*` and
promotes `*.duoupdater-old` over the live app without holding `InstallLock`, while
the app's install and rollback and the CLI's install and backup restore all claim it.
Those suffixes are exactly what a live install parks beside its target, so the
menu-bar app's first refresh could delete the staging directory a `duo` in a terminal
is still `ditto`-ing into.

It now claims `ProcessInstallLock` first and, when refused, logs at `.notice`
("another installer holds the install lock, sweep deferred: …") and returns. The
claim is non-blocking because `InstallLock.acquire` always is (`LOCK_EX | LOCK_NB`) —
a launch-time sweep has nothing to wait for, and it will find the same leftovers next
launch. The sweep body moved verbatim into a private `sweepInterruptedSwaps(in:)` so
the lock brackets all of it; the public entry point is now `async` and takes
`lock: ProcessInstallLock = .shared`.

**What the lock does not cover, and the doc comment now says so:** `ProcessInstallLock`
is reference-counted per process, so an install running in *this* process joins the
claim rather than refusing it. That is the documented refcount (it is what lets the app
apply two updates at once); this sweep is serialised only against other processes.

### Test seam

`ProcessInstallLock.init(url:)` is already public, so no new seam was needed. The
"other holder" is simulated with a second `InstallLock.acquire(at:)` on the same path:
`flock` is scoped to the open **file description**, not the process, so the second
acquire is refused exactly the way a second process would be — deterministic, no
helper process, no sleeps. The two pre-existing sweep tests now pass a per-test scratch
lock too, so they never touch the machine-global `install.lock` (a test taking that
would interfere with a real install and go red because of one).

### Red → green

New: `InterruptedSwapSweepLockTests.theSweepDefersWhileAnotherProcessHoldsTheInstallLock`.
Before the fix the API does not exist:

```
InPlaceSwapTests.swift:386:28: error: extra argument 'lock' in call
InPlaceSwapTests.swift:392:28: error: extra argument 'lock' in call
```

After: `✔ Test run with 10 tests in 2 suites passed`.

### Mutation check

Changed `try await lock.claim()` to `try? await lock.claim()` so a refusal no longer
skips the sweep:

```
✘ theSweepDefersWhileAnotherProcessHoldsTheInstallLock() recorded an issue at
  InPlaceSwapTests.swift:395:9: Expectation failed: fm.fileExists(atPath: leftover.path)
  ↳ "the sweep must leave another installer's staging directory alone"
```

Restored, green again. The test's second half is its own vacuity guard: once the
holder releases, the same call *does* delete the leftover, so the test cannot pass by
the sweep simply doing nothing.

## `make test`

```
$ make test > /tmp/mt-sealsweep.log 2>&1; echo exit=$?
exit=0
$ grep -E "Test run with" /tmp/mt-sealsweep.log
━ Test run with 2781 tests in 229 suites passed after 20.998 seconds with 2 known issues.
✔ Test run with 280 tests in 34 suites passed after 0.146 seconds.
✓ App tests — 34 of 34 cases executed
✓ version comparisons discriminate — 261 files …
✓ no machine-population claims in code comments — 531 files
```

No `make gallery` run: nothing under `App/Sources` that draws a row was touched (the
only App change is one `await` in `recoverInterruptedSwapsOnce`). No recipe or parser
rule changed, so no `duo verify`.

## Behaviour a user could notice

- A backup of an app with an unreadable main executable or `Info.plist` is now refused
  instead of silently stored as a rollback point that cannot run. Rare input (a `0700`
  main executable), and the previous "success" was the worse outcome.
- The launch recovery sweep is skipped when another process is installing. It runs
  again next launch, and the leftovers it would have cleaned belong to that install.

## Declined / out of scope

Nothing declined — both findings hold. Two things noticed and deliberately not fixed:

1. `recoverInterruptedSwaps` still does its `SecStaticCodeCheckValidity` work on the
   cooperative pool (the `#351` shape; it is row 4 of the review's P1 table and belongs
   to whoever owns P1). Making the entry point `async` does not change that — the
   caller was already inside `Task.detached`, which is the same pool. The body is now a
   separate private function, which should make an `offCooperativePool` hop a one-liner.
   **Heads-up for the lead: P1 names this same function, so expect a conflict there.**
2. The in-process refcount means this sweep is not serialised against an install in the
   same process. Documented in the new doc comment rather than changed, since narrowing
   it would need a new `ProcessInstallLock` API and would affect every other caller.

### DIFF
--- a/App/Sources/AppListModel.swift
+++ b/App/Sources/AppListModel.swift
@@ -2257,6 +2257,13 @@ final class AppListModel {
     /// is an input method macOS can no longer load. Sweeping only `/Applications`
     /// would have left exactly that state unrecoverable, because a rotation's
     /// leftovers sit INSIDE the bundle rather than beside it.
+    ///
+    /// "Once" means once it has actually run. The flag is set before the task so a
+    /// second refresh cannot start a second sweep beside the first, and cleared
+    /// again if any root was skipped because another process held the install
+    /// lock — a deferred sweep has recovered nothing, and latching on it would
+    /// leave a genuine leftover from an earlier crash sitting there until the next
+    /// launch.
     private func recoverInterruptedSwapsOnce() {
         guard !didRecoverSwaps else { return }
         didRecoverSwaps = true
@@ -2266,8 +2273,14 @@ final class AppListModel {
             URL(fileURLWithPath: "/Library/Input Methods", isDirectory: true),
             home.appendingPathComponent("Library/Input Methods", isDirectory: true),
         ]
-        Task.detached(priority: .utility) {
-            for root in roots { await InPlaceSwap.recoverInterruptedSwaps(in: root) }
+        Task.detached(priority: .utility) { [weak self] in
+            var swept = true
+            for root in roots {
+                // Every root is attempted; one deferral must not skip the rest.
+                if await InPlaceSwap.recoverInterruptedSwaps(in: root) == false { swept = false }
+            }
+            guard !swept else { return }
+            await MainActor.run { self?.didRecoverSwaps = false }
         }
     }
 
--- a/App/Sources/AppListModel.swift
+++ b/App/Sources/AppListModel.swift
@@ -2267,7 +2267,7 @@ final class AppListModel {
             home.appendingPathComponent("Library/Input Methods", isDirectory: true),
         ]
         Task.detached(priority: .utility) {
-            for root in roots { InPlaceSwap.recoverInterruptedSwaps(in: root) }
+            for root in roots { await InPlaceSwap.recoverInterruptedSwaps(in: root) }
         }
     }
 
--- a/DuoUpdaterCore/Sources/DuoUpdaterCore/Install/BackupManifest.swift
+++ b/DuoUpdaterCore/Sources/DuoUpdaterCore/Install/BackupManifest.swift
@@ -37,6 +37,11 @@ public struct BackupManifest: Codable, Equatable, Sendable {
     /// recreates them. A file the seal *does* cover is shipped payload
     /// (EasyConnect's setuid helpers), and a copy without it is a broken app, so
     /// there is nothing honest to store.
+    ///
+    /// "The seal covers it" is not the same as "`CodeResources` lists it": the
+    /// bundle's own main executable, its `Info.plist` and the signature directory
+    /// are sealed by the CodeDirectory's special slots instead and appear in no
+    /// resource list. See ``directlySealedKeys(of:interior:fileManager:)``.
     public struct UnreadableFiles: Sendable {
         /// Relative to the bundle root, e.g. `Contents/mmkv.default`.
         public let sealed: [String]
@@ -54,6 +59,7 @@ public struct BackupManifest: Codable, Equatable, Sendable {
         // plist and its seal under `Wrapper/<Inner>.app/`. Read from the bundle
         // rather than assumed, so the seal keys below line up with the walk.
         let interior = BundleLayout.interiorPrefix(for: bundle, fileManager: fm)
+        let directlySealed = directlySealedKeys(of: bundle, interior: interior, fileManager: fm)
         let base = bundle.standardizedFileURL.path
         var sealed: [String] = []
         var unsealed: [String] = []
@@ -70,7 +76,13 @@ public struct BackupManifest: Codable, Equatable, Sendable {
             let sealKey = relative.hasPrefix(interior)
                 ? String(relative.dropFirst(interior.count)) : relative
             guard let sealedPaths else { sealed.append(relative); continue }
-            if sealedPaths.contains(sealKey) { sealed.append(relative) } else { unsealed.append(relative) }
+            if sealedPaths.contains(sealKey)
+                || directlySealed.contains(sealKey)
+                || sealKey.hasPrefix("_CodeSignature/") {
+                sealed.append(relative)
+            } else {
+                unsealed.append(relative)
+            }
         }
         return UnreadableFiles(sealed: sealed.sorted(), unsealed: unsealed.sorted())
     }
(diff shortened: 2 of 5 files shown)


jizhi0v0:
Review items addressed in `f1cedfb5`.

## 1. A deferred sweep must not latch `didRecoverSwaps` (confirmed — this was a regression my lock introduced)

`recoverInterruptedSwaps(in:lock:)` is now `@discardableResult … async -> Bool`, false only when the claim was refused. `recoverInterruptedSwapsOnce` sets the flag before the task (so a second refresh still cannot start a second sweep beside the first) and **clears it afterwards if any root was skipped**:

```swift
Task.detached(priority: .utility) { [weak self] in
    var swept = true
    for root in roots {
        // Every root is attempted; one deferral must not skip the rest.
        if await InPlaceSwap.recoverInterruptedSwaps(in: root) == false { swept = false }
    }
    guard !swept else { return }
    await MainActor.run { self?.didRecoverSwaps = false }
}
```

Note the loop deliberately does not short-circuit: a busy lock during `/Applications` must not skip the two input-method roots.

New Core test `theReturnValueReportsWhetherTheSweepActuallyRan`, asserted on an **empty** directory so the answer can only be "did I hold the lock", never "did I find anything". The existing test also now asserts both return values.

**Mutation** — make a refused claim report success (`return false` → `return true` in the catch):

```
✘ theReturnValueReportsWhetherTheSweepActuallyRan() recorded an issue at InPlaceSwapTests.swift:419:9
✘ theSweepDefersWhileAnotherProcessHoldsTheInstallLock() recorded an issue at InPlaceSwapTests.swift:395:9:
    Expectation failed: deferred == false
```

Restored → `✔ Test run with 2 tests in 1 suite passed`.

**Honest gap:** the App-side latch itself has no test. `AppListModel.swift` cannot join `DuoUpdaterAppTests` (it pulls in `SettingsView` and therefore all of SwiftUI, and `AppListModel.init` registers notifications, FSEvent streams and a check loop — CLAUDE.md's "App 层的测试 target" section). The Core half is gated; the three-line wiring above is convention plus your review, same as the rest of that file.

## 2. `lock.release()`

Kept unconditional rather than `defer`-style — `defer` cannot hold an `await`, so there is no defer form of this. What makes it correct is now written down, plus a note on the sweep function itself so the next person does not silently invalidate it:

```swift
sweepInterruptedSwaps(in: directory)
// Unconditional rather than deferred, and it stays correct only because
// the sweep below neither throws nor suspends: there is no early return
// and no cancellation point between the claim and here. (`defer` cannot
// hold an `await` anyway.) Anything added to that function that can bail
// out early has to release the claim on its way.
await lock.release()
return true
```

```swift
/// The sweep itself, once the lock is held. Deliberately non-throwing and
/// non-`async` — see the release note in the caller.
private static func sweepInterruptedSwaps(in directory: URL) {
```

## 3. Wrapped-bundle fixture

New `theMainExecutableOfAWrappedBundleCountsAsSealed`: no `Contents/` at all, a relative `WrappedBundle` symlink (the text `interiorPrefix` actually reads), `Wrapper/ZZFixture-Inner.app/` holding a flat `Info.plist`, `Assets.car`, `_CodeSignature/CodeResources` listing only `Assets.car`, and the executable `ZZWrapped` at that root, chmod `000`. Asserts `sealed == ["Wrapper/ZZFixture-Inner.app/ZZWrapped"]` and `unsealed.isEmpty`.

**Mutation** — flip `? nested : executable` to `? executable : nested`:

```
✘ theMainExecutableCountsAsSealedThoughCodeResourcesOmitsIt() failed … with 3 issues
✘ theMainExecutableOfAWrappedBundleCountsAsSealed() recorded an issue at BackupStoreTests.swift:676:13
    Expectation failed: classified.sealed == ["\(innerName)/ZZWrapped"]
```

Both branches of the ternary are pinned — each layout's test goes red when it gets the other layout's answer. Restored → `✔ Test run with 36 tests in 1 suite passed`.

## `make test`

```
$ make test > /tmp/mt-sealsweep3.log 2>&1; echo exit=$?
exit=0
━ Test run with 2783 tests in 229 suites passed after 20.285 seconds with 2 known issues.
✔ Test run with 280 tests in 34 suites passed after 0.164 seconds.
✓ App tests — 34 of 34 cases executed
✓ version comparisons discriminate — 261 files, 8 marketing-first pick(s), all display-only
✓ no machine-population claims in code comments — 531 files
```

(The run before this one died as `make: *** [test] Killed: 9` while linking `duo-product`, with the Core suite already green at 2782. Not this branch: that is the machine-wide `pkill -9 -f '…swift-test|xcodebuild'` in `scripts/run-with-hang-report.sh:91` — review finding P9 — reaching into another worktree. Reran clean.)

PR #566, opened 2026-09-12
Version comparisons: stop five decisions collapsing a version pair to one string

jizhi0v0:
Fixes five review findings that share one shape: a decision made on a single
version string where the two sides each carry a marketing version AND a build.

## Findings addressed

| # | Site | Change |
|---|---|---|
| L1 | `DuoUpdaterCore/.../Engine/UpdateChecker.swift:605-611` | folded-build fallback now requires the remote to BE the folded pair |
| L2 | `App/Sources/AppListModel.swift:6579-6600` | notification baseline keyed on the full `VersionSide`; rule extracted to Core as `NotifiedUpdateVersions` and tested |
| L14 | `DuoUpdaterCore/.../Sources/SelfUpdaterStaging.swift:130-133, 169-172` | both strictly-newer gates compare `VersionSide` pairs |
| L15 (timeline) | `DuoUpdaterCore/.../Releases/ReleaseTimelineStore.swift:101` | `record` dedupes on version + vendor date |
| extra | `DuoUpdaterCore/.../Engine/UpdatePolicy.swift:624-626` | `stagedBlocksInstall` uses `VersionComparator.isSame` on `VersionSide`s |

Nothing declined. Two notes on how two of them were implemented differ from the
brief's suggested mechanism — see "Deviations".

## L1 — the folded-build fallback

`if !VersionComparator.isNewer(rs, than: "\(isv).\(build)") { return .upToDate }`
fabricates a version component out of a build number. A build number is a large
integer, so any installed marketing version with fewer components than the build
has digits outranks every real release the vendor can ship on that line.

Measured on this Mac 2026-09-13, from Telegram's own `Info.plist`:
```
$ defaults read /Applications/Telegram.app/Contents/Info.plist CFBundleShortVersionString  → 12.10
$ defaults read /Applications/Telegram.app/Contents/Info.plist CFBundleVersion             → 282987
```
so a vendor `12.10.1` compared against the fabricated `12.10.282987` and the row
read `.upToDate`. Every source that reports no separate build (Homebrew, GitHub,
the App Store, Electron, most vendor probes) reaches this line.

The fallback now fires only on `compare(rs, combined) == .orderedSame`. Known
consequence, written into the comment: a folded remote naming an OLDER build than
what is on disk now reads as an update instead of being suppressed — the two
shapes are not separable from the strings, and offering the vendor's currently
published package to a copy ahead of it is visible and recoverable, where the old
direction hid real updates silently.

Not fixed, and still correctly described where it is documented: the Anki gap in
`GitHubReleasesSource.swift:2797` (short `26.8` + build `1` folds to exactly
`26.8.1`, so equality still matches it). Tightening that needs a rule about how
build-like a tail has to be, which I have no measurement for.

## L15 / `record` — deviation from the brief's mechanism

The brief asked for `record` to be build-aware "the same way the observe path is"
(`side.text(withBuild: true)`). `record`'s two call sites cannot both supply a
build: `ReleaseHistoryEntry` has no build field (`Models/UpdateResult.swift:108`),
and giving it one means changing the model, both producers, and — because the
stored `version` string IS the dedupe key — re-recording whole backfilled appcast
histories under new keys beside the ones already persisted.

Deduping on version + vendor date needs none of that and is sound for this tier
specifically: `record` refuses to log anything without a date, and two releases
the vendor dated differently are two releases. Same-release re-checks still
dedupe because the latest release and its own entry in `releaseHistory` read that
date from the same item through the same `ReleaseDate.publishedFields` call
(`SparkleAppcastSource.swift:60` / `:128`, `GitHubReleasesSource.swift:1246` /
`:1290`). Surge's four `6.9.0` releases now get four events.

## L2 — deviation on where the migration lives

Extracted to `Engine/NotifiedUpdateVersions.swift` (Core, pure, executed by
tests) rather than reasoned about in `AppListModel`. Legacy migration is the
"treat a stored marketing-only value as matching" option, not a rewrite-on-read:
it is two lines, it costs at most one suppressed announcement per app, and the
next pass rewrites every actionable app's entry build-aware, so it expires by
itself. A rewrite-on-read would have to decide what build an old bare `"1.0"`
meant, which is not recoverable.

## `stagedBlocksInstall` — the finding holds, with a caveat

`staged.buildVersion ?? staged.version` against
`app.buildVersion ?? app.shortVersion` can land in two namespaces: Spotify's
staged update is built with `buildVersion: nil` by construction
(`SelfUpdaterStaging.swift:233`), so the staged MARKETING string is compared with
the installed BUILD number and can never be equal — an applied-but-unswept
staging directory blocks installs. Caveat, measured 2026-09-13: this Mac's
Spotify copy carries `CFBundleVersion` == `CFBundleShortVersionString` ==
`1.2.99.317`, so on this copy the wrong pick lands right by luck. The
discriminator is wrong whether or not today's data exposes it, so it is fixed.
`buildIsDerived:` is passed for the scanner-substituted builds (Xcode, 豆包),
matching `VersionComparator.isSame`'s own documented rule.

## Red → green

Before the fixes (`swift test --filter …`):
```
✘ evaluateFoldingDoesNotSwallowAnExtraMarketingComponent() … Expectation failed: evaluate(installed: app(short: "2.0", build: "15"), remote: remote("2.0.3")) == .updateAvailable(latest: "2.0.3")
✘ evaluateFoldingDoesNotSwallowAnExtraMarketingComponent() … Expectation failed: evaluate(installed: app(short: "12.10", build: "282987"), remote: remote("12.10.1")) == .updateAvailable(latest: "12.10.1")
✘ aStagedBuildWhoseMarketingVersionTrailsIsNotReportedAsAnUpdate() failed after 0.010 seconds with 1 issue.
✘ ignoresStagedBuildWhoseMarketingVersionTrails() failed after 0.012 seconds with 1 issue.
✘ recordsEachDatedReleaseSharingOneMarketingVersion() … tl?.events.count → 1  (expected 3)
✘ anAppliedStagedUpdateWithNoBuildNumberDoesNotBlock() failed after 0.011 seconds with 1 issue.
✘ Test run with 5 tests in 3 suites failed after 0.012 seconds with 9 issues.
```
After:
```
✔ Test run with 10 tests in 4 suites passed after 0.030 seconds.
```
(The pre-fix code IS the mutation for those five: each test was written first and
observed red against it.)

`NotifiedUpdateVersions` could not be red before it existed, so it was mutated
explicitly, one property at a time:

1. `announceKey` → `offered.marketing ?? offered.build ?? ""`
```
✘ aNewBuildUnderAFrozenMarketingVersionIsNotYetAnnounced() … Expectation failed: !wasAnnounced(VersionSide(marketing: "1.0", build: "2002"), …)
✔ (the other four passed)
```
2. `wasAnnounced` → `return stored == key` (drop the legacy acceptance)
```
✘ aMarketingOnlyBaselineEntryStillCountsAsAnnounced() … 2 issues (per-path key and legacy key)
✔ (the other four passed)
```
Both restored; full suite green.

## Verification

```
$ make test > /tmp/mt-fix-vcfb2.log 2>&1; echo exit=$?
exit=0
```
tail:
```
━ Test run with 2787 tests in 229 suites passed after 21.560 seconds with 2 known issues.
✔ Test run with 280 tests in 34 suites passed after 0.150 seconds.
✓ App tests — 34 of 34 cases executed
✓ version comparisons discriminate — 262 files, 2 ledger call(s) keyed on the build, 9 marketing-first pick(s), all display-only
✓ app audits consistent — 121 docs, all indexed, no machine inventory, no pointers into untracked evidence
✓ engine-notes pointers resolve — 533 Swift files scanned, 3 doc(s) indexed
✓ skill docs consistent — 3 files mirrored, documented parameter counts (25 / 26) match the initializers
✓ no machine-population claims in code comments — 533 files
```

No `make gallery`: nothing under `App/Sources` that draws a row was touched
(`AppListModel.notifyNewUpdates`, two comments, one `Preferences` doc). No
`duo verify`: no recipe or parser rule changed.

## What a user could notice

- Updates that were previously invisible for apps whose marketing version is
  shorter than their build number now appear.
- An app that ships several builds under one marketing name gets a banner for
  each build instead of only the first, and a Release Log entry for each dated
  release instead of only the first.
- Relaunch is no longer offered for a staged build that trails the installed
  release.
- The first run after upgrading suppresses at most one announcement per app
  (the marketing-keyed baseline migration); nothing is re-announced.

### DIFF
--- a/App/Sources/AppListModel.swift
+++ b/App/Sources/AppListModel.swift
@@ -1568,7 +1568,7 @@ final class AppListModel {
 
     /// Log every release in `checked` that arrived with a trustworthy vendor
     /// date into the release timeline, then refresh the UI snapshot. The store
-    /// dedupes by (app, version), so re-checks are cheap; only a genuinely new
+    /// dedupes by (app, version, vendor date), so re-checks are cheap; only a genuinely new
     /// release or changed display metadata writes the timeline file. Results
     /// with neither a `publishedAt` nor a `vendorDay` (vendor probes, MAS,
     /// Homebrew) use the observation path below rather than this path.
@@ -1590,7 +1590,9 @@ final class AppListModel {
             }
             // …plus any prior releases the source surfaced (Sparkle appcast items,
             // a GitHub releases list), so an app's history backfills in one shot.
-            // The store dedupes by version, so the latest overlapping here is free.
+            // The store dedupes on version + vendor date, and the latest release's
+            // date comes from the same feed item as its entry here, so the overlap
+            // is free.
             for entry in remote.releaseHistory {
                 await releaseTimelineStore.record(
                     appID: result.app.id,
@@ -6576,8 +6578,13 @@ final class AppListModel {
         let actionable = results.filter(isActionableUpdate)
         // The version we'd announce for each app; key by `key(for:)` to match how
         // ignore/skip identify an app (survives the app moving on disk).
+        //
+        // Build-aware: `displayVersion` is marketing-first and a vendor may leave
+        // that string alone across any number of builds, which made every build
+        // after the first read as already-announced. See `NotifiedUpdateVersions`,
+        // where the rule and its baseline migration are executed.
         func version(_ r: UpdateResult) -> String {
-            r.remote?.displayVersion ?? r.remote?.shortVersion ?? ""
+            NotifiedUpdateVersions.announceKey(r.remote?.versionSide)
         }
         var baseline = prefs.notifiedVersions
 
@@ -6593,10 +6600,14 @@ final class AppListModel {
 
         // Consult the legacy key too: a baseline recorded before the switch to
         // per-path keys lives under the old bundle-id key, and we don't want that
-        // migration to re-announce everything once as "new".
+        // migration to re-announce everything once as "new". The same is true of a
+        // baseline recorded before the key became build-aware, which
+        // `wasAnnounced` handles by also accepting a stored marketing-only value.
         func wasNotified(_ r: UpdateResult) -> Bool {
-            baseline[prefs.key(for: r.app)] == version(r)
-                || baseline[prefs.legacyKey(for: r.app)] == version(r)
+            NotifiedUpdateVersions.wasAnnounced(
+                r.remote?.versionSide,
+                under: [prefs.key(for: r.app), prefs.legacyKey(for: r.app)],
+                in: baseline)
         }
         let newly = actionable.filter { !wasNotified($0) }
 
--- a/App/Sources/Preferences.swift
+++ b/App/Sources/Preferences.swift
@@ -327,7 +327,9 @@ final class Preferences {
     }
 
     /// Per-app version we've already posted a "new update available" notification
-    /// for (key → the offered version). Persisted so the banner fires regardless
+    /// for (key → the offered version, as `NotifiedUpdateVersions.announceKey`
+    /// spells it: build-aware, so a vendor that freezes its marketing string can
+    /// still announce its next build). Persisted so the banner fires regardless
     /// of *which* refresh path first surfaces the update — manual menu-open or
     /// scheduled background — instead of the in-memory list silently becoming the
     /// baseline; and so the same version isn't re-announced across relaunches.
--- a/App/Sources/ReleaseLogView.swift
+++ b/App/Sources/ReleaseLogView.swift
@@ -137,7 +137,23 @@ struct ReleaseLogView: View {
         let appID: String
         let appName: String
         let event: ReleaseEvent
-        var id: String { appID + "|" + event.version }
+
+        /// The version is NOT unique within an app, so it cannot be the id on its
+        /// own: `ReleaseTimelineStore.record` dedupes on the version PLUS the
+        /// vendor date, because several releases can ship under one marketing
+        /// version (Surge has shipped four called "6.9.0"). Two rows with the same
+        /// `Identifiable` id is SwiftUI's undefined-behaviour case, and this id
+        /// also decides which row hides its separator (`feed`).
+        ///
+        /// So the id is exactly the store's dedupe key with the app in front of
+        /// it, which makes it unique by construction rather than by luck: a dated
+        /// release cannot repeat that triple, and an estimated release (both date
+        /// fields nil) is still deduped on the version alone.
+        var id: String {
+            let published = event.publishedAt?.timeIntervalSince1970 ?? -1
+            let day = event.vendorDay?.timeIntervalSince1970 ?? -1
+            return "\(appID)|\(event.version)|\(published)|\(day)"
+        }
     }
 
     private struct DayGroup {
--- a/DuoUpdaterCore/Sources/DuoUpdaterCore/Engine/NotifiedUpdateVersions.swift
+++ b/DuoUpdaterCore/Sources/DuoUpdaterCore/Engine/NotifiedUpdateVersions.swift
@@ -38,10 +38,12 @@ public enum NotifiedUpdateVersions {
     /// A stored value equal to the offer's MARKETING half alone also counts as
     /// announced. That is the migration from the marketing-keyed baseline: without
     /// it, the first run after an upgrade re-announces every pending update at
-    /// once. It costs at most one suppressed announcement per app — the very next
-    /// pass rewrites that entry build-aware, after which only a genuinely
-    /// unannounced build can match — and it is deliberately not time-limited,
-    /// because an app that stops being actionable keeps its old entry indefinitely.
+    /// once. It costs at most one PASS per app, not one announcement: the pass
+    /// that sees the stale entry rewrites it build-aware, so only builds offered
+    /// during that single pass are swallowed — two builds shipped between two
+    /// passes are both swallowed, and after it only a genuinely unannounced build
+    /// can match. Deliberately not time-limited, because an app that stops being
+    /// actionable keeps its old entry indefinitely.
     public static func wasAnnounced(
         _ offered: VersionSide?, under keys: [String], in baseline: [String: String]
     ) -> Bool {
--- a/DuoUpdaterCore/Sources/DuoUpdaterCore/Engine/NotifiedUpdateVersions.swift
+++ b/DuoUpdaterCore/Sources/DuoUpdaterCore/Engine/NotifiedUpdateVersions.swift
@@ -1,0 +1,55 @@
+import Foundation
+
+/// The "have we already told the user about this update?" decision behind the
+/// **new updates available** banner, keyed on BOTH halves of the offered version.
+///
+/// Pulled out of `AppListModel.notifyNewUpdates` so it can be executed: the model
+/// itself is not constructible in a test (its `init` registers notification
+/// permissions, starts timers and installs FS watchers), and this is the part
+/// that can be wrong.
+///
+/// **Why not the marketing string.** The baseline used to store
+/// `remote.displayVersion`, which is `shortVersion ?? version` — a marketing
+/// string a vendor is free to leave alone across any number of builds. Amp
+/// shipped ten builds called "1.0" on 2026-08-28: the first was announced, the
+/// baseline recorded "1.0", and the remaining nine were `isActionableUpdate` —
+/// row lit, badge lit — with `wasNotified` permanently true. Quiet becoming
+/// silence, the same defect `StagedNudgeLedger` documents itself as avoiding, on
+/// a different ledger. `scripts/check_staged_version_use.py` does not see this
+/// shape: it hunts `shortVersion ?? buildVersion` spellings, and this one went
+/// through a computed property.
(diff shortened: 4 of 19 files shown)


jizhi0v0:
## Follow-up: the Anki gap is closed here (5e9f4acf)

The PR body above said the Anki gap survived the L1 fix. It no longer does.

**Rule.** The folded-build fallback now also requires the installed build to
parse as an integer of **at least 100**:

```swift
if let build = installed.buildVersion, !build.contains("."), !isv.hasSuffix(build),
   let counter = Int(build), counter >= 100 {
```

**Rationale, written into the comment as a judgment call rather than a measured
universal.** A vendor folds its build into the version string *because* the build
is a large monotonic counter (Oray's 30757). A one- or two-digit
`CFBundleVersion` is a constant somebody typed — Anki stamps a literal `1` on
every build — so it is evidence *against* the folded reading. Three digits clears
that class while leaving real counters alone; a vendor who genuinely folds a 99
would be offered a re-install of the version it is already on, which is the
harmless direction.

**Red → green** (`evaluateFoldingIgnoresBuildsTooSmallToBeFoldedCounters`, before
the floor):
```
✘ … Expectation failed: evaluate(installed: app(short: "26.8", build: "1"), remote: remote("26.8.1")) == .updateAvailable(latest: "26.8.1")
✘ … Expectation failed: evaluate(installed: app(short: "1.5", build: "6"), remote: remote("1.5.6")) == .updateAvailable(latest: "1.5.6")
✘ … Expectation failed: evaluate(installed: app(short: "1.5", build: "99"), remote: remote("1.5.99")) == .updateAvailable(latest: "1.5.99")
✘ Test run with 1 test in 0 suites failed after 0.001 seconds with 3 issues.
```
After: `✔ Test run with 4 tests in 0 suites passed` (the new test plus the three
existing folding tests). Boundary is pinned on both sides — 99 → updateAvailable,
100 → upToDate — and Oray `16.5.0` + `30757` vs `16.5.0.30757` still settles to
upToDate.

**Mutation** — floor dropped (`counter >= 0`):
```
✘ evaluateFoldingIgnoresBuildsTooSmallToBeFoldedCounters() … 3 issues (Anki, 1.5+6, the 99 boundary)
✔ the other three folding tests still passed
```
Restored.

**A gate fired, as designed.** `GitHubReleaseRuleTests.ankiZeroPaddedTagComparesEqualToInstalled`
had its assertion wrapped in `withKnownIssue` *precisely* so that tightening the
fallback would report `Known issue was not recorded`. It did — that was the only
failure in the first `make test` run after the floor (exit=2). The wrapper is
removed and the assertion made straight.

**Prose copies.** `grep`ed the claim before editing: three copies, all updated in
the same commit — the batch header at `GitHubReleasesSource.swift:2022`, the Anki
rule's `KNOWN GAP` at `:2797`, and the noTunes rule at `:2613` (same latent
shape, also closed, since its build is `1` too). The residual exposure is stated
rather than glossed: a *constant* build of 100 or more under a two-component
marketing version would still fold.

**Verification**
```
$ make test > /tmp/mt-fix-vcfb4.log 2>&1; echo exit=$?
exit=0
━ Test run with 2788 tests in 229 suites passed after 19.931 seconds with 1 known issue.
✔ Test run with 280 tests in 34 suites passed after 0.124 seconds.
✓ App tests — 34 of 34 cases executed
✓ version comparisons discriminate — 262 files, 2 ledger call(s) keyed on the build, 9 marketing-first pick(s), all display-only
✓ app audits consistent — 121 docs, all indexed, no machine inventory, no pointers into untracked evidence
✓ engine-notes pointers resolve — 533 Swift files scanned, 3 doc(s) indexed
✓ skill docs consistent — 3 files mirrored, documented parameter counts (25 / 26) match the initializers
✓ no machine-population claims in code comments — 533 files
```
(The remaining known issue is the pre-existing `EventStoreSchemaDriftTests` one,
present before this branch.)

jizhi0v0:
## Follow-up 2 (b04ede7a): row ids, the fold floor's arithmetic, two doc corrections

### Release Log row ids

`ReleaseLogView.Row.id` was `appID + "|" + event.version`, used by the `ForEach`
at `:91` and `:405` — and also by `:94`'s `row.id == group.events.last?.id`
separator test. The timeline dedupe change made that non-unique (Surge's four
`6.9.0`), which is SwiftUI's undefined-behaviour case.

The id is now the store's dedupe key with the app in front of it:

```swift
var id: String {
    let published = event.publishedAt?.timeIntervalSince1970 ?? -1
    let day = event.vendorDay?.timeIntervalSince1970 ?? -1
    return "\(appID)|\(event.version)|\(published)|\(day)"
}
```

Unique **by construction**, not by luck: a dated release cannot repeat
`(version, publishedAt, vendorDay)`, and an estimated release (both date fields
nil) is still deduped on the version alone. `ReleaseTimelineStore`'s own test now
asserts the triple does not repeat within an app, so the invariant the view
depends on is pinned in Core rather than assumed.

### Other version-keyed readers — audited

- **`ReleaseStats` / the heatmap** — buckets each event by its own `publishedAt`
  hour and weekday; no version keying at all. Strictly more accurate now: three
  of Surge's four `6.9.0` releases previously contributed nothing to the
  histogram.
- **`AppReleaseTimeline.latest`** — `max` by `timestamp`. Unaffected.
- **`ReleaseLogView` `:348`** — keyed on `app.id`. Unaffected.
- **`observeForChange`'s `contains(where: { $0.version == version })` (`:206`)** —
  **still correct, left as is**, reasoning written at the guard. Two arguments,
  both directions:
  - *Within its tier* the version already IS the full discriminator: the caller
    passes `side.text(withBuild: true)`, so a frozen-marketing vendor's next
    build arrives as `"6.9.0 (2382)"` and is not suppressed. Adding a date here
    would key on **our polling clock**, which changes every check — it would
    defeat the guard entirely rather than sharpen it.
  - *Across tiers* suppressing is the right answer: this path only runs when the
    source stated no date at all, so an estimated window for a version some
    source already dated is strictly worse information.
  - Known asymmetry, pre-existing and not worsened here: the two tiers store
    different strings (`"6.9.0 (2381)"` vs `"6.9.0"`), so an app that later gains
    a dated source can hold one of each. That dates from the observe path
    becoming build-aware, not from this PR.

### The fold floor is now measured by string

`Int(build)` returns nil for builds wider than `Int64` — epoch-ms, concatenated
counters — which is the *most* counter-like shape there is, so the floor would
have silently switched fold recognition **off** for exactly those vendors. This
is the case `VersionComparator`'s tokenizer already keeps digit runs as strings
for (`Token.number`'s doc, `:220-222`).

```swift
static func buildIsFoldableCounter(_ build: String) -> Bool {
    guard !build.isEmpty, build.allSatisfy({ $0.isASCII && $0.isNumber }) else { return false }
    return build.drop(while: { $0 == "0" }).count >= 3
}
```

Leading zeros stripped first, matching `Token.compareNumeric`'s own
normalisation, so `"007"` is the counter 7 — tested.

**Mutation** — floor swapped back to `Int(build).map { $0 >= 100 } ?? false`:

```
✘ evaluateFoldingIgnoresBuildsTooSmallToBeFoldedCounters() … VersionComparatorTests.swift:208: evaluate(installed: app(short: "1.5", build: wide), …)
✘ Test run with 1 test in 0 suites failed after 0.001 seconds with 1 issue.
```

Exactly the 25-digit case, nothing else. Restored. The 99/100 boundary and the
Anki case are unchanged and still pinned.

### Accepted-risk case pinned

Added to `evaluateFoldingDoesNotSwallowAnExtraMarketingComponent`: installed
`16.5.0` + `30757` against a fold-shaped but OLDER remote `16.5.0.29000` →
`updateAvailable`. A later move back toward `!isNewer(rs, combined)` now fails a
test instead of passing quietly.

### Doc correction

`NotifiedUpdateVersions.wasAnnounced` said the legacy match "costs at most one
suppressed announcement per app". That holds per *pass*, not per *build* — two
builds offered between two passes are both swallowed. Reworded to say one PASS,
and to spell out that case. (The PR body above carries the same wrong phrasing;
this comment is the correction.)

### Verification

```
$ make test > /tmp/mt-fix-vcfb6.log 2>&1; echo exit=$?
exit=0
━ Test run with 2788 tests in 229 suites passed after 20.260 seconds with 1 known issue.
✔ Test run with 280 tests in 34 suites passed after 0.138 seconds.
✓ App tests — 34 of 34 cases executed
✓ version comparisons discriminate — 262 files, 2 ledger call(s) keyed on the build, 9 marketing-first pick(s), all display-only
✓ app audits consistent — 121 docs, all indexed, no machine inventory, no pointers into untracked evidence
✓ engine-notes pointers resolve — 533 Swift files scanned, 3 doc(s) indexed
✓ skill docs consistent — 3 files mirrored, documented parameter counts (25 / 26) match the initializers
✓ no machine-population claims in code comments — 533 files
```

No `make gallery`: `ReleaseLogView` is not one of the row views the gallery
renders (`PopoverRowAction` / `WorkbenchRowAction` / `RowActionViews` / `AppRow`),
and none of those were touched.

PR #569, opened 2026-09-12
Move every blocking install/CLI call off the cooperative pool, and gate it

jizhi0v0:
Fixes the whole **P1** cluster of the 2026-09-13 review (report section 二): every blocking call reached from `async` code that was parking a cooperative thread, plus the two false comments the cluster rests on.

`OffPool.swift`'s doc is the premise: the cooperative pool is width-capped near the core count and does **not** overcommit when a thread parks, so a handful of blocking calls on it stops the runtime scheduling anything at all (#351, run 33959023008: three parked threads, 0.00s and 0.01s of CPU across sixty seconds, silence for the rest of the job). `Task.detached` is not an alternative — it still runs on that pool. Before this PR, `grep offCooperativePool App/Sources CLI/Sources` answered **0**.

## Findings addressed

| # | Site | What it was | What it is now |
|---|---|---|---|
| 1 | `Install/InPlaceSwap.swift` `replace`, via `SparkleInstaller.swift:262`, `VendorInstaller.swift:267`, `BackupStore.restore:413` | `NSLock` + osascript admin panel held for the whole time the user takes to answer | both installers' `installApp` are `async` and hop; the rollback paths (App + CLI) hop their `BackupStore.restore` |
| 2 | `Install/DeltaApplier.swift` `reconstruct` | `BinaryDelta`'s `waitUntilExit()`, no watchdog | hopped at both call sites; SIGTERM at 600s / SIGKILL at 605s, the `ArchiveExtractor.run` shape |
| 3 | `Install/InstallCoordinator.swift` `backUp` | `Task.detached` around `BackupStore.save` + `BackupManifest` + `InputMethodDataBackup` (8.7s × 2 for Word) | one `offCooperativePool` around the same block, unchanged inside |
| 4 | `InPlaceSwap.recoverInterruptedSwaps` via `AppListModel:2269` | `Task.detached` around `SecStaticCodeCheckValidity` — the #351 call | hop at the call site only (the function body is untouched, so it cannot conflict with `fix/backup-seal-and-sweep-lock`) |
| 5 | `Install/PackageInstaller.swift` | gate (`pkgutil`/`xar`/`lsbom`/`SecStaticCode`) and two full-file SHA-256 seals inline on the actor | **one hop per contiguous sequence**: gate+seal together, twice; the dmg fingerprint and `resolveInstaller` separately. Gate/seal ordering untouched |
| 6 | `Sources/VendorProbeSource.swift` `zipEntryPlistValue` | `unzip` read+wait on every check and every `duo verify` | hopped |
| 7 | `AppListModel.runningBuildVersions`, `PrivilegedHelperClient.restartDaemon` | `Task.detached` | `offCooperativePool`; `offCooperativePool` is now `public` |
| 8 | `Sources/GitHubToken.swift` via `ChangelogService:379` and `AppListModel:1394` | `Task.detached`; the ChangelogService comment already *claimed* "off the cooperative pool" | both hop inside the task that races the 2s deadline, so the deadline behaviour is unchanged |
| 9 | `Install/BrewFormulaReleaseService.swift:563` | `standardError = Pipe()`, never drained → deadlock past 64KB | `FileHandle.nullDevice`, with `BrewFormulaService.realExecutor`'s reasoning |
| 10 | CLI `Doctor.swift`, `Backups.swift`, `GitHub.swift`, `Triage.swift` | blocking on the pool (single-task, so not starving, but the rule is the rule) | `doctor` and `backups restore` hop directly; `triage` and `reconcile` are synchronous end to end, so each gets one hop at its `main.swift` seam |

Nothing was declined. Two sites the brief did not list turned up when the gate was run and are fixed here too: `AppStoreAXInstaller.ensureAppStoreRunning` (`open -g` + `waitUntilExit()` directly in an `async func`) and `GitHubToken.run`/`runCapturingStderr`, which both handed the half they do not read an undrained `Pipe()`.

### Comment corrections

* `InPlaceSwap.swift:645` said *"the callers are synchronous"* — both are actor-isolated `async` and always were. Replaced with the actual invariant (every async caller must reach `replace` through a hop) and what the old sentence got wrong.
* `PackageInstaller.swift:178` said a direct pkg avoids *"an extra full read"* — `handOver` hashes the file whole twice either way, deliberately. Rewritten to say what the `approvedFingerprint` hand-off really buys.
* `GitHubToken.ghCLIToken`'s *"cheap subprocess"* now says it is cheap only when it answers, and must be called off-pool.

Both were grepped for copies first (`grep -rn "callers are synchronous"` → 1; the pkg sentence → 1).

### Out of scope, reported not fixed

* **The download permit still spans the whole pkg gate + two hashes** (`InstallCoordinator` vs `InstallPermits.swift:5-6`'s "held only while fetching bytes"). Moving the boundary is a design change; untouched.
* `MASInstaller.runOutdated` is **exempted, not hopped**: its wait is the `terminationHandler` continuation, and by the time the drain runs the child has exited and Foundation has closed the parent's write end. The exemption carries that reason.
* `CLAUDE.md`'s 「别在协作池上做阻塞调用」 now has one stale line — *"`recoverInterruptedSwapsOnce` 还从 `Task.detached` 再加"* — which this PR makes false, and the section has no mention of the new gate. Left for the lead: I do not edit CLAUDE.md on an agent brief.

## Changes a user could notice

None intended. Same work, same order, same timeouts — on a Dispatch thread instead of a cooperative one. Three things worth a reviewer's eye:

1. `offCooperativePool` gained a **non-throwing overload** so callers that cannot throw need not write `(try? …) ?? fallback` around work that never throws (an unreachable branch that reads as a handled failure). The two existing `try?` call sites drop theirs; overload resolution picks the throwing one whenever the closure body throws (verified: no `UnnecessaryEffectMarker` warnings in the build).
2. `PackageInstaller`'s blocking helpers became `nonisolated`. The actor was not ordering them — `handOver` already suspends between its two gates and `downloadAndOpen` awaits its download — but two installs of *different* packages can now overlap their gates in wall time where before they interleaved around suspension points.
3. `DeltaApplier` sends `BinaryDelta`'s stdout to `nullDevice`. It was read and discarded, and reading it *after* stderr was itself a deadlock.

## The gate, and its evidence

`scripts/check_offpool.py` (in `make test`) walks out from each `waitUntilExit()` / `readDataToEndOfFile()` / `SecStaticCodeCheckValidity` to the first scope that decides whose thread it is: an `offCooperativePool` closure passes, an `async func` or a `Task` fails, a plain synchronous function is **not** reported. That limit is written into the file: `InPlaceSwap.replace` is synchronous and was reached from two `async` callers for months, which nothing in its own body can reveal — answering that needs a call graph.

`offpool-lint:allow — <reason>` exempts a function; the reason is mandatory and a **dead exemption fails the build** (`mayLookAlike`'s shape, not `mayBeBlank`'s #271 hole).

Red on the pre-fix tree, green on this branch — same script, `git archive origin/main` into a scratch dir:

```
$ python3 …/prefixcheck.py …/prefix        # origin/main @ 924be257
scanned 261 calls 43 offences 8
   …/Install/AppStoreAXInstaller.swift 536 waitUntilExit()
   …/Install/MASInstaller.swift 370 readDataToEndOfFile()
   …/Sources/VendorProbeSource.swift 1401 readDataToEndOfFile()
   …/Sources/VendorProbeSource.swift 1402 waitUntilExit()
   App/Sources/AppListModel.swift 4774 readDataToEndOfFile()
   App/Sources/AppListModel.swift 4775 waitUntilExit()
   App/Sources/PrivilegedHelperClient.swift 183 readDataToEndOfFile()
   App/Sources/PrivilegedHelperClient.swift 184 waitUntilExit()

$ python3 scripts/check_offpool.py          # this branch
✓ no blocking calls left on the cooperative pool — 42 call(s) in 261 files
exit=0
```

Mutation-tested by hand against the real tree (each applied, run, reverted):

```
--- restored: exit=0                         ✓ … 42 call(s) in 261 files
--- mutation 1 (bare waitUntilExit in an async func): exit=1
✗ …/SignatureVerifier.swift:83: `waitUntilExit()` parks a cooperative thread …
--- mutation 2 (the same call inside Task.detached): exit=1
✗ …/SignatureVerifier.swift:84: `waitUntilExit()` parks a cooperative thread …
--- mutation 3 (marker, no reason): exit=1   (names the line, and the stale marker)
--- mutation 4 (marker + reason): exit=0
--- mutation 5 (stale exemption): exit=1
✗ …/SignatureVerifier.swift:80: `offpool-lint:allow` here no longer exempts anything …
```

`scripts/test_check_offpool.py` (also in `make test`) pins eleven of them, including the two that made earlier drafts of this script silently green:

* judging the `{`'s own line rather than the accumulated declaration — every wrapped `async` signature then reads as synchronous, which is how the first draft missed `VendorProbeSource` entirely (it is in the pre-fix output above only because that was fixed);
* counting braces per line instead of walking them in order — `func f() -> Int { 5 }` and `} else {` break a count in opposite directions.

Plus both emptiness floors (roots gone; call spellings gone stale), because a check that inspects nothing prints the same ✓ as one that inspects everything.

## Verification

```
$ make test > /tmp/duo-mt-offpool2.log 2>&1; echo exit=$?
exit=0

$ grep -n "Test run with .* tests" /tmp/duo-mt-offpool2.log
6665:━ Test run with 2777 tests in 228 suites passed after 20.444 seconds with 2 known issues.
7346:✔ Test run with 280 tests in 34 suites passed after 0.144 seconds.

$ tail -8 /tmp/duo-mt-offpool2.log
python3 scripts/check_skill_docs.py
✓ skill docs consistent — 3 files mirrored, documented parameter counts (25 / 26) match the initializers
python3 scripts/check_prose_claims.py
✓ no machine-population claims in code comments — 531 files
python3 scripts/check_offpool.py
✓ no blocking calls left on the cooperative pool — 42 call(s) in 261 files
```

`make gallery` not run: the `App/Sources` changes are all model-side (`runningBuildVersions`, `rollback`, the token resolve, the swap sweep, the helper kickstart). No row view, no `RowAction.state`, and none of those files is in the gallery target. `duo verify` not run: no recipe or parser rule changed.

### DIFF
--- a/App/Sources/AppListModel.swift
+++ b/App/Sources/AppListModel.swift
@@ -1399,8 +1399,15 @@ final class AppListModel {
         // none per menu-bar open — everything else goes through
         // `githubTokenForRecheck`, which only reaches here on a miss.
         Log.app.info("GitHub token: resolving (explicit=\(explicit != nil, privacy: .public))")
+        // The resolve shells out to `gh auth token` and blocks in
+        // `waitUntilExit()`, so it goes to Dispatch rather than to the cooperative
+        // pool a detached task shares (see `offCooperativePool`). The task wrapper
+        // stays: it is what `firstResult` races the deadline against, and all it
+        // does itself is await.
         let loader = Task.detached(priority: .utility) {
-            GitHubToken.resolve(explicit: explicit)
+            await offCooperativePool(qos: .utility) {
+                GitHubToken.resolve(explicit: explicit)
+            }
         }
         if let token = await firstResult(of: loader, within: timeout) {
             return token
@@ -2283,6 +2290,10 @@ final class AppListModel {
             URL(fileURLWithPath: "/Library/Input Methods", isDirectory: true),
             home.appendingPathComponent("Library/Input Methods", isDirectory: true),
         ]
+        // The sweep's own blocking half (`SecStaticCode…` on every orphan it
+        // finds — the exact call #351 measured) hops inside
+        // `InPlaceSwap.recoverInterruptedSwaps`, which is async now that it takes
+        // the install lock. Nothing to do here but await it.
         Task.detached(priority: .utility) { [weak self] in
             var swept = true
             for root in roots {
@@ -4831,7 +4842,11 @@ final class AppListModel {
     /// relaunch a batch of apps during an install. Blocking `@MainActor` here
     /// froze the whole UI (spin report) and stranded the half-restarted app.
     nonisolated private static func runningBuildVersions() async -> [String: String] {
-        await Task.detached(priority: .utility) {
+        // Dispatch, not `Task.detached`: a detached task still runs on the
+        // cooperative pool, and the pair below (`readDataToEndOfFile()` +
+        // `waitUntilExit()`) parks whatever thread it lands on for as long as
+        // `coreservicesd` takes. See `offCooperativePool`.
+        await offCooperativePool(qos: .utility) {
             let process = Process()
             process.executableURL = URL(fileURLWithPath: "/usr/bin/lsappinfo")
             process.arguments = ["list"]
@@ -4867,7 +4882,7 @@ final class AppListModel {
             // instead of a live process list. This closure only owns invoking the
             // tool and its timeout/kill backstop.
             return LSAppInfoParser.runningBuildVersions(from: text)
-        }.value
+        }
     }
 
     /// Take down a note one of the restart paths wrote — and only if it is still
@@ -5784,9 +5799,13 @@ final class AppListModel {
             return
         }
         do {
-            let restored = try await Task.detached(priority: .userInitiated) { () -> String? in
+            // Off the cooperative pool, not merely off this actor: the restore
+            // dittos the stored bundle out and then runs the same blocking
+            // `InPlaceSwap.replace` an install does, and a detached task still runs
+            // on the pool. See `offCooperativePool`.
+            let restored = try await offCooperativePool(qos: .userInitiated) { () -> String? in
                 try BackupStore.restore(forKey: key, over: target)
-            }.value
+            }
             // The swap has landed; everything below is bookkeeping and needs no
             // exclusion, so hand the claim back rather than holding it through a
             // rescan (same reasoning as the apply permit in `performInstall`).
--- a/App/Sources/AppListModel.swift
+++ b/App/Sources/AppListModel.swift
@@ -489,7 +489,8 @@ final class AppListModel {
         Task {
             let page = await Self.firstResult(
                 of: Task.detached(priority: .userInitiated) {
-                    TestFlightInventory().frontier(forBundleID: bundleID)?.appPageURL
+                    await TestFlightInventory.loadOffPool()
+                        .frontier(forBundleID: bundleID)?.appPageURL
                 },
                 within: .seconds(2))
             if let url = page ?? nil {
@@ -1399,7 +1400,13 @@ final class AppListModel {
         // none per menu-bar open — everything else goes through
         // `githubTokenForRecheck`, which only reaches here on a miss.
         Log.app.info("GitHub token: resolving (explicit=\(explicit != nil, privacy: .public))")
-        // The resolve shells out to `gh auth token` and blocks in
+        // A settings value or an env var answers here, with no subprocess, no hop
+        // and no deadline. Racing those was a regression: the hop has to be
+        // ADMITTED to a Dispatch queue, and on a loaded machine admission alone
+        // can outlast the two seconds — which turns a token the user pasted into
+        // "continuing without a token". See `GitHubToken.preresolved`.
+        if let cheap = GitHubToken.preresolved(explicit: explicit) { return cheap }
+        // Only `gh auth token` is left. It shells out and blocks in
         // `waitUntilExit()`, so it goes to Dispatch rather than to the cooperative
         // pool a detached task shares (see `offCooperativePool`). The task wrapper
         // stays: it is what `firstResult` races the deadline against, and all it
@@ -2036,7 +2043,7 @@ final class AppListModel {
     private static func beginTestFlightLoad(
         timeout: Duration = .seconds(2)
     ) async -> (inventory: TestFlightInventory, pendingLoader: Task<TestFlightInventory, Never>?) {
-        let loader = Task.detached(priority: .utility) { TestFlightInventory() }
+        let loader = Task.detached(priority: .utility) { await TestFlightInventory.loadOffPool(qos: .utility) }
         if let loaded = await firstResult(of: loader, within: timeout) {
             return (loaded, nil)
         }
@@ -2113,7 +2120,7 @@ final class AppListModel {
         let unread = TestFlightInventory(macRows: [], accessible: false)
         let inventory = mayRead
             ? await Self.firstResult(
-                of: Task.detached(priority: .userInitiated) { TestFlightInventory() },
+                of: Task.detached(priority: .userInitiated) { await TestFlightInventory.loadOffPool() },
                 within: .seconds(2)) ?? unread
             : unread
         // Granted, but the store did not open in time or at all: leave the rows as
@@ -2132,7 +2139,7 @@ final class AppListModel {
         guard !targets.isEmpty else { return }
         let announcements: TestFlightAnnouncements? = mayRead
             ? await Self.firstResult(
-                of: Task.detached(priority: .userInitiated) { TestFlightAnnouncements() },
+                of: Task.detached(priority: .userInitiated) { await TestFlightAnnouncements.loadOffPool() },
                 within: .seconds(2))
             : nil
         let signedIn: Bool? = mayRead ? await AppStoreSignIn.current() : nil
@@ -2388,7 +2395,7 @@ final class AppListModel {
         // them again, and they say they cannot tell (`ScanRowAssembly.roundPlan`,
         // below).
         let tfLoader: Task<TestFlightInventory, Never>? =
-            allowTestFlight ? Task.detached(priority: .utility) { TestFlightInventory() } : nil
+            allowTestFlight ? Task.detached(priority: .utility) { await TestFlightInventory.loadOffPool(qos: .utility) } : nil
         if allowTestFlight { testFlightReadThisSession = true }
 
         // The refresh button is the one intent that asks TestFlight to sync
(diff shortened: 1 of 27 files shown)


jizhi0v0:
Rebased onto `main` (through #571) and pushed `effcad47`. Four things landed since the first push.

## A. The CI failure — root cause

`ChangelogGitHubAuthTests.explicitTokenFromSettingsReachesTheChangelogFetcher`, red on the 3-core runner after 18.9 s, green here in a millisecond.

Not the non-throwing overload, and not `gh` being absent or slow — the test's token never needed `gh` at all. `GitHubToken.resolve` answers an explicit settings value (step 1) and the env vars (step 2) with two memory reads. My hop wrapped **all three** steps, so even those answers had to be *admitted to a Dispatch queue* before they could be returned, inside a 2-second deadline race that was previously won trivially. On a loaded 3-core runner queue admission is not instant — CLAUDE.md's own timing note says pool admission there is measured in seconds — so `firstResult` timed out, `gitHubToken()` returned nil, and a token sitting in Settings resolved to "anonymous".

The regression was mine and it was in the *shape* of the fix, not its correctness: a hop is the right thing for work that can block and the wrong thing for work that cannot.

**Fix**: `GitHubToken.preresolved(explicit:)` is steps 1 and 2 alone. `ChangelogService.gitHubToken` and `AppListModel.resolveGitHubToken` answer from it when they can and spend the deadline only on `gh auth token`.

**Made deterministic** rather than reproduced by simulation: a one-nanosecond deadline is the same race, lost on every machine.

```
$ swift test --filter ChangelogGitHubTokenResolutionTests     # after
✔ Test anExplicitTokenDoesNotRaceTheDeadline() passed after 0.001 seconds.

# mutation: `if false, let cheap = GitHubToken.preresolved(…)`
✘ Test anExplicitTokenDoesNotRaceTheDeadline() recorded an issue …
      Expectation failed: await ChangelogService.gitHubToken(timeout: .nanoseconds(1))
✔ Test explicitTokenFromSettingsReachesTheChangelogFetcher() passed after 0.001 seconds.
```

Note the second line of that mutation run: **the original test still passes with the bug in place**, on this machine. That is the whole reason it reached CI.

## B. Gate — seven findings, seven tests

All seven verified against the previous script: **red there, green here.**

```
$ cp <previous check_offpool.py> scripts/ && python3 scripts/test_check_offpool.py
FAIL: test_semaphore_and_group_waits_are_blocking
FAIL: test_gcd_closures_are_not_the_cooperative_pool
FAIL: test_a_stale_marker_is_not_covered_by_a_longer_live_one
FAIL: test_a_brace_inside_a_block_comment_is_not_a_scope
FAIL: test_an_async_getter_is_an_async_scope
FAIL: test_a_default_closure_argument_does_not_end_the_signature
FAIL: test_an_extended_string_literal_is_not_code
Ran 21 tests — FAILED (failures=7)

$ python3 scripts/test_check_offpool.py                        # current
Ran 21 tests — OK
```

1. **`.wait()` / `.wait(timeout:` added**, plus `.run(key:` (`BoundedBlockingWork.run` — a five-second cap, still a park). The cost of carrying `.wait()` is `await x.wait()`, which is an async wait and not a parked thread; those are dismissed on the line's own evidence, pinned by `test_an_awaited_wait_is_not_a_parked_thread` against the two real `AppListModel` lines.
2. **GCD false positive fixed** — `.async`/`.asyncAfter`/`DispatchWorkItem`/`Thread { }` are sync frames. ⚠️ **`.sync { }` deliberately is not**: it runs the block on the *calling* thread, so exempting it would hide the violation rather than find it. Both halves pinned (`test_dispatch_sync_is_still_the_calling_thread`).
3. **Dead markers by line number**, not substring. The fixture is the direction that actually bites: a live `— network` covering a stale `— network retry budget`.
4. **Block comments stripped.** Stripping is now a character walk rather than regexes in some order, because every order is wrong for something — this repo has `/*` inside a `//` comment (`AppRuntime.swift:252`), `#"…"#` regexes full of quotes, and multi-line SQL. Two tests.
5. **`get async { }`** is an async scope.
6. **A default closure argument** (`onDone: () -> Void = { }`) no longer ends a wrapped signature.
7. **Extended literals.** The discriminating fixture is `#"say " then }"#`: the old pair-matching stripper consumed `"say "` and left the `}` as code, which popped the enclosing function and made everything after it read as type-level — silently green.

**Limit now written into the docstring**, because the addendum ran straight into it: the gate sees a blocking call where it is *written*, so anything one call away from an async body — `BoundedBlockingWork.run` inside an initializer — is invisible. Adding token spellings is cheap; a call graph is not, and a gate that implies it has one is worse than this one.

## C. TestFlight (the addendum)

`BoundedBlockingWork` bounds the *wait*, it does not un-block it: up to five seconds of `DispatchSemaphore.wait` on the calling thread, and the cap is reached exactly when the app-data gate goes unanswered — when the app has the most other work in flight. `loadOffPool()` on `TestFlightInventory` and `TestFlightAnnouncements` keeps the eleven call sites one-liners and keeps the detached task the deadline races are written against.

The two `AppScanner` sites hop **as a whole** rather than just their TestFlight argument: `scan()` is synchronous to the bottom and bounded the same way.

None of this was found by the gate. It was found by reading.

## D. Rebase

One conflict, `App/Sources/AppListModel.swift`, in `recoverInterruptedSwapsOnce` — #565 made `InPlaceSwap.recoverInterruptedSwaps` async and lock-taking, exactly where my call-site hop was. **Resolved to main's version in place** (lock loop, per-root deferral, `didRecoverSwaps` reset) and the hop moved to where it now belongs: inside `recoverInterruptedSwaps`, around `sweepInterruptedSwaps`, which is where the `SecStaticCode` call now sits under an `async` signature.

That required correcting the note #565 wrote beside the release — it said the sweep "neither throws nor suspends", and the hop suspends. It now says what still holds: `offCooperativePool` is not cancellable and always resumes, so no path from the claim reaches the end without releasing.

Nothing else conflicted; `InPlaceSwap.replace` returning `SwapOutcome` (#568) merged cleanly through both installers' hops.

## Verification on `effcad47`

```
$ make test > /tmp/duo-mt-offpool4.log 2>&1; echo exit=$?
exit=0

$ grep -n "Test run with .* tests\|cases executed\|Ran 21 tests" /tmp/duo-mt-offpool4.log
70:Ran 21 tests in 0.055s
6715:━ Test run with 2805 tests in 233 suites passed after 20.577 seconds with 2 known issues.
7454:✔ Test run with 280 tests in 34 suites passed after 0.120 seconds.
7455:✓ App tests — 34 of 34 cases executed

$ python3 scripts/check_offpool.py
✓ no blocking calls left on the cooperative pool — 51 call(s) in 262 files
```

PR #591, opened 2026-09-13
helper: check XPC peers with public API instead of the auditToken SPI

jizhi0v0:
## What / why

The root helper's only gate against "any local process drives `mas` as root" read the **private** `NSXPCConnection.auditToken` via KVC, then `SecCodeCopyGuestWithAttributes` + `SecCodeCheckValidity`, and derived uid/gid from the same token.

This PR replaces that with public API, and only after measuring that the gate does not get weaker. It has two layers with the same requirement string:

1. **Listener:** `NSXPCListener.setConnectionCodeSigningRequirement(_:)` (macOS 13; deployment target 14). A non-matching peer is dropped before the delegate is consulted.
2. **Connection (review round 1):** `NSXPCConnection.setCodeSigningRequirement(_:)` on every accepted connection, before `resume()`. It is checked per message. It exists because layer 1 is only *observed* on macOS 27.

The uid and gid come from `effectiveUserIdentifier` / `effectiveGroupIdentifier`, read in the delegate.

`App/Helper/HelperPeerGate.swift` is the only way to start the listener:

- `install(on:team:)` → `requirement(team:)` → install the requirement, attach the delegate, resume.
- It rejects everything when the team is missing or malformed, or when the gate is attached to a listener it did not install on.
- It is Foundation-only and compiled into `DuoUpdaterAppTests`, together with `HelperService.swift`, `Shared/OwnTeamIdentifier.swift` and `Shared/MASHelperProtocol.swift`.

Other changes:

- A team is spliced into a requirement only if it is exactly ten `A-Z0-9` characters. Before this, a malformed team would have crashed the helper at launch through an ObjC exception.
- `installMASApp`'s uid/gid/userName check is now `HelperClientIdentity.matchesClaim`. The `mas` execution is untouched.
- `-lbsm` is dropped.
- `scripts/xpc-peer-probe/` holds the two-process experiment behind the claims below. It is not in `make test`, because it bootstraps launchd user agents.
- Round 2:
  - The "no requirement" log now tells *no team* apart from *malformed team* (`OwnTeamIdentifier.missingRequirementReason`), in both the helper and `PrivilegedHelperClient`.
  - Doc wording corrected: the listener-identity invariant, and `OwnTeamIdentifier`'s trust claim about the team.
  - The `installMASApp` test has a tripwire against a future `TEST_HOST`.
  - `scripts/xpc-peer-probe/InProcessDoubleRequirement.swift` was added, and run on macOS 14 and 15 (below).
  - **A flake this PR introduced into `make test` is fixed** (4ff2e8b3, see "app-tests coverage gate" below).

## app-tests coverage gate: a flake this PR introduced

What happens:
- `HelperPeerGateTests` make Security (`[logging-persist] … /private/var/db/DetachedSignatures`) and libxpc log, and the gate `NSLog`s its rejections.
- xcodebuild puts those lines in the same stream as swift-testing's records.
- A line landing mid-record splits a case name across two lines.
- `scripts/app_test_coverage.py` then reported a **passing** case as never run, and `scripts/app-tests.sh` failed.

Measured on this machine:
- 3 failures in 13 consecutive runs. The first split name was an unrelated `ScanRowAssembly` case.
- With `HelperPeerGateTests` disabled, the run contained **no** such log lines.
- Switching the gate from `NSLog` to `os.Logger` did not help: 2 failures in 12 runs. It was reverted.

Fix:
- The parser cuts out lines with the exact shape `YYYY-MM-DD HH:MM:SS.ffffff±ZZZZ name[pid:tid] …\n` before matching, which rejoins the record.

Checks:
- Replayed on the 3 failing logs: all cases found.
- In 12 further runs, all green. One log had a split record that the old parser would have failed (`aTestFlightRowSettlesOnceTheOfferIsInstalled`).
- A log from a run with the suite `.disabled` still reports those 14 cases missing.
- `scripts/test_app_test_coverage.py` (in `make test`) uses the real split lines, with three mutations seen red: C1 drop the removal, C2 loosen the regex, C3 drop the `passed` anchor.

CI run 34752901721 passing on 520a56bc was not evidence against the flake; it can pass by luck.

## Coverage by OS (dated 2026-09-13)

| OS | What ran | Result |
|---|---|---|
| macOS 27.0 (26A428), this machine | launchd probe (Mach service, separately signed peers, exec race); disassembly; `HelperPeerGateTests`; `InProcessDoubleRequirement.swift` | all as below; probe a/b/c PASS |
| macOS 26 (image macos-26-arm64 20260907.0351), [CI run 34752901721](https://github.com/jizhi0v0/duo-updater/actions/runs/34752901721), head 520a56bc | app tests 47/47, including every `HelperPeerGateTests` case: listener + per-connection requirement on one connection, anonymous listeners | pass |
| macOS 15.7.9 (24G830, image macos-15-arm64 20260907.0337, Xcode 16.4 / Swift 6.1.2), [Actions run 34753854027](https://github.com/jizhi0v0/duo-updater/actions/runs/34753854027), job 103714799755 | `InProcessDoubleRequirement.swift` | a/b/c PASS, exit 0 |
| macOS 14.8.9 (23J631, image macos-14-arm64 20260831.0302, Xcode 15.4 / Swift 5.10), same run, job 103714799924 | `InProcessDoubleRequirement.swift` | a/b/c PASS, exit 0 |

Raw lines from run 34753854027. The workflow ran only on this branch; commit d0a7c0d5 removed it, so it never reaches main.

```
probe (macos-14)  Xcode 15.4 / Build version 15F31d / Apple Swift version 5.10 / ProductVersion: 14.8.9 / BuildVersion: 23J631 / arm64
PROBE host: Version 14.8.9 (Build 23J631)
SUMMARY a: PASS observed=[failed(NSCocoaErrorDomain 4097) delegate=0 invoked=0] expected=[failed delegate=0 invoked=0]
SUMMARY b: PASS observed=[replied delegate=1 invoked=1] expected=[replied delegate=1 invoked=1]
SUMMARY c: PASS observed=[failed(NSCocoaErrorDomain 4097) delegate=1 invoked=0] expected=[failed delegate=1 invoked=0]
OVERALL: PASS
probe exit status: 0

probe (macos-15)  Xcode 16.4 / Build version 16F6 / Apple Swift version 6.1.2 / ProductVersion: 15.7.9 / BuildVersion: 24G830 / arm64
PROBE host: Version 15.7.9 (Build 24G830)
SUMMARY a: PASS observed=[failed(NSCocoaErrorDomain 4097) delegate=0 invoked=0] expected=[failed delegate=0 invoked=0]
SUMMARY b: PASS observed=[replied delegate=1 invoked=1] expected=[replied delegate=1 invoked=1]
SUMMARY c: PASS observed=[failed(NSCocoaErrorDomain 4097) delegate=1 invoked=0] expected=[failed delegate=1 invoked=0]
OVERALL: PASS
probe exit status: 0
```

Scenarios:
- **a:** listener never-matching.
- **b:** listener matching + per-connection matching. This is the double set that must not trap.
- **c:** listener matching + per-connection never-matching.

The probe's trap reporting works: a deliberately wrong variant that sets the requirement **twice on the same connection** printed `SUMMARY b: CRASHED signal 5` and `OVERALL: FAIL` on macOS 27 (local temp copy, not committed). So scenario b passing means the listener + connection pair is *not* treated as a double set on 14, 15 or 27.

**Limits on 14, 15 and 26:** anonymous listener, self-connection, not root, not a Mach service, no separately signed peer, no exec race.

## Evidence

Runtime observations: macOS 27.0 (26A428), Xcode 27 (27A5237l) / Swift 6.4. Probe = `PROBE_DEVID=<Developer ID sha1> scripts/xpc-peer-probe/run.sh`.

Probe setup:
- The listener is a launchd **user** agent using `NSXPCListener(machServiceName:)`.
- Peers are separate binaries, ad-hoc signed as `…zzprobe.good` / `…zzprobe.bad`.
- In the prod-shaped scenarios, one peer is Developer ID signed (RS59HDH7Y3) and one is an ad-hoc binary forging the identifier.

Full SUMMARY from the round-1 re-run (exit 0, no `zzprobe` agents left afterwards):

| Scenario | Listener req | Conn req | Peer | delegate | invoked | opened | closed |
|---|---|---|---|---|---|---|---|
| control-none | – | – | bad | 1 | 6 | 1 | 1 |
| listener-good | good | – | good | 1 | 6 | 1 | 1 |
| listener-bad | good | – | bad | **0** | **0** | 0 | 0 |
| listener-bad-burst (200 + 200 queued) | good | – | bad | **0** | **0** | 0 | 0 |
| listener-exec-race (unforced) | good | – | bad → exec good | 1 | 2 | 1 | 1 |
| listener-exec-stopped (forced, see Q1) | good | – | warmup; bad → exec good | 2 | 4 | 2 | 2 |
| control-exec-stopped | – | – | warmup; bad → exec good | 3 | 10 | 3 | 3 |
| delegate-reject (old-style rejection) | – | – | bad | 1 | 0 | 0 | 0 |
| conn-good | – | good | good | 1 | 6 | 1 | 1 |
| conn-bad | – | good | bad | 1 | **0** | 1 | 1 |
| conn-bad-burst | – | good | bad | 1 | **0** | 1 | 1 |
| **both-good** | good | good | good | 1 | 6 | 1 | 1 |
| prod-devid | prod | – | Developer ID | 1 | 6 | 1 | 1 |
| prod-forged | prod | – | ad-hoc, forged identifier | **0** | **0** | 0 | 0 |

- In `listener-exec-stopped`, the delegate calls are the warmup peer and the fresh post-exec connection. There were zero `bad-*` invocations.
- In `control-exec-stopped`, `bad-*` calls *were* invoked.
- `both-good` is clean: a matching peer was served, and there was no API-misuse trap. `/usr/bin/log` has no misuse lines, and no crash report was written for `listener`. That is the basis for layer 2.

### Q1: timing

**Listener variant:**
- The delegate is not called for a non-matching peer, and nothing reached the exported object, including bursts.
- This matches the Foundation header in both the 26.6 and 27 SDKs: *"the incoming connection is automatically rejected before consulting the delegate"*.
- **Disassembly (macOS 27):** the method calls `xpc_connection_set_peer_code_signing_requirement` on the listener connection. Its header says: *"For a listener connection, requests that do not satisfy the requirement are dropped."*

**Connection variant (`conn-*`):**
- The delegate is consulted, and the counter opens and closes once.
- No method ran, even with 200 queued messages.

**Exec race, ordering forced:**
1. The listener is SIGSTOPped.
2. The bad process queues a request plus a burst, then `execv`s the good binary under the same PID.
3. The listener is SIGCONTed.

- Result: rejected. libxpc logged `xpc_support_check_token … status: -67065` (errSecCSNoSuchCode). A plain mismatch logs -67050.
- `control-exec-stopped` shows the race really ran.
- The script now fails loudly if the forcing didn't happen; see the probe section below.

**What a rejected client sees:** interruption + `NSCocoaErrorDomain 4097`, the same as `delegate-reject`.

### Q2: IdleExit

- A peer rejected by layer 1 never reaches the delegate, so it is never counted.
- A peer rejected by layer 2 is counted and then released:
  - probe `conn-bad`: opened 1, closed 1;
  - `aConnectionRequirementStopsWhatTheListenerLetThrough` waits for the close hook.
- Accepted connections were balanced in every scenario.

### Q3: uid binding

- **Disassembled on macOS 27:**
  - `effectiveUserIdentifier` → `xpc_connection_get_euid` → `[conn,#0x6c]`
  - `effectiveGroupIdentifier` → `xpc_connection_get_egid` → `[conn,#0x70]`
  - the old SPI `-[NSXPCConnection auditToken]` → `xpc_connection_get_audit_token` → copies `[conn,#0x68]` (32 bytes)
  - So the new values are `val[1]`/`val[2]` of the token the old code read: the same numbers `audit_token_to_euid/egid` produced, from the same field, in the same callback.
- **Header:** `xpc_connection_get_euid`: *"The EUID of the remote peer at the time the connection was made."*
- **Computest, CVE-2023-32405 write-up:** *"within -listener:shouldAcceptNewConnection: the audit token is safe"*. Source: https://web.archive.org/web/20241108053003/https://sector7.computest.nl/post/2023-10-xpc-audit-token-spoofing/
- **Also in the disassembly:** `_xpc_connection_check_peer_requirement` evaluates the requirement against each message trailer's audit token, and cancels a connection whose message pid/pidversion differs from the cached token (*"A message is received from other than initial peer."*).
- **Apple DTS:**
  - Quinn recommends `setCodeSigningRequirement` for "only my client may talk to my daemon": https://developer.apple.com/forums/thread/681053
  - Thread https://developer.apple.com/forums/thread/797913 says *"If the client process's code signature becomes invalid before it connects, this check will reject the connection."* That thread is about the **per-connection** `NSXPCConnection.setCodeSigningRequirement`. Applying it to the listener method is **my inference**. It does apply directly to layer 2.
- **Inference, not observed:** that the requirement check and the cached euid come from the *same* connection-request message.

## Tests and mutations

`App/Tests/HelperPeerGateTests.swift`: 14 test functions (20 cases counting the 7 parameterised team inputs).

- Real in-process XPC uses anonymous listeners, invented fixtures (`com.duoupdater.zzfixture.app`, team `ZZFIXTURE0`, uid 4242 / gid 4343) and no host state.
- The "matching" requirement is the test process's own designated requirement, read at run time. `xctest` is ad-hoc on both Xcode 26.6 and 27.
- Layer 2 is tested on its own with `RequirementSwappingListener`. It gives the real listener a requirement this process *meets*, while the gate's own (per-connection) requirement is `neverMatching`. The test asserts the delegate **did** accept (`exportedObjectsMade == 1`, `opened == 1`), so removing layer 1 can't make it vacuous.
- `NSXPCListener.anonymous()` can't be subclassed for this: called on a subclass it still returns `NSXPCListener`.

Every mutation below was applied through `scripts/app-tests.sh` (Xcode 27) and reverted automatically. Every red is a runtime failure. The one compile error, a first try at M11b, was fixed and re-run.

| # | Mutation | Result |
|---|---|---|
| M1 | delete `listener.setConnectionCodeSigningRequirement(requirement)` | red: `aPeerFailingTheRequirement…` (delegate now consulted), `installPassesTheRequirementVerbatim…`, `installForATeam…` |
| M2 | listener requirement → `"anchor apple"` | red: `aPeerMeetingTheRequirement…`, `installPassesTheRequirementVerbatim…`, `installForATeam…` |
| M3 | nil team accepts (`if let installation, !(… === listener)` instead of the guard; per-connection call only `if let`) | red: `withoutATeamEveryPeerIsRejected` |
| M4 | drop `installedOn === listener` | red: `aGateAttachedToAnotherListenerRejects` |
| M5 | `connectionOpened()` before the guard | red: `withoutATeam…`, `aGateAttachedToAnother…`, `aPeerMeeting…`, `aConnectionRequirementStops…` |
| M6 | `HelperClientIdentity(uid: 0, …)` | red: `aPeerMeeting…` |
| M7 | drop `conn.invalidationHandler = connectionClosed` | red (60 s time limit): `aPeerMeeting…`, `aConnectionRequirementStops…` |
| M8 | `listener.resume()` moved above the requirement | red: `installPassesTheRequirementVerbatim…`, `installForATeam…` |
| M10 | drop the `certificate leaf[subject.OU]` clause | red: `theClientRequirementIsPinnedForATeam`, `installForATeam…` |
| M11a | drop `team.utf8.count == 10` | red: `aMalformedTeamYieldsNoRequirement` |
| M11b | drop the `A-Z0-9` check | red: `aMalformedTeamYieldsNoRequirement` |
| M12a | `requirement(team:)` uses bundle id `com.duoupdater.cli` | red: `theClientRequirementIsPinnedForATeam`, `installForATeam…` |
| M12b | `install(on:team:)` passes `OwnTeamIdentifier.requirement(bundleIdentifier: "com.duoupdater.cli", team:)` | red: `installForATeam…` |
| M13 | **delete `conn.setCodeSigningRequirement(installation.requirement)`** | red: `aConnectionRequirementStopsWhatTheListenerLetThrough` |
| M14 | `identity(uid:gid:)` falls back to `"unknown"` | red: `aUidWithNoAccountHasNoIdentity` |
| M15a | `matchesClaim` ignores uid | red: `aClaimMustMatch…`, `installMASAppRefusesAClaim…` |
| M15b | `matchesClaim` ignores gid | red: `aClaimMustMatch…` |
| M15c | `matchesClaim` ignores userName | red: `aClaimMustMatch…` |
| M16 | delete the `matchesClaim` guard in `installMASApp` | red: `installMASAppRefusesAClaimThatDiffersFromItsConnection` |
| M17 | per-connection requirement → `"anchor apple"` | red: `aPeerMeeting…` |
| R1 (round 2) | `missingRequirementReason` always returns the "unavailable" text | red: `noTeamAndAMalformedTeamAreLoggedDifferently` |
| T2 (round 2) | tripwire fires (`#require(Bundle.main.bundleIdentifier != Bundle.main.bundleIdentifier)`) | red: `installMASAppRefusesAClaim…`, fails at the `#require` before `installMASApp` is called |
| T1 (round 2) | tripwire always passes (both `#require`s neutralised) | green, as intended. The case still asserts the matching request ends at `"invalid log path"`, i.e. before `bundledMASPath`/`mas`. Confirmed the build recompiled the test (`Compiling HelperPeerGateTests.swift` in that run's log). `mas` was not run. |
| **G-3 (M9)** | `install(on:team:)` default `team: String? = nil` | **green (known uncaught)** |
| **G-2** | `effectiveUserIdentifier`/`GroupIdentifier` → `geteuid()`/`getegid()` | **green (known uncaught)** |

The fixture team is valid under the new rule (`theFixtureTeamIsWellFormed`).

### Known uncaught mutations (also in the test-file header)

| | Mutation | Why no in-process case sees it | Manual step that covers it |
|---|---|---|---|
| G-2 | euid/egid from `geteuid()`/`getegid()` | peer and listener are one process | step 1: in the real root helper this would bind every client to uid 0, so `installMASApp` would refuse the app's own uid ("client identity did not match its XPC connection"). A successful App Store install rules it out. |
| G-3 | default `team:` changed (nil or another literal) | test host is ad-hoc, `OwnTeamIdentifier.current` is nil anyway | steps 1 and 2: a signed install serves the app and refuses a forged client |
| G-6 | `main.swift` writes `_ = HelperPeerGate.install(...)` | top-level code isn't compiled into tests; the weakly held delegate is deallocated | steps 1 and 2 |
| (listener-only weakening) | a requirement the test process also fails (e.g. `anchor apple`) | same reason | caught by the recorder cases instead (M2) |

G-4 (a uid with no passwd entry) is covered via `identity(uid: uid_t.max, gid: 0)`: `(uid_t)-1` is the `setreuid`/`chown` "no change" sentinel. Mutation M14.

## Probe script hardening

- `cleanup` now resumes any SIGSTOPped listener, boots out every agent and runs `rm -rf "$T"`.
- The forced exec race fails loudly in three cases:
  - no listener pid;
  - `ps -o stat=` doesn't show `T` after STOP;
  - a listener is still stopped when the peer function returns.
- Each failure path was run once by editing `run.sh` in place and restoring it:

```
F1 (no listener pid):     exit=1  PROBE FATAL in listener-exec-stopped: no listener pid in …/listener-exec-stopped.listener.log — the race would run unforced
F2 (kill -STOP → kill -0): exit=1  PROBE FATAL in listener-exec-stopped: listener pid=14335 not stopped after SIGSTOP (stat='S   ') — the race would run unforced
F3 (exit between STOP/CONT): exit=1 PROBE FATAL in listener-exec-stopped: a listener was still SIGSTOPped when the peer finished (resumed now)
all three: workdir still exists: False, zzprobe agents left: 0, stopped listener processes: 0
```

## Verification

Round 2, head 4ff2e8b3:

```
$ scripts/run-with-hang-report.sh 1800 make test      → exit 0
Ran 5 tests in 0.000s   (test_app_test_coverage.py)
━ Test run with 2867 tests in 245 suites passed after 19.644 seconds with 1 known issue.
✔ Test run with 300 tests in 36 suites passed after 0.230 seconds.
✓ version comparisons discriminate …   ✓ app audits consistent …   ✓ engine-notes pointers resolve …
✓ skill docs consistent …   ✓ no machine-population claims in code comments — 551 files
✓ no blocking calls left on the cooperative pool — 51 call(s) in 266 files

$ scripts/app-tests.sh ×12                                                      → 12/12 ✓ 48 of 48 (Xcode 27)
$ DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer scripts/app-tests.sh → ✓ App tests — 48 of 48 cases executed (Xcode 26.6 / Swift 6.3.3)
$ swiftc -O scripts/xpc-peer-probe/InProcessDoubleRequirement.swift && ./probe → OVERALL: PASS (Xcode 27; also built with Xcode 26.6 in Swift 5 mode, PASS)
```

Round 1 (head 520a56bc), unchanged since: `DuoUpdaterHelper` Release builds on Xcode 27 and 26.6; `PROBE_DEVID=… scripts/xpc-peer-probe/run.sh` exit 0 with the table above.

## 未验证

- **Mach-service, root and separately signed peers below macOS 27.** On 14, 15 and 26, only the in-process anonymous-listener checks ran (see Coverage by OS). The launchd probe, the exec race and the disassembly are macOS 27 only.
- **Layer 2 double-set on 14/15/26 in the real configuration.** In-process it did not trap on any of them. That a Mach-service listener in a root daemon behaves the same is inferred, not observed.
- **The real helper.** The probe used a user agent with a plain client, not a root LaunchDaemon with a `.privileged` client. I did not build, install or register the helper.
- **Cross-user.** No peer ran as a different user. G-2 stays manual.
- **Same message for both values.** That the requirement check and `effectiveUserIdentifier` come from the same message is inferred.
- **Team casing.** That Team IDs are uppercase is observed, not documented. Apple documents "10-character" (https://developer.apple.com/help/glossary/team-id) and "10 alphanumeric" (DTS, https://developer.apple.com/forums/thread/767608). A lowercase team would be refused: closed, not open.
- **CI for this head.**

## Risks (attack these)

1. **Apple behaviour, two layers.** Layer 1 relies on a header sentence; layer 2 relies on per-message checking. Both were observed with a Mach service only on macOS 27, and in-process on 14, 15, 26 and 27. With both in place, a regression in one layer alone no longer opens the gate.
2. **Layer 2 on older OSes:** in-process, no trap on 14.8.9, 15.7.9, 26 or 27. If the real configuration differed, the trap would be a crash, i.e. no service, rather than open access.
3. **Malformed requirement:** now unreachable from a malformed team (nil → reject all). A malformed `clientBundleIdentifier` would still raise; it is a pinned constant.
4. **Diagnostics:**
   - Layer-1 rejections are no longer `NSLog`ged by the helper. libxpc logs `xpc_support_check_token … status: -670xx` at Error level, with the requirement `<private>`.
   - Layer-2 rejections don't log anything from our code either.
5. **`HelperService.swift` is now compiled into the test bundle.** The tests stop at the lexical log-path check (`/ZZFixture-not-a-temp-dir/…`), before any `open` or `Process`. A future reorder of those guards could let a test reach `bundledMASPath()`. That resolves relative to `xctest` and doesn't exist there, so the request is refused ("bundled mas not found") before `/bin/sh` runs. The test also has a tripwire (round 2): it fails before calling the service if the main bundle is `com.duoupdater.app`, or if an executable `mas` sits beside the main executable.

## Pre-existing, not changed

- No version floor in the requirement (an older signed DuoUpdater from the same team still satisfies it).
- Debug builds of the helper lack the hardened runtime (Release-only setting).
- The helper runs the bundled `mas` without verifying its signature.
- `PrivilegedHelperClient.isConnectionFailure`'s comment swaps the codes: 4097 is `NSXPCConnectionInterrupted`, 4099 is `NSXPCConnectionInvalid`. The check itself covers both.

## Manual runtime verification (for you, with the real installed helper)

1. **Normal install.** Install this branch and trigger a Mac App Store update from DuoUpdater. Expected:
   - it installs;
   - success itself covers G-2 (a helper bound to uid 0 would refuse the app's uid) and G-3 (no requirement ⇒ every connection refused);
   - `/usr/bin/log show --last 10m --predicate 'process == "DuoUpdaterHelper"'` shows no rejection.
2. **Forged client** (covers G-3 and G-6). Build a tiny client that calls `helperVersion` on `com.duoupdater.helper` with `NSXPCConnection(machServiceName:options: .privileged)` and `MASHelperProtocol` (copy `App/Shared/MASHelperProtocol.swift`), ad-hoc sign it with `codesign -s - -i com.duoupdater.app`, and run it. Expected:
   - no reply, and an error of 4097;
   - the helper logs `xpc_support_check_token … -67050`;
   - no `duo-helper: rejected connection` line.
3. **Idle exit.** After step 2, `pgrep -x DuoUpdaterHelper` should disappear within ~60 s of idleness.
4. **Older OS.** On a macOS 14/15 machine (the mini?), repeat steps 1–3 and run `scripts/xpc-peer-probe/run.sh`. Its `both-good` and `conn-bad` are the Mach-service versions of what `InProcessDoubleRequirement` already showed in-process there.


jizhi0v0:
## Runtime verification against the real installed helper (macOS 27.0, 26A428)

The installed build is this branch with `origin/main` merged in (17011fa9), installed with `make install` (Developer ID). The helper ran as a root LaunchDaemon (`root 49753 Contents/MacOS/DuoUpdaterHelper`) and was reached over `NSXPCConnection(machServiceName: "com.duoupdater.helper", options: .privileged)`.

Two throwaway clients were built from `MASHelperProtocol.swift`. Both use identifier `com.duoupdater.app`; one is ad-hoc signed, the other Developer ID signed with team RS59HDH7Y3. Neither installed anything: both `installMASApp` calls use `adamID: 1` and a log path outside the temp directories, so the helper returns before `mas` is looked up.

| Client | Call | Before: installed 0.3.93, old code | After: this branch |
|---|---|---|---|
| ad-hoc | `helperVersion` | `NSCocoaErrorDomain 4097` | `NSCocoaErrorDomain 4097` |
| Developer ID | `helperVersion` | `version=102` | `version=102` |
| Developer ID, uid+7777 | `installMASApp` | `client identity did not match its XPC audit token` | `client identity did not match its XPC connection` |
| Developer ID, own uid | `installMASApp` | `invalid log path` | `invalid log path` |

The changed message text in the third row shows the second run reached the new helper.

Helper log for the new binary (`log show --predicate 'process == "DuoUpdaterHelper"'`):

```
19:38:14.710 (libxpc.dylib) Code-signing requirement (anchor apple generic and identifier "com.duoupdater.app" and certificate leaf[subject.OU] = "RS59HDH7Y3") set on connection
19:38:14.710 [com.apple.xpc:connection] activating connection: mach=true listener=true peer=false name=com.duoupdater.helper
19:38:14.713 (XPCSupport) xpc_support_check_token: <private> error: <private> status: -67050
19:38:14.713 [com.apple.xpc:connection] Dropping check-in message due to code signing requirement: <private>
19:38:14.729 (libxpc.dylib) Code-signing requirement (…same…) set on connection
19:38:14.729 activating connection: … peer=true name=com.duoupdater.helper.peer[49758]…
19:38:14.743 (libxpc.dylib) Code-signing requirement (…same…) set on connection
19:38:14.743 activating connection: … peer=true name=com.duoupdater.helper.peer[49760]…
19:38:14.760 (libxpc.dylib) Code-signing requirement (…same…) set on connection
19:38:14.760 activating connection: … peer=true name=com.duoupdater.helper.peer[49761]…
```

What the log shows:

- **Layer 1, in the real root Mach-service daemon.** libxpc dropped the ad-hoc peer (-67050), and no `duo-helper: rejected connection` line was logged, so the delegate was never consulted.
- **Layer 2, in the same setup.** Each of the three accepted Developer ID connections had the requirement set on it a second time, with no trap. All three calls were served.
- **G-2 (uid binding).** A wrong uid is refused and the caller's own uid gets past the identity check, so identity is bound to the connection's euid, not the helper's.
- **G-3 / G-6.** A signed client is served and an unsigned one refused, so the requirement is installed and the gate object is alive.
- **IdleExit.** After the last connection closed at about 19:38:15, the helper was gone by 19:39:18, within the 60 s window.

Also on the merged tree (17011fa9): `scripts/run-with-hang-report.sh 1800 make test` exited 0. Core 2916 tests passed (1 known issue), CLI 300 passed, App tests 48 of 48, and every guard passed.

**Still unverified:** a real App Store install through this helper (none was pending); Mach-service/root behaviour below macOS 27 (only the in-process probe ran on 14.8.9, 15.7.9 and 26); cross-user peers.

PR #597, opened 2026-09-13
Run every child process through swift-subprocess (ChildProcess)

jizhi0v0:
## What and why

Every child process in `DuoUpdaterCore`, `duo` and the menu-bar app now goes through one helper, `ChildProcess` (`DuoUpdaterCore/Sources/DuoUpdaterCore/Support/ChildProcess.swift`), on [swift-subprocess 1.0.0](https://github.com/swiftlang/swift-subprocess/tree/1.0.0). It replaces `Foundation.Process` + `waitUntilExit()` / `readDataToEndOfFile()`, whose waits each held a thread for the child's life: on the cooperative pool that is #351, off it one Dispatch thread per wait. swift-subprocess monitors exit with kqueue `EVFILT_PROC`/`NOTE_EXIT` and reads the pipes through the same kqueue.

Excluded per the brief: `App/Helper/`, `App/Spike/`. Also kept on `Process`: `DiagnosticsSettingsPage.relaunch`. It runs `open -n` and then terminates the app, so nothing waits on it (exempted in `check_offpool.py`, with the reason).

Dependencies are pinned exactly, like Sparkle: swift-subprocess 1.0.0 and swift-system 1.8.1. swift-system is listed so the transitive `from: "1.5.0"` cannot float. `Package.resolved` stays gitignored.

### `ChildProcess` policy
- **Pipes:** both are drained concurrently, always.
- **No output limit.** In swift-subprocess, `limit:` **throws** `outputLimitExceeded` on overflow rather than truncating (read in `IO/Output.swift`). Every site therefore collects unbounded, as it did.
- **`terminationStatus`** keeps `Process`'s convention: the exit code, or the signal number with `uncaughtSignal`. Error strings like "tar failed (15)" read the same.
- **stdin** is inherited unless bytes are given. A process started with **no fd 0** (`duo … <&-`) calls `ChildProcess.ensureStandardInputIsOpen()` once at startup, so children get `/dev/null` as a `Process` child did. `duo` does this in `main.swift`; launchd already gives the app `/dev/null`. A spawn that still finds fd 0 closed logs it once (round 1 item 8, reworked in round 2).
- **Environment** is inherited unless given, in which case it replaces the whole environment (as `Process.environment` did). Entries POSIX forbids (key with `=`/NUL or a leading digit, value with NUL) are **dropped and their keys logged**, instead of failing the spawn (round 1, item 6).
- **Process group:** the child leads its own (`processGroupID = 0`), as a `Process` child did, so a terminal Ctrl-C does not reach a `ditto` mid-swap (round 1, item 7).
- **`Deadline(terminateAfter:killAfter:)`:** SIGTERM, then SIGKILL after the difference. **The clock starts at launch** (round 1, item 4), as the old work items started after `Process.run()` returned. Every spawn in the process goes through swift-subprocess's single spawn worker thread; a slow spawn delays the others. That latency is **not measured** (see round 1, item 4).
- **`onCancel` has no default.**
  - `.runToCompletion` runs the child in `Task.detached`, so the caller's cancellation cannot reach it. SE-0504's shield is not back-deployed.
  - `.terminateChild` tears the child down (the ladder, else SIGKILL) and throws `CancellationError`. An already-cancelled task never spawns.
- **The drain runs in an unstructured task.** swift-subprocess's reads wait on an `AsyncStream`, which ends when its task is cancelled. Read inline, a deadline's SIGTERM closed the pipes at once, and a child writing during shutdown died of SIGPIPE (status 13). Found by `aChildHonouringSIGTERMExitsOnItsOwnTerms`.
- **One deliberate difference:** reads end when the child exits, even if a grandchild still holds the pipe. Measured with `sh -c 'sleep 5 & echo hi'`: `Process` + `readDataToEndOfFile()` took **5.08 s**, `ChildProcess.run` took **0.01 s**.
- **`onLaunch(pid)`** was added for #595's process-tree kill.

### Cancellation decisions (install path)
- **Everything that writes is `.runToCompletion`, which is what the uncancellable hops gave:**
  - `ditto`, `tar`, `hdiutil attach/detach`
  - `BinaryDelta`
  - the swap's `xattr`/`chmod`/privileged `osascript`
  - backup `ditto`/`chflags`
  - `brew install`/`upgrade`/`update`
  - the pkg gate's `pkgutil`/`xar`/`lsbom` (a killed read would turn into a fail-closed "declares nothing", not a cancellation)
  - `gh issue …`, the helper kickstart
- **Reads whose result outlives a cancel are also `.runToCompletion`:** `lsappinfo`, the login-shell probe, `brew config`, and since round 1 `brew outdated/leaves/list`, `brew info`, `gh auth token/status`, `mas outdated`, `launchctl print`, the probe's `unzip`.
- **`.terminateChild`, only where nothing uses the result after a cancel:**
  - `open -g -b` / `open -g <url>` (App Store): a cancel throws out of the install; the two `try? navigate` callers are best-effort and sleep, cancellably, straight after.
  - opencode (`duo triage`): the throw becomes a failure `run` logs and skips; a suggestion is only written on success. Nothing in `duo` cancels it today.

### Coordination
Rebased over #587, #589, #590, #592, #593, #594, #595 and #596. Not pushed before the last rebase, so no merge commits.
- **#590:** kept the explicit `_ =` on `InPlaceSwap.replace` in both installers, and the reason. Rewrote `aCleanupFailureReasonReachesTheInstallLog` around the async `replace`: fixture and log reads stay in hops, `replace` is awaited between them.
- **#592:** kept its `stripQuarantine` (`-drs`, exit status, log line, walk). The launch moved to `ChildProcess`, the walk to a hop, and `QuarantineStripTests` became async.
- **#596:** kept `StreamedLines` (lines not chunks, full text, tail). `StreamedLines.run` now runs on `ChildProcess`. Its `end()`/`waitForEnd` have no production caller now; they and their tests are kept. Added `standardErrorArrivesInTheSameStream`, because discarding stderr passed the whole suite.
- **#595:** migrated `LoginShellEnvironment` and `HomebrewConfig.read` (rows 28–29).
- **#593:** its CLI `Settings` changes kept; `load()` is async.

## Review
An adversarial review (read-only, separate model) found the following; each was checked against the code before acting.
- **Fixed, B:** `lsappinfo` under `.terminateChild` let a cancelled restart recheck proceed with an empty map. That clears every Restart badge and writes marketing history without the running builds. Now `.runToCompletion`.
- **Fixed, C:** synchronous FileManager work that deletes whole bundles or data dirs moved onto the pool when its hop disappeared. Each segment now has its own hop (listed under the per-site table).
- **Fixed, D:** the chmod/xattr cancellation mutations depended on SIGKILL winning a race. `.terminateChild` now refuses to spawn into a cancelled task.
- **Fixed, E:** stale comments (the offpool fixture, "measured from launch", this doc).
- **Fixed:** `check_offpool` launch spellings: `Process.init(`, `Process.run(`, `Process.launchedProcess(`, spaced `Process ( )`, `Foundation.Process()`, `NSTask`, `posix_spawn`.
- **Fixed:** the deadline tests now leave 8 s before SIGTERM, so a slow spawn on a loaded runner cannot beat the child's `trap`.
- **Refuted for shipped binaries, A (`duo` needs `libswiftCompatibilitySpan.dylib` on macOS 14/15):**
  - A **debug** `swift build` of `duo` does link `@rpath/libswiftCompatibilitySpan.dylib`.
  - The Release `duo-cli` (xcodebuild, as `make cli` builds it) does not.
  - Neither does the Release `DuoUpdater.app`, with Xcode 27 RC or with Xcode 26.6.
  - Not run on a 14/15 machine — see the unverified list.
- **Remaining, documented, not fixed:**
  - SIGPIPE in our process if bytes are written to a child that exits without reading them. No caller passes bytes.
  - swift-subprocess tears down a `.runToCompletion` child if a pipe read throws.
  - A microsecond PID-reuse window in upstream teardown.
  - `offpool-lint:allow` on a launch also exempts the waits in that scope.
  - `let p: Process = .init()` is not caught.

## Review round 2

1. **Back-deploy gate false positive: fixed.** Dependencies now come from the dylib load commands in `otool -l` (`LC_LOAD_DYLIB`, `LC_LOAD_WEAK_DYLIB`, `LC_REEXPORT_DYLIB`, `LC_LAZY_LOAD_DYLIB`, `LC_LOAD_UPWARD_DYLIB`), never `LC_ID_DYLIB`.
   - `test_a_correctly_embedded_real_shim_passes` copies the active toolchain's `libswiftCompatibilitySpan.dylib` into a scratch bundle and runs `check()` on it. It first asserts, with `otool -D`, that the copy names itself `/usr/lib/swift/…`. Nothing loads the copy.
   - The round-1 parser fails that test (G1), and so does counting `LC_ID_DYLIB` as a dependency (G3).
   - Parity: across the four Release products and the Debug `duo`, the new dependency lists equal the old `otool -L` lists minus the file's own id. The only file where that differed was Sparkle's framework binary.
   - `@executable_path` is now per executable (`otool -hv` says `EXECUTE`), which covers Sparkle's `Updater.app`, `Autoupdate` and the XPC services (G2). Documented, not modelled: a library loaded by a nested executable is still resolved against the main one.
   - The test passes with both Xcode 27 and 26.6 as the active toolchain.
2. **`fetchThenSwap` work dir:** removed through `removeItemOffCooperativePool` after the apply phase, on every exit. `FetchThenSwapCleanupTests` drives `fetchThenSwap` with stub closures and covers success, a throwing apply, and a cancel before apply (W1, W2). The pkg gate's `PackageInfo`/`Bom` scratch and the probe's stub zip keep their synchronous `defer`, with a comment saying why.
3. **Leaked login-shell timer:** the runner finishing now cancels the timer, and `waitForLaunch` honours cancellation (same shape as `LaunchSignal`). `aShellThatCannotStartLeavesNoTimerBehind` uses an invented shell path and a 600 s timeout, so the timer can only end by being released. The test sees that end through a `timerExited` seam (T1, T2).
4. **`theTimeoutCountsFromTheShellsLaunch`:** now `timeout: 5` against 8 s before the spawn. Still red with the clock from the call (U4r2).
5. **fd 0:** no per-spawn fill any more. `ensureStandardInputIsOpen()` is the first statement of `duo`'s `main.swift`.
   - The app does not call it. `lsof` shows fd 0 `/dev/null` on the running DuoUpdater and on Finder. The SwiftUI `App`'s first code of ours also runs after `AppListModel`, which starts threads, is built.
   - A spawn that finds fd 0 closed logs once and inherits the hole.
   - Probe under `<&-`, same binary: without the call the child prints `/dev/fd/0: Bad file descriptor`; with it first, `cr--r--r-- … /dev/fd/0`.
   - The fill itself keeps F1/F2. The old decision-function test went away with the function.
6. **Docs:**
   - `Deadline.init` now says "from the child's launch".
   - `zipEntryPlistValue`'s doc comment is back above it, and `extractZipEntry` has its own.
   - The test comments name `removeItemOffCooperativePool`.
7. **SIGTTOU:** added to the unverified list.

**Where I disagree or qualify:** nothing substantive.
- Item 1: parsing load commands made `otool -D` unnecessary, and `-D` alone would not have excluded the id from `-L`'s dependency list any more simply.
- Item 5: "before other threads exist" holds for `duo`. For the app it cannot, so it rests on launchd's `/dev/null`, which I measured rather than assumed.

## Review round 1

All 14 items. Behaviour changes carry a test whose named mutation was seen red (table below, rows after O2).

1. **Brew reads killed on cancel:** `brew outdated/leaves/list` and `brew info` are `.runToCompletion`. Every remaining `.terminateChild` site was audited; the consumers are in the per-site table (rows 14–25). Four more reads turned out to have the same shape and moved too: `gh auth token/status` (token cached ten minutes), `mas outdated` (nil means "could not check", which the pre-flight treats as leave to skip), `launchctl print` in `duo doctor`. Only App Store `open` and triage opencode stay killable. CLAUDE.md now lists them. Tests go through the seams with invented scripts from a task cancelled before it starts, so under `.terminateChild` the child never spawns and the mutations fail deterministically.
2. **Probe `unzip`:** `.runToCompletion`, with `aCancelledZipEntryReadStillAnswers`.
3. **Login-shell timeout before the pid is known:** the timer now starts when the shell launches (`LaunchedPID.waitForLaunch`), so a timeout always has a pid to kill. That makes the late-kill case impossible rather than handling it. `theTimeoutCountsFromTheShellsLaunch` injects a 2 s pre-spawn wait against a 0.5 s timeout. `aShellThatHangsIsUnknownAndKilled` is not timing-sensitive: its rc loops forever, so the call can only finish through the timeout. It now has a 60 s watchdog, so a lost pid fails the test instead of hanging it.
4. **Deadline clock:** starts at launch (`LaunchSignal`, fired from the `onLaunch` wrapper). `theDeadlineClockStartsAtLaunch` injects 6 s before the spawn against `terminateAfter` 3 s. The SIGTERM tests' 8 s margin was about a slow spawn beating the child's `trap`. From launch that race is gone, and the margin is kept anyway. The single spawn thread is documented. **The latency measurement was abandoned, and it went wrong on this machine.** The probe gave swift-subprocess a slow-to-exec binary by forging a quarantine xattr on an ad-hoc binary. That put a Gatekeeper dialog ("q1 Not Opened") on the user's desktop. The coordinator stopped it. A second invocation I had already sent before seeing the stop ran the same binary again (its log shows a 10.59 s first run), so a second dialog very likely appeared too. The binaries are deleted, the numbers are discarded, and nothing in this PR relies on them.
5. **Whole-bundle deletions on the pool:** all go through a new `removeItemOffCooperativePool(at:)`:
   - `BackupStore.save`: its five failure-branch staging removals and the legacy-key cleanup.
   - `BackupStore.restore`: the scratch pre-clean and cleanup.
   - `InputMethodDataBackup.save`/`restore`: the staging and scratch cleanup.
   - Found by grepping: the leftover removals before a copy in `ArchiveExtractor.copyApp`, `PackageInstaller.copyPackage` and `DeltaApplier.reconstruct`.
   - Also found: `DeltaApplier.reconstruct` read and hashed the whole patch on the pool once it became async, so that is back on Dispatch.
   - The app's orphan pruning and "delete backups" used `Task.detached`, which is still the pool; they now use `offCooperativePool`. These two were pre-existing on main.

   `defer` cannot `await`, so the work between creating a directory and removing it moved into a helper, and the removal follows the call on every exit. Each removal is pinned by a test (B1–B3, I1–I2); which thread does it is not observable. Left alone: `PackageInstaller.declaredDestinations`' scratch (a few `PackageInfo`/`Bom` files), and the installers' `workDir` removals in `download` and `fetchThenSwap`, which were already on the pool on main.
6. **Environment validation:** invalid entries are dropped and their keys logged (`Log.app.error`). Tested with `1BAD`, `A=B`, a NUL key, a NUL value, end to end through a spawn.
7. **Ctrl-C / process groups:** measured with a probe that puts itself in its own group and `killpg`s it with SIGINT. Before the fix, the `Process` child was in its own group and survived. The `ChildProcess` child shared the probe's group and died (status 2, signaled). After `processGroupID = 0`:
   ```
   PID  PPID  PGID COMM
   43856 42562 43856 probe
   43865 43856 43865 /bin/sleep      # Foundation.Process
   43866 43856 43866 /bin/sleep      # ChildProcess
   Foundation.Process child running after SIGINT: true
   ChildProcess child alive after SIGINT: true
   ```
   Pinned by `theChildLeadsItsOwnProcessGroup` (P1).
8. **stdin with fd 0 closed:** measured under `<&-`. Both spawn: neither fails. They differ: a `Process` child gets a character-device fd 0 (`/dev/null`), and a `ChildProcess` child had none (`ls: /dev/fd/0: Bad file descriptor`). **My first fix was wrong while its tests were green.** It chose swift-subprocess's own `/dev/null` input when fd 0 was closed. The decision test was right, but with 0 a hole, swift-subprocess opens that descriptor (or a pipe end) at 0, and its spawn file actions close it in the child. Re-running the probe on the pushed head showed that. The fix: point our own fd 0 at `/dev/null` (once), then inherit it. (Round 2, item 5 moved that fill to process start: done per spawn it raced other threads.) Probe on the head, then with that one call removed:
   ```
   head:      ChildProcess: out "via-ChildProcess\ncr--r--r--  1 root  wheel  0x3000002 … /dev/fd/0\n"
   mutation:  ChildProcess: out "via-ChildProcess\nls: /dev/fd/0: Bad file descriptor\nno-fd0\n"
   ```
   Not switched to `/dev/null` in general: an open fd 0 is still inherited (brew's sudo prompt). `aClosedDescriptorIsPointedAtDevNull` pins the fill on a closed high descriptor (F1, F2). Closing the test process's own fd 0 would reach every parallel test, so the fd 0 wiring rests on the probe run above.
9. **Back-deployment gate:** `scripts/check_swift_backdeploy.py` (tests in `make test`). It fails a `libswiftCompatibility*` shim not embedded in what ships, and any other `@rpath/libswift*` that neither an in-bundle run path nor `/usr/lib/swift` resolves. It runs after the build in `build-cli.sh` (duo-cli), `notarize.sh` (what `release-build.yml` runs via `make notarize`: app, helper, BinaryDelta, Sparkle) and `install.sh`. Against the Debug `duo` it fails. The first draft also failed Sparkle's `BinaryDelta`, which links `@rpath/libswiftCore` through a `/usr/lib/swift` run path, the pre-5.1 way; that is now accepted and has a test. The otool results are under Verification. Context: this machine (macOS 27) loads `/usr/lib/swift/libswiftCompatibilitySpan.dylib`; older systems lack it, and a Subprocess-using helper failing on exactly that is reported at [developer.apple.com/forums/thread/817488](https://developer.apple.com/forums/thread/817488).
10. **check_offpool floor:** per spelling. Each `BLOCKING` spelling must match at least one call, or sit in `RETIRED` with a reason (`waitUntilExit()`, bare `.wait()`). A retired spelling that matches again fails, and so does a retired key not in `BLOCKING`. Renaming the single `SecStaticCodeCheckValidity` is red now. The old total floor would still have passed it: 8 calls left against a floor of 5.
11. **H17:** `anAlreadyCancelledTaskNeverSpawns` counts `onLaunch` calls, so it is deterministic. The ladder margins are covered in item 4.
12. **Leaked `sleep 120`:** the fixture uses a per-run duration and kills it at the end.
13. **Docs:** `homeOverride` now says every read of `home` is on the calling task, so the seam also reaches a test that drives `backUp`. The stray blank line in `brewInfoOffActor` is gone, the `BrewFormulaServiceTests` header is rewritten, and the CLAUDE.md section is updated (killable list, deletions, per-spelling floor, Span gate). A grep of `.claude/skills`, `.agents` and `docs` found no copies. The "Timeouts … from launch" line in this body is true again.
14. **`revision:` vs `exact:`:** unchanged. Opinion: `exact:` is enough here. SwiftPM refuses a tag whose commit changed once it is in `Package.resolved`, but that file is gitignored in this repo, so neither pin protects a fresh clone from a re-pointed tag. `revision:` would, at the cost of Dependabot and readable diffs. swift-system is pinned `exact: 1.8.1` alongside, so nothing floats transitively. If tag mutation is a real concern, committing `Package.resolved` protects both packages at once. That is a repository decision, not this PR's.

**Things in the list I think were off:**
- **Item 3:** fixed by starting the timer at launch rather than by adding a late kill. With the clock started at launch, a timeout without a pid cannot happen.
- **Item 4:** the measurement approach raised system UI. Measuring a slow exec needs a binary that is slow to exec, and the obvious way to get one on macOS is Gatekeeper.
- **Item 8:** "stdin with fd 0 closed" did not make spawns fail, as the item suggested it might. The two differed in what the child got, and my first fix for that difference was itself wrong.
- **Item 9:** "any @rpath Swift compat dylib" would have failed every app build on Sparkle's `BinaryDelta`.

## Per-site table

29 `Process` launch sites in scope. Recount at the start (`origin/main` 6d5773aa): 39 textual `waitUntilExit` matches including comments, 27 launches (the brief's 41 was not reproduced). #595 (`HomebrewSelfUpdate.swift`) added two more while this branch was open (rows 28–29). After this PR the scoped roots contain one `Process()` (the exempted row —). "Streams" = stdout / stderr; "inh" = inherited from our process. Timeouts are SIGTERM / SIGKILL counted from launch (since round 1; before it they counted from the call).

| # | Site | Tool | Timeout ladder | Streams (before → after) | Env / cwd / stdin | Status & strings | onCancel | Test |
|---|---|---|---|---|---|---|---|---|
| 1 | `ArchiveExtractor.run` (Core) | `hdiutil attach/detach`, `ditto`, `tar` | 300 / 305 s kept | pipe+DispatchGroup drain / pipe → both captured, concurrent | inh / none / inh | `toolFailed(tool, code, stderr)` — code is signal no. on a kill, as before | runToCompletion | `aCancelledExtractionStillUnpacksTheApp`, `aFailingToolReportsItsStatusAndStderr`, download gate (real zip + dmg) |
| 2 | `ArchiveExtractor.detach` | `hdiutil detach -force` ×2 | via #1 | via #1 | — | exit code only | runToCompletion; the 0.5 s retry pause runs in a detached task so a cancel cannot skip it | download gate (dmg); no leaked mount afterwards (`hdiutil info`) |
| 3 | `DeltaApplier.apply` | `BinaryDelta` | 600 / 605 s kept | `/dev/null` → drained+discarded / pipe → captured | inh / none / inh | `applyFailed(code, lastStderrLine)` | runToCompletion | `aFailedPatchReportsTheToolsStatusAndLastStderrLine`, `aCancelledPatchStillFinishes` (invented tool in invented bundle) |
| 4 | `InPlaceSwap.stripQuarantine` (#592's version) | `xattr -drs` | none | null / null → discard / discard | inh | #592's exit-status handling and log line kept; its bundle walk hopped | runToCompletion | `aCancelledSwapStillStripsQuarantine`, #592's `QuarantineStripTests` (made async) |
| 5 | `InPlaceSwap.rotateContents` | `chmod -R g+w` | none | **inh** / null → discard / discard | inh | non-zero → log line, as before | runToCompletion | `aCancelledRotationStillCarriesTheGroupWriteBit` |
| 6 | `InPlaceSwap.privilegedReplace` | `osascript … with administrator privileges` | none | **inh** / pipe → discard / captured | inh | `isAuthorizationDeclined` on stderr, cleanup classification unchanged; `NSLock` panel lock → one-permit `InstallPermits` | runToCompletion | **untested** (needs the admin panel) |
| 7 | `BackupStore.runDitto` | `ditto` | none | null / **temp file** → discard / captured | inh | `DittoOutcome(status, stderrTail)` same format | runToCompletion | `aCancelledBackupAndRollbackStillComplete`, existing `BackupStoreTests` |
| 8 | `BackupStore.clearUserImmutableFlags` | `chflags -R nouchg` | none | null / null → discard / discard | inh | Bool | runToCompletion | existing `anImmutableFileInTheSourceDoesNotFreezeTheStoredCopy` (mutation B1) |
| 9 | `InputMethodDataBackup.copyTree` | `ditto` | none | null / null → discard / discard | inh | Bool | runToCompletion | existing `aSnapshotRestoresTheDataTheUpdateChanged` (mutation M1) |
| 10 | `PackageInstaller.runCapturing` | `xar -tf`, `xar -xf` (cwd), `lsbom` | none | one pipe for both → merged | inh / **cwd kept** / inh | (-1, "") on launch failure, as before | runToCompletion (feeds a fail-closed gate) | `aCancelledPackageGateStillReadsTheDeclaredDestinations` (real `pkgbuild` package) |
| 11 | `PackageInstaller.runCapturingOutput` | `pkgutil --check-signature` | none | merged → merged (now delegates to #10) | inh | same | runToCompletion | parser unit tests only; **the `pkgutil` launch itself is untested** |
| 12 | `PackageInstaller.run` | `hdiutil attach/detach`, `ditto` (dmg-wrapped pkg) | none | null / null → discard / discard | inh | exit code; detach still a single attempt | runToCompletion | **untested** (no dmg-pkg fixture) |
| 13 | `HomebrewInstaller.upgrade` | `brew install --cask --force` | none | (#596's `StreamedLines.run` on `Process`) → merged, lines via `StreamedLines` on `onOutputChunk` | `*_proxy` env kept | `BrewError.failed(code, full text)` | runToCompletion | `StreamedLinesTests` (real `/bin/sh` through the same runner, incl. new `standardErrorArrivesInTheSameStream`); brew itself **untested** |
| 14 | `BrewFormulaService.executor(brewPath:)` | `brew outdated/leaves/list` | none | pipe / null → captured / discard | `*_proxy` + `HOMEBREW_NO_AUTO_UPDATE` kept | (status, stdout) | **runToCompletion** (r1: `refreshBrewFormulae` runs from a view `.task` and writes what it read) | `aCancelledRefreshStillReadsRealData` (real executor, invented `brew` script; B2); overlap test (I11) |
| 15 | `BrewFormulaService.run` | `brew upgrade` / `brew update` (#595) | none | as #13 | as #13 + user's shell `HOMEBREW_*` | as #13 | runToCompletion | as #13 |
| 16 | `BrewFormulaReleaseService.brewInfoOffActor` | `brew info --json=v2` | none | pipe / null → captured / discard | as #14 | nil on failure | **runToCompletion** (r1) | `aCancelledBrewInfoStillAnswers` (invented script; B3); `BrewFormulaReleaseActorTests` |
| 17 | `MASInstaller.outdatedAdamIDs(mas:)` | `mas outdated` | **20 s TERM kept, 25 s KILL added** | pipe read after exit / null → captured concurrently / discard | inh | nil on failure/timeout | **runToCompletion** (r1: nil reads as "could not check", which the pre-flight treats as leave to skip) | `aCancelledMasOutdatedStillAnswers` (M2). Old code could deadlock on >64 KB stdout; new cannot |
| 18 | `AppStoreAXInstaller.ensureAppStoreRunning` | `open -g -b` | none | **inh / inh** → discard / discard | inh | not read, as before | terminateChild — on cancel the install throws out; nothing reads the result | **untested** |
| 19 | `AppStoreAXInstaller.navigate` | `open -g <url>` | none | **inh / inh** → discard / discard | inh | non-zero → `appStoreUnavailable` | terminateChild — `try navigate` throws the cancel out of the install; the two `try? navigate` callers are best-effort and sleep cancellably right after | **untested**; was a sync `waitUntilExit` on the actor (invisible to the offpool check) |
| 20 | `GitHubToken.run` | `gh auth token` | none | pipe / null → captured / discard | inh | nil on non-zero | **runToCompletion** (r1: `ChangelogService` caches the answer ten minutes) | `aCancelledGhTokenReadStillAnswers` (G1); callers race a 2 s deadline, as before |
| 21 | `GitHubToken.runCapturingStderr` | `gh auth status` | none | null / pipe → discard / captured | inh | (stderr, ok) | **runToCompletion** (r1: killed reads as "not logged in") | `aCancelledGhStatusReadStillAnswers` (G2) |
| 22 | `VendorProbeSource.extractZipEntry` | `unzip -p` | none | pipe / null → captured / discard | inh | "unzip exited N extracting '…'" | **runToCompletion** (r1: a killed `unzip` was filed as a recipe fault) | `aCancelledZipEntryReadStillAnswers` (Z3); `VendorProbeZipEntryTests` |
| 23 | `GitHub.run` (CLI) | `/usr/bin/env gh …` | none | two pipes, semaphore drain → both captured | inh / none / bytes or inh | "gh X failed (N): stderr" | runToCompletion (an issue created but unrecorded would be refiled) | **untested** (needs `gh` on PATH) |
| 24 | `Triage.shell` (CLI) | `/usr/bin/env opencode …` | **TERM at timeout kept, KILL +5 s added**; old code returned right after TERM | stdout drained on a queue, stderr read after exit → both concurrent | inh / **cwd kept** / inh | "timed out after Ns", "opencode exited N: stderr" | terminateChild — `ask` turns the throw into a failure `run` logs and skips; no suggestion is written | `TriageShellTests` (3) |
| 25 | `Doctor.exitsZero` (CLI) | `launchctl print` | none | null / null → discard / discard | inh | Bool | **runToCompletion** (r1: killed prints "not registered") | `aCancelledStatusReadStillAnswers` (DR1) |
| 26 | `AppListModel.runningBuildVersions` (App) | `lsappinfo list` | 5 / 8 s kept | pipe / null → captured / discard | inh | `[:]` on failure | **runToCompletion** (a cancelled restart recheck keeps using the result — see review) | **untested** (App test target does not compile `AppListModel`) |
| 27 | `PrivilegedHelperClient.restartDaemon` (App) | `osascript` kickstart | none | **inh** / pipe → discard / captured | inh | failure string | runToCompletion | **untested** (admin panel) |
| 28 | `LoginShellEnvironment.resolveHomebrewVariables` (#595) | `$SHELL -l -i -c 'env -0'` | own 10 s timeout kept, **counted from the shell's launch** (r1): SIGSTOP shell, SIGKILL descendants + shell (needs the pid → new `ChildProcess` `onLaunch`) | temp file / null → captured / discard (rc children holding the pipe no longer matter) | env given or inh / none / **/dev/null → empty bytes** | nil on failure/timeout | runToCompletion | #595's `readsVariablesExportedByAFixtureZshrc` (its background `sleep` now killed), `aShellThatHangsIsUnknownAndKilled` (60 s watchdog), `theTimeoutCountsFromTheShellsLaunch` (U4) |
| 29 | `HomebrewConfig.read` (#595) | `brew config` | none | pipe / null → captured / discard | `*_proxy` + shell vars | nil on failure | runToCompletion | **untested** (needs brew) |
| — | `DiagnosticsSettingsPage.relaunch` (App) | `open -n` | — | — | — | — | **excluded**: fire-and-forget, app terminates; `offpool-lint:allow` with reason | check_offpool R5 |

Ripple (sync → async, no new launch): `ArchiveExtractor.extractApp`, `DeltaApplier.reconstruct`, `InPlaceSwap.replace/rotateContents`, `BackupStore.save/restore/sweepStagingLeftovers`, `InputMethodDataBackup.save/restore`, `PackageInstaller` gate helpers + `declaredDestinations`, `GitHubToken.resolve/cliStatus`, CLI `Settings.load`, `GitHub.*`, `Triage.run`, `Reconcile.run`. Hops removed where they only parked a subprocess wait (Sparkle/Vendor installers, `InstallCoordinator.backUp`, CLI/App rollback, `ChangelogService`, `AppListModel.resolveGitHubToken`, `main.swift` triage/reconcile). Hops kept or added around real blocking work: `SecStaticCode…` (package gate's `teamIdentifier`, restore's integrity check), `BackupManifest.unreadableFiles/unexpectedOmissions/compute`, `PackageInstaller.contentSeal`. Synchronous disk work that used to ride inside the removed hops now has its own hops (review finding C): `InPlaceSwap`'s unprivileged exchange and both halves of `rotateContents`, `BackupStore.save`'s generation replace and staging sweep, `InputMethodDataBackup`'s exchanges, #592's quarantine walk. `InstallLock` coverage is unchanged (no claim/release moved). `SwapOutcome` is unchanged and still returned; the installers keep #590's explicit `_ =`.

## Mutation table

Every row was applied to the file, the named test run alone, seen red, and reverted. Round-1 rows (after O2) were run on round-1 code (script-driven; a compile error is reported as such — none were). Re-run after the review fixes (post-#592/#590/#594 rebase): H1–H16, H18, I1–I11, B1, M1, Z1–Z2, S1–S2, R1–R5, O1–O3. L1–L2 and U1–U3 were run on the last two commits. C1–C3 were run before the rebases; `Triage.swift` did not change in them.

| ID | Mutation | Test | Result |
|---|---|---|---|
| H1 | `Sink.drainNow`: read stdout to EOF before starting the stderr task | `bothStreamsPastAPipeBufferDrainTogether` | red (timedOut, sizes short) |
| H2 | `.mergeIntoOutput` routed to the separate-streams spelling | `mergedErrorArrivesInOrderOnStandardOutput` | red |
| H3 | `Sink` keeps stdout regardless of policy | `discardedStreamsCollectNothing` | red |
| H4 | stdin bytes spelling passes `.currentStandardInput` | `standardInputBytesReachTheChild` | red |
| H5 | per-buffer `onOutputChunk` call dropped | `outputChunksArriveWhileTheChildRuns` | red |
| H6 | `environment` → `.inherit.updating(…)` | `anEnvironmentReplacesTheInheritedOne` | red |
| H7 | nil environment → `.custom([:])` | `noEnvironmentInheritsOurs` | red |
| H8 | `workingDirectory` dropped | `theWorkingDirectoryIsTheChilds` | red |
| H9 | `.signaled` reported as `uncaughtSignal: false` | `exitCodesAndSignalsKeepProcessConventions` | red |
| H10 | launch error swallowed into status -1 | `aMissingExecutableThrows` | red |
| H11 | deadline fires but does not cancel the launch | `aChildIgnoringSIGTERMIsKilled` | red (exit 7) |
| H12 | `timedOut` never raised | `aChildIgnoringSIGTERMIsKilled` | red |
| H13 | ladder `[]` (straight to SIGKILL) | `aChildHonouringSIGTERMExitsOnItsOwnTerms` | red |
| H14 | `timedOut: { true }` | `aChildInsideItsDeadlineIsNotTimedOut` | red |
| H15 | `.runToCompletion` not shielded | `runToCompletionOutlivesTheCallersCancellation` | red |
| H16 | `.terminateChild` shielded | `terminateChildTearsTheChildDownOnCancellation` | red |
| H17 | drop the pre-spawn `checkCancellation` | first test for it: **green** (race), deleted. Round 1: `anAlreadyCancelledTaskNeverSpawns` (counts `onLaunch` calls) | red (round 1) |
| H18 | drain read inline (not in an unstructured task) | `aChildHonouringSIGTERMExitsOnItsOwnTerms` | red (status 13) |
| I1 | rotation `chmod` → `.terminateChild` | `aCancelledRotationStillCarriesTheGroupWriteBit` | red |
| I2 | `stripQuarantine` → `.terminateChild` | `aCancelledSwapStillStripsQuarantine` | red |
| I3 | `ArchiveExtractor.run` → `.terminateChild` | `aCancelledExtractionStillUnpacksTheApp` | red |
| I4 | extractor `err` slot filled from stdout | `aFailingToolReportsItsStatusAndStderr` | red |
| I5 | delta message takes first stderr line | `aFailedPatchReportsTheToolsStatusAndLastStderrLine` | red |
| I6 | `DeltaApplier.apply` → `.terminateChild` | `aCancelledPatchStillFinishes` | red |
| I7 | `BackupStore.runDitto` → `.terminateChild` | `aCancelledBackupAndRollbackStillComplete` | red |
| I8 | `PackageInstaller.runCapturing` → `.terminateChild` | `aCancelledPackageGateStillReadsTheDeclaredDestinations` | red |
| I9 | `runCapturing` drops `workingDirectory` | same | red |
| I10 | restore reads the recorded manifest inside the hop (task-local lost) | `tamperingWithTheStoredCopyIsRefused`, `anExtraFileInTheStoredCopyIsRefused` | red — this was the actual first draft |
| I11 | `installedLeaves` reads serialized | `installedLeavesReadsGenuinelyOverlap` (async gate) | red |
| B1 | `clearUserImmutableFlags` never runs `chflags` | `anImmutableFileInTheSourceDoesNotFreezeTheStoredCopy` | red |
| M1 | `InputMethodDataBackup.copyTree` success inverted | `aSnapshotRestoresTheDataTheUpdateChanged` | red |
| Z1 | unzip plist read from stderr | `theEntryIsReadFromUnzipsStandardOutput` | red |
| Z2 | unzip message built from a non-status value | `aMissingEntryReportsUnzipsExitStatus` | red |
| C1 | `Triage.shell` ignores `timedOut` | `aCallPastItsTimeoutIsReportedAsATimeout` | red |
| C2 | `Triage.shell` message from stdout | `aNonZeroExitCarriesItsStatusAndStderr` | red |
| C3 | `Triage.shell` drops cwd | `itRunsInTheGivenDirectory` | red |
| R1 | real tree: `Process()`+`waitUntilExit()` in an async func (`VendorProbeSource`) | `check_offpool.py` | red (launch + blocking) |
| R2 | real tree: stale `offpool-lint:allow` on `MASInstaller.runOutdated` | `check_offpool.py` | red (dead exemption) |
| R3 | real tree: `Process()` in a synchronous helper (`ArchiveExtractor`) | `check_offpool.py` | red (launch) |
| R4 | real tree: `SecStaticCodeCheckValidity` in an async func | `check_offpool.py` | red |
| R5 | real tree: remove `relaunch`'s exemption | `check_offpool.py` | red |
| S1 | `replace` unprivileged route stops setting `cleanupFailure` | #590's `aCleanupFailureReasonReachesTheInstallLog` (rewritten around the async `replace`) | red |
| S2 | #592 quarantine walk skipped after `xattr` fails | `aPartialStripReportsWhatIsStillQuarantined` | red |
| L1 | `StreamedLines.run` drops the unterminated tail | `aLineSplitAcrossWritesArrivesWhole` | red |
| L2 | `StreamedLines.run` discards stderr | new `standardErrorArrivesInTheSameStream` | red (was **green** before that test existed) |
| U1 | `onLaunch` never reports the pid | `aShellThatHangsIsUnknownAndKilled` | red (round 1: the timer never starts, watchdog fires at 60 s) |
| U2 | kill only the shell | same | red (survivors) |
| U3 | login-shell race with a zero timeout | `readsVariablesExportedByAFixtureZshrc` | red |
| O3 | script: launch regex back to literal `Process()` | `test_every_launch_spelling_is_refused` | red |
| O1 | script: judge launches by scope like waits | `test_check_offpool.py` | red (3 tests) |
| O2 | script: launch regex without look-behind | `test_check_offpool.py` | red (`lookalike`) |
| B2r | `BrewFormulaService` executor → `.terminateChild` | `aCancelledRefreshStillReadsRealData` | red |
| B3r | `brewInfoOffActor` → `.terminateChild` | `aCancelledBrewInfoStillAnswers` | red |
| G1 | `GitHubToken.run` → `.terminateChild` | `aCancelledGhTokenReadStillAnswers` | red |
| G2 | `GitHubToken.runCapturingStderr` → `.terminateChild` | `aCancelledGhStatusReadStillAnswers` | red |
| M2 | `MASInstaller.outdatedAdamIDs` → `.terminateChild` | `aCancelledMasOutdatedStillAnswers` | red |
| Z3 | `VendorProbeSource.extractZipEntry` → `.terminateChild` | `aCancelledZipEntryReadStillAnswers` | red |
| DR1 | `Doctor.exitsZero` → `.terminateChild` | `aCancelledStatusReadStillAnswers` | red |
| U4 | login shell: drop `await launched.waitForLaunch()` (clock from the call) | `theTimeoutCountsFromTheShellsLaunch` | red |
| P1 | drop `options.processGroupID = 0` | `theChildLeadsItsOwnProcessGroup` | red |
| E1 | spawn with the unfiltered environment | `invalidEnvironmentEntriesAreDroppedNotFatal` | red |
| E2 | leading-digit rule dropped | `theEnvironmentFilterNamesWhatItDrops` | red |
| S1s | closed stdin → `.inherit` | `aClosedStandardInputGivesTheChildDevNull` (removed in round 2 with the function it tested) | red |
| F1 | fill: skip the `dup2` | `aClosedDescriptorIsPointedAtDevNull` | red |
| F2 | fill: drop both "is it closed" checks | `aClosedDescriptorIsPointedAtDevNull` | red |
| F3 | remove the fill call from the `.devNull` branch (round 1 design, replaced in round 2) | probe under `<&-` | child shows `Bad file descriptor` |
| D1 | `ChildProcess` deadline: drop `try await launched.wait()` | `theDeadlineClockStartsAtLaunch` | red |
| D2 | `LaunchSignal.wait` ignores cancellation | `aFailedLaunchWithADeadlineStillReturns` | red |
| B1s | `BackupStore.save`: ditto-failed branch keeps its staging dir | `failedRebackupKeepsPriorBackup` | red |
| B2s | `BackupStore.save`: legacy-key removal dropped | `saveDropsOrphanedLegacyBundleIDBackup` | red |
| B3s | `BackupStore.restore`: scratch removal after the call dropped | `aRestoreLeavesNoScratchCopyBehind` | red |
| I1s | `InputMethodDataBackup.save`: staging removal dropped | `aSnapshotThatStoresNothingLeavesNoStagingBehind` | red |
| I2s | `InputMethodDataBackup.restore`: scratch removal dropped | `aRestoreLeavesNoScratchBehind` | red |
| P1o | real tree: rename `SecStaticCodeCheckValidity` in `BLOCKING` | `check_offpool.py` | red (the old total floor: 8 ≥ 5, green) |
| P2o | no per-spelling check | `test_one_spelling_that_matches_nothing_fails` | red |
| P3o | `RETIRED` ignored | `test_a_retired_spelling_may_match_nothing` | red |
| P4o | no revival check | `test_a_retired_spelling_that_matches_again_fails` | red |
| P5o | `RETIRED` keys not checked against `BLOCKING` | `test_a_retired_spelling_not_in_blocking_fails` | red |
| D1b | backdeploy: `/usr/lib/swift` satisfies shims | `test_an_unembedded_span_shim_fails`; the real Debug `duo` then passes | red / (gate green under mutation) |
| D2b | backdeploy: absolute run paths count | `test_an_absolute_run_path_does_not_count` | red (a first fixture was green: the containment check also caught it; fixture changed) |
| D3b | backdeploy: never accept an embedded copy | `test_a_shim_embedded_in_the_bundle_passes` | red |
| D4b | backdeploy: a lone binary's directory counts as shipped | `test_a_shim_beside_a_lone_binary_is_not_embedded` | red |
| D5b | backdeploy: `/usr/lib/swift` run path not honoured | `test_the_os_runtime_by_run_path_passes` | red |
| D6b | backdeploy: no "no Mach-O" floor | `test_a_path_without_mach_o_fails` | red |
| D0 | (unmutated) gate on `CLI/.build/out/Products/Debug/duo` | `check_swift_backdeploy.py` | exit 1: `@rpath/libswiftCompatibilitySpan.dylib … not embedded` |
| G1 | r2: `load_paths` back to `otool -L` parsing | `test_a_correctly_embedded_real_shim_passes` (real Mach-O) | red |
| G2 | r2: `@executable_path` = outer `Contents/MacOS` for every Mach-O | `test_a_nested_executable_uses_its_own_executable_path` | red |
| G3 | r2: `LC_ID_DYLIB` counted as a dependency | `test_a_correctly_embedded_real_shim_passes` | red |
| — | r2: D1b–D6b re-run on the new parser | as above | all red |
| W1 | r2: `fetchThenSwap` never removes the work dir | `FetchThenSwapCleanupTests` (3) | red ×3 |
| W2 | r2: removal only after a successful apply | `FetchThenSwapCleanupTests` | red (throw + cancel) |
| T1 | r2: runner does not cancel the timer | `aShellThatCannotStartLeavesNoTimerBehind` | red (not released in 60 s) |
| T2 | r2: `waitForLaunch` cancellation handler does not resume | same | red |
| U4r2 | r2: clock from the call, with the widened 5 s / 8 s margins | `theTimeoutCountsFromTheShellsLaunch` | red |
| U1r2 | r2: `onLaunch` never reports the pid (after the timer change) | `aShellThatHangsIsUnknownAndKilled` | red |
| F3r2 | r2: probe without `ensureStandardInputIsOpen()` vs with it, under `<&-` | probe | no fd 0 vs `/dev/null` |

## Non-blocking evidence (measured outside the suite; no wall-clock asserts in tests)

A release-built probe (a scratch package depending on this `DuoUpdaterCore`) starts 8 concurrent `sleep 1` children three ways. Each mode ran in its own process, 3 runs each, on a 14-core Apple Silicon Mac with Xcode 27 RC:

```
LIBDISPATCH_COOPERATIVE_POOL_STRICT=1
parked  (Process + waitUntilExit in a task)    wall 8.04s / 8.05s / 8.07s   peak threads 5 / 5 / 5
hopped  (same wait inside offCooperativePool)  wall 1.01s / 1.01s / 1.01s   peak threads 18 / 17 / 19
awaited (ChildProcess.run)                     wall 1.01s / 1.01s / 1.01s   peak threads 8 / 7 / 7

LIBDISPATCH_COOPERATIVE_POOL_STRICT unset
parked   wall 1.01s ×3   peak threads 17 / 17 / 16
hopped   wall 1.01s ×3   peak threads 16 / 17 / 17
awaited  wall 1.01s ×3   peak threads 13 / 13 / 13
```

## Verification (head ca4b4aa5, commands and trimmed output)

```
$ scripts/run-with-hang-report.sh 1800 make test                  # Xcode 27 RC / Swift 6.4, rc=0
Ran 45 OK · Ran 7 OK · Ran 11 OK · Ran 32 (test_check_offpool) OK · Ran 9 (test_check_swift_backdeploy) OK · Ran 4 OK · Ran 5 OK
━ Test run with 2982 tests in 256 suites passed after 20.100 seconds with 1 known issue.   # DuoUpdaterCore
✔ Test run with 334 tests in 42 suites passed after 6.317 seconds.                         # CLI
✓ localizable keys agree — 565 in the catalog, all reachable
✓ version comparisons discriminate — 269 files …   Ran 11 OK
✓ app audits consistent …  ✓ engine-notes pointers resolve — 571 Swift files …  ✓ skill docs consistent …
✓ no machine-population claims in code comments — 571 files
✓ no blocking calls left on the cooperative pool — 9 call(s) in 269 files

# Swift 6.3.3 / Xcode 26.6, same head
$ DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer swift test --package-path DuoUpdaterCore --scratch-path $T/build633
━ Test run with 2982 tests in 256 suites passed after 20.475 seconds with 1 known issue.
$ DEVELOPER_DIR=… swift test --package-path CLI --scratch-path $T/build633-cli
✔ Test run with 334 tests in 42 suites passed after 6.251 seconds.
$ DEVELOPER_DIR=… APP_TESTS_DD=$T/dd-apptests-633 scripts/app-tests.sh
$ DEVELOPER_DIR=… python3 scripts/test_check_swift_backdeploy.py          # the real-shim test against 26.6's toolchain
Ran 9 tests … OK
```

**Linkage (round 1, item 9).** Release builds, unsigned (`CODE_SIGNING_ALLOWED=NO`), into scratch derived data, at bd577168. The later commit changes only `ChildProcess`'s stdin branch and its tests, not dependencies. Swift libraries and `@rpath` entries from `otool -L`:

```
duo-cli, Xcode 27.0 (27A5237l)
  /usr/lib/swift/: libswiftCompression(weak) Core CoreFoundation CoreImage(weak) Darwin Dispatch IOKit(weak) Metal(weak)
                   OSLog(weak) ObjectiveC QuartzCore(weak) Spatial(weak) System UniformTypeIdentifiers(weak) XPC(weak)
                   _Concurrency os simd(weak)                       — no @rpath entry
duo-cli, Xcode 26.6 (17F113)          identical list                — no @rpath entry
DuoUpdater (app executable), Xcode 27.0
  @rpath/Sparkle.framework/Versions/B/Sparkle
  /usr/lib/swift/: Compression Core CoreAudio(weak) CoreFoundation CoreImage Darwin Dispatch IOKit Metal OSLog ObjectiveC
                   Observation QuartzCore Spatial System UniformTypeIdentifiers XPC _Concurrency os simd CoreGraphics
DuoUpdater (app executable), Xcode 26.6   same without CoreAudio — no other @rpath entry
DuoUpdaterHelper, both Xcodes   /usr/lib/swift/: Core CoreFoundation Darwin Dispatch IOKit ObjectiveC XPC
BinaryDelta (Sparkle 2.9.6 artifact)  @rpath/libswiftCore, CoreFoundation, Darwin, Dispatch, Foundation, IOKit, ObjectiveC, XPC
                                      via LC_RPATH /usr/lib/swift (minos 11.0)

$ scripts/check_swift_backdeploy.py <product>
duo-cli (27)           1 Mach-O file(s): no Swift runtime library missing from macOS 14     exit 0
duo-cli (26.6)         1 Mach-O file(s): …                                                exit 0
DuoUpdater.app (27)    9 Mach-O file(s): …                                                exit 0
DuoUpdater.app (26.6)  9 Mach-O file(s): …                                                exit 0
CLI/.build/out/Products/Debug/duo (swift build)
  ✗ … links @rpath/libswiftCompatibilitySpan.dylib — a Swift back-deployment shim, absent from macOS before 26,
    and not embedded in what ships.                                                       exit 1
```

**Download gate:** `DUO_DOWNLOAD_GATE=1 … --filter vendorDownloadPassesSignatureGate` passed locally once, on the tree before the review fixes and the last two rebases:
- ChatWise zip: sha512 ✓, `ditto -x -k`, Team ID GJE9R5VE87 matched.
- VLC dmg: `hdiutil attach`/`ditto`/`detach`, Team ID 75GAHG3SZQ matched.
- `hdiutil info` showed no mount left behind.
- A second local run on the final head was stopped: the local proxy was downloading at 40–50 KB/s. **The final head's download gate is this PR's CI run** (`ci.yml` sets `DUO_DOWNLOAD_GATE=1`).

Flake check: the new suites ran ×8, plus ×2 at pool width 1, all green (before the review fixes).

## 未验证 (unverified)
- **Old macOS:** no binary from this branch has been launched on macOS 14 or 15. The linkage evidence and the new gate are the only support for "no `libswiftCompatibilitySpan` problem" there.
- **The back-deployment gate's wiring** into `build-cli.sh`, `install.sh` and `notarize.sh` was not executed end to end (`make cli`/`make install`/`make notarize` are off limits here). The script was run on the four Release products and the Debug `duo`, and its own tests run in `make test`.
- **Spawn latency through swift-subprocess's single spawn thread:** not measured (round 1, item 4).
- **`make test` from an interactive terminal: check for T-state zsh.** With `processGroupID = 0`, a `zsh -l -i` child of a process that has a controlling terminal might stop itself with SIGTTOU. The reviewer notes main had the same exposure (a `Process` child also led its own group). Not tried: no interactive terminal here.
- **`duo`'s `ensureStandardInputIsOpen()` call in `main.swift`** was not exercised by running `duo` under `<&-`. The function was, at process start, in a probe. `main.swift` has no test target.
- **The gate's nested-library case:** a Swift library loaded by a nested executable is resolved against the main executable's `@executable_path`. Documented; nothing ships in that shape today.
- **The off-pool deletions (round 1, item 5)** are not observable to a test; the tests pin that each removal still happens on every exit, not which thread does it.
- **Admin panel:** the privileged `osascript` swap and the helper kickstart were not run (no admin prompts on this machine). Neither was the `elevationPanel` serialisation (`InstallPermits` as an async mutex).
- **Host-dependent sites** (they need Homebrew, `mas`, `gh` or the App Store; mutation-free):
  - `brew` itself through `StreamedLines.run` (the runner is tested with `/bin/sh`)
  - `brew config`, `brew info`
  - `mas outdated`, including the new 25 s SIGKILL
  - `gh auth token/status`, `GitHub.run`
  - `open -g` App Store navigation
  - `launchctl print`
  - `lsappinfo` (App target)
- **No fixture:** `PackageInstaller.run`'s dmg-wrapped pkg route; the `pkgutil --check-signature` launch (its parser is tested).
- **The final head's download gate** (CI).
- **Whether swift-subprocess's final non-blocking drain ever loses bytes** when a grandchild holds the pipe and the pending data exceeds one read. Read from the source, and `outputWrittenJustBeforeExitIsNotLost` (~209 KB before exit) passes; not stress-tested with grandchildren.
- **Intermediate commits:** commit 2 (Core) does not build the CLI or app; commits 3–4 restore them. Only the head was built and tested.
- **CI:** not waited for.

## Risks
- **Blast radius:** every install, backup, rollback, swap and pkg hand-over path changed shape (sync → async plus hops), in 66 files.
  - Ordering is lexical and preserved.
  - `SecStaticCode*` and the heavy disk segments are hopped.
  - Hops lose task-locals: one real regression of this kind (restore's integrity check) was caught by existing tests and fixed. The reviewer found no others.
- **Timeouts:** counted from launch, as before. A spawn stuck behind another on swift-subprocess's single spawn thread is not bounded by any deadline until it launches.
- **fd 0:** when our process starts without one, `ChildProcess` points it at `/dev/null` for the rest of the process's life.
- **Deliberate ladder additions:** SIGKILL for `mas outdated` (25 s) and for `Triage.shell` (timeout + 5 s). Before, a child that ignored SIGTERM kept the caller waiting (mas) or was left running (triage).
- **Output after exit:** output a grandchild writes after the direct child exits is no longer collected. Not checked whether any tool here relies on that (未验证).
- **stdout that used to be inherited** (`open`, `chmod`, `osascript`) is now drained and discarded, so `open`'s error text no longer reaches a `duo` terminal.
- **Dependency surface:** two new packages. An `internal import Subprocess` keeps them out of `DuoUpdaterCore`'s public API.


jizhi0v0:
## Runtime verification: real updates and a rollback (macOS 27.0, 26A428)

Built from `ca4b4aa5` and installed on the reviewer's machine with `make install` and `make cli`. The back-deploy gate ran in both builds:

```
1 Mach-O file(s) under …/Release/duo-cli: no Swift runtime library missing from macOS 14
9 Mach-O file(s) under …/Release/DuoUpdater.app: no Swift runtime library missing from macOS 14
```

| | Mole (Sparkle, `/Applications`) | Chatbox (Vendor, `~/Applications`) |
|---|---|---|
| `duo install <app> --yes` | 1.13.0 (196) → **1.14.0 (259)**, 20.7 MB downloaded in 2 s | 1.23.1 → **1.23.2**, 161 MB downloaded in 9 s, `backed up 1.23.1` |
| `codesign --verify --deep --strict` | OK | OK |
| `spctl -a -vv` | accepted, Notarized Developer ID | accepted, Notarized Developer ID |
| quarantine xattr | present before, stripped after | none after |
| bundle inode | 237420991 → 241367085 | 222975870 → 241372100 |
| leftovers (`.duoupdater-staged/old`, `hdiutil info` mounts) | none | none |
| log | `swap start … elevated=false` → `swap done` → `applied=true` | same |
| `duo backups restore <app> --yes` | not possible: `KeepBackups` is off on this machine, so no rollback point exists (by design, not a regression) | 1.23.2 → **1.23.1**; codesign OK, notarized; inode → 241373804; size back to 441 M; `swap start/done elevated=false` |

`KeepBackups` was switched on only for the Chatbox run and then restored to off.

No error-level log entries came from the installed `DuoUpdater` process during the window. The errors in that window came from `swiftpm-testing-helper` and probe fixtures run by other sessions.

**Not exercised:**
- the privileged (`osascript`) swap route;
- an input-method `Contents` rotation;
- the pkg installer route;
- a delta patch;
- Homebrew `brew upgrade`.

jizhi0v0:
## Post-merge runtime verification: the remaining routes (macOS 27.0, 26A428)

These runs cover the routes the earlier comment listed as not exercised.

**Which build ran:**
- **CLI runs:** `~/.local/libexec/duo` built from this PR, UUID `686F61C3…`. Every `duo` log line below carries that `processImageUUID`.
- **App runs:** `/Applications/DuoUpdater.app` built from `aa22b72f` (main, tree-identical to this PR's head), UUID `5BD0FD70…`, installed after checking with the other session using the app.

| Route | App | Result |
|---|---|---|
| Sparkle **delta** | Shotbase 1.4.0 (13) → 1.4.1 (14) | log `delta route: Shotbase build 13 → 14, patch 66473042 B instead of 154656942 B`; codesign OK, notarized, new inode |
| **Privileged `osascript` swap** | Waku 0.1.17 → 0.1.19 | bundle made `root:wheel` and unwritable first; log `swap start … elevated=true` → `swap done` ~45 s later (admin panel answered) → `applied=true`; owner kept `root:wheel`; codesign OK, notarized; no `.duoupdater-old` left |
| **pkg installer** | wechatwebdevtools 2.02.2608060 → 2.02.2608070 | 254 MB pkg; gates (`pkgutil` team check, `xar`/`lsbom` destinations) passed, since `Installer.app` was opened and the gates throw before that; user completed Installer; `duo check` now up to date; codesign OK, notarized |
| **Input-method `Contents` rotation** | DoubaoIme 0.9.7 → 1.0.0 | outer bundle inode unchanged (root-owned), `Contents` inode changed; `swap start … elevated=false` → `swap done`; `duo restart DoubaoIme` relaunched the IME (new pid); user typed with it afterwards without problems |
| **App Store, AX route** (app) | AndroMeld 1.8.0 → 1.9.0 | `appstore-ax: … pressed Update` → progress → `install complete — bundle version changed on disk` |
| **App Store, root helper + `mas`** (app, strategy `full`) | WhatsApp 26.35.72 → 26.35.74 | helper listener activated, accepted DuoUpdater's peer connection, `mas install 310633997 --force` via `launchctl asuser`; `install ok in 33s (attempt 1)`; codesign OK, receipt present |
| **`brew upgrade`** (app, workbench "Upgrade All") | local fixture cask `zzduo-probe` 1.0 → 1.1 | process sampler under DuoUpdater's pid: `brew upgrade --formula`, then `brew upgrade --cask zzduo-probe`, then the refresh reads; cask landed at 1.1 |

**`brew upgrade` failure path:** the per-row `oven-sh/bun/bun` upgrade ran (sampler: `brew.sh upgrade --formula oven-sh/bun/bun`, auto-update, ruby for about 6 s) and exited 1. The terminal reproduction shows `Error: Your Command Line Tools are too outdated` (no bottle, and the CLT don't match Xcode 27), so this is a host brew limitation. The UI only showed brew's last line; that predates this PR and is tracked separately.

**Cleanup:**
- fixture cask, tap and trust entries removed;
- bun unpinned;
- Waku ownership restored to `bobby:admin`;
- `AppStoreUpdateStrategy` restored to `incremental`;
- `KeepBackups` still off.

PR #606, opened 2026-09-13
recipes: one file per app family (step 1 of 3)

jizhi0v0:
## What

Step 1 of the recipe refactor: one file per app family, the registries derived from one index.

- New `DuoUpdaterCore/Sources/DuoUpdaterCore/Recipes/`: **224 family files**, `AppRecipeSet.swift` (the value type), and `AppRecipeIndex.swift` (the one hand-maintained list, sorted by slug, case-insensitive).
- The ten per-app tables keep their names, public API and files. Only their literals became derivations over `AppRecipeIndex.all`:
  - arrays via `flatMap`: `VendorProbeRegistry.recipes`, `ChangelogRecipeRegistry.recipes`, `GitHubReleaseRegistry.rules`, `MacAppStoreProbeRegistry.cases`
  - dictionaries via `AppRecipeIndex.merged(_:into:)`, which traps on a duplicate key and names both families: `ChannelProofRegistry.proofs` / `.githubProofs` / `.bindingProofs` (kept as three separate maps, per the collision note on `githubProofs`), `SparkleFeedCatalog.feeds` / `.supersededFeeds`, `ChangelogCatalog.pages`
- Entry code moved verbatim; the only code edits are the helper qualifications listed below. Comments moved **by subject**, not by position (review round 1): each comment block sits in the file of the app it is about. The only comment changes are the listed MARK deletions and pointer rewrites.

Commits:
- round 0: engine `99ac7af9`, tests `78ce66ec`, docs+script `6949e731`
- round 1: A `8c150ac3`, C `24f5c37c`, B `cc6fe9ef`, D `69bd8094`, E `c4f50980`, F `9b3c7805`
- round 2: group rationale `dda5a62e`
- round 3: tests `6b0507bd`, docs `7cad5045`

## Layout decisions

[...]

PR #607, opened 2026-09-13
recipes step 2a: history-migration convention, guard, two pilots, carried items

jizhi0v0:
Step 2a of the recipe refactor: the convention and guard that step-2 batches follow, two fully migrated pilot families, and the three items left over from step 1 (#606).

## Commits

1. `check_app_audits: guard recipe history pointers`: the guard
2. `app-audits: write down where recipe comment history goes`: the convention section in `docs/app-audits/README.md`, plus the new index section 「仅迁出历史（未审计）」
3. `recipes: move Windscribe and CapCut comment history into their audits`: the two pilots
4. `Resolve three items carried from the recipe split`: Alcove, WorkBuddy, VSCodium

## 1. Convention

Added `docs/app-audits/README.md` §「从 recipe 注释迁出的历史」, placed above the index. It points at the `docs/engine-notes/README.md` checklist instead of copying it, and covers:
- what stays (a) and what moves (b)
- (c): anything no longer true of the current code gets fixed or deleted, never moved
- classification works sentence by sentence with no rewording. A sentence mixing contract and measurement stays whole. Only punctuation may change where a paragraph is split.
- the fixed heading `## 历史与实测`, goes at the end of the file, one doc per family (even if another app in the family has its own audit)
- a source line and a `转引自 recipe 注释，未复测。` label for each group. The text is verbatim: the `// ` prefix is stripped, the original language and line breaks are kept, and indented excerpts are fenced.
- machine-state sentences are reworded per copy, with no exemption
- the pointer is exactly `// History: docs/app-audits/<family>.md#历史与实测`
- the history-only form for a family with no audit, listed in the index section 「仅迁出历史（未审计）」 and never checked off

## 2. Guard (`scripts/check_app_audits.py`)

Four new rules, all in `check_history_pointers`:
- every `docs/app-audits/<f>.md#历史与实测` in a Swift comment under `DuoUpdaterCore|App|CLI` `Sources|Tests` must name a file that is **tracked by git** (`git ls-files`)
- that file must contain a line reading exactly `## 历史与实测`
- inside `Recipes/`, `<f>` must be the family's own file name
- every tracked audit with the heading must be pointed at (otherwise it's orphan history)

Vacuity check: fails if fewer than 100 Swift files or 100 `Recipes/` files are scanned. No exemption mechanism.

### Mutation results (each introduced once, then restored)

```
### M1 pointer to a missing file
✗ app audits disagree with the rules that keep them publishable:
    DuoUpdaterCore/Tests/DuoUpdaterCoreTests/ZZHistoryPointerMutation.swift:1: history pointer names `docs/app-audits/zz-no-such-app.md`, which does not exist
exit=1
### M2 pointer to an untracked file
✗ app audits disagree with the rules that keep them publishable:
    docs/app-audits/zz-untracked.md: not linked from README.md
    DuoUpdaterCore/Tests/DuoUpdaterCoreTests/ZZHistoryPointerMutation.swift:1: history pointer names `docs/app-audits/zz-untracked.md`, which exists on disk but is not tracked by git (docs/* is gitignored by default — check the carve-out in .gitignore)
exit=1
### M3 pointer to a tracked file without the heading
✗ app audits disagree with the rules that keep them publishable:
    DuoUpdaterCore/Tests/DuoUpdaterCoreTests/ZZHistoryPointerMutation.swift:1: history pointer names `docs/app-audits/com-vscodium.md`, which has no line reading exactly `## 历史与实测`
exit=1
### M4 pointer copied into the wrong family
✗ app audits disagree with the rules that keep them publishable:
    DuoUpdaterCore/Sources/DuoUpdaterCore/Recipes/com-lemon-lvoverseas.swift:10: history pointer names `com-windscribe-client.md` inside the `com-lemon-lvoverseas` family — a family's history goes to `docs/app-audits/com-lemon-lvoverseas.md`
exit=1
### M5 orphan history heading (pointer removed)
✗ app audits disagree with the rules that keep them publishable:
    docs/app-audits/com-lemon-lvoverseas.md: has `## 历史与实测` but no Swift comment points at it (expected `// History: docs/app-audits/com-lemon-lvoverseas.md#历史与实测`)
exit=1
### restored
✓ app audits consistent — 124 docs, all indexed, no machine inventory, no pointers into untracked evidence, history pointers resolve (799 Swift files, 226 families)
exit=0
### vacuity: RECIPES path pointed at a directory that does not exist
✗ history pointers: only 799 Swift files / 0 recipe family files scanned — too few to be a real run (did DuoUpdaterCore/Sources/DuoUpdaterCore/RecipesMoved move?)
exit=1
```

M2 also shows the existing index rule firing on the untracked file. That is expected: the file is not in the index either.

## 3. Pilots

**How to verify the accounting.** Run `account.py` (script at the end) against the Swift file before and after, with base rev `7b41abf5` for the audit. It checks four things:
1. comment line counts
2. every moved paragraph is a whitespace-normalised verbatim substring of the **original** comments
3. after cutting the moved paragraphs out of the original comments, a word-level diff against the new comments shows exactly the pointer, the (c) fixes and seam punctuation
4. a line reconciliation

### Windscribe (`Recipes/com-windscribe-client.swift`)

- **Comment lines:** 289 → 237 (−52)
- **Audit:** +81 lines = 56 moved body lines + 25 structure (heading, intro, 3 source lines, 3 labels, blank lines)
- **Reconciliation:** 52 removed = 49 whole lines moved + 3 blank separators dropped + 29 original lines rewritten/split − 29 new lines. The 29 are the (c) rewrites of two paragraphs, 7 lines split between code and audit at sentence boundaries, and the pointer.
- **Verbatim:** 9/9 moved paragraphs.

```
swift comment lines: before 289, after 237, removed 52
audit lines added: 81 = moved body 56 + structure 25 (headings, labels, blank lines, fences)
moved paragraphs: 9, not verbatim: 0
remaining word diff (original minus moved  →  new comments):
  insert: -[]  +[History: docs/app-audits/com-windscribe-client.md#历史与实测]
  replace: -[below)]  +[History)]
  replace: -[prevented and is what actually happens —]  +[prevented, by design:]
  replace: -[string (proven by extracting both the 2.24.12 stable and the 2.24.10 beta bundles),]  +[string,]
  replace: -[signal]  +[signal,]
  replace: -[calls every copy stable. `channel-verify`]  +[`WindscribeChannel` answers nil for the Release preference — the default, which installing a beta dmg does not change. A prerelease build]
  insert: -[]  +[that preference is therefore compared against]
  replace: -[real beta bundle reports `UPDATE 2.24.10 → 2.24.12` through the full production chain.]  +[release track.]
  insert: -[]  +[it is the vendor's own ladder (a client on Release is served the release track),]
  insert: -[]  +[only]
  delete: -[Windscribe's own client on the Beta channel offers the same build (its API answers "this track or better"),]  +[]
  replace: -[It cannot be gated, either, since]  +[A copy on]
  replace: -[gate would need the channel we cannot read.]  +[Beta or Guinea Pig preference is bound to its own track recipe below.]
  insert: -[]  +[`duo verify` files an installed copy under its resolved channel, so a copy on the Beta or Guinea Pig preference is compared against its own track.]
  delete: -[machine carrying a]  +[]
  insert: -[]  +[left on the Release preference]
  replace: -[therefore]  +[not: it is]
  replace: -[build — measured, that state holds on 618 of the last 819 days —]  +[build,]
  replace: -[here; it goes away once the channel recipes land and such a copy is compared against its own track.]  +[here.]
reconciliation: moved body lines 56, of which partial (line split between code and audit) 7; blank comment lines 28->25; code lines not present verbatim in original (pointer/(c)/reflow) 29; original non-blank lines neither kept verbatim nor moved verbatim ((c)/reflow/split) 29
check: removed 52 == full lines moved 49 + blank dropped 3 + gone-only 29 - new-only 29 -> 52
```

| # | original lines | block | class |
|---|---|---|---|
| 1 | 7 | `MARK` | stay |
| 2 | 9–14 | no Sparkle / cask `auto_updates` → recipe needed | stay; **(c)-fixed** "see below" → "see History" (the paragraph it pointed at moved) |
| 3 | 16–24 | endpoint is the vendor site's; `Bearer 1234` is a presence check | stay (the measurement is inside the contract sentence) |
| 4 | 26–34 | why `summary` not `CheckUpdate` | first two sentences stay; "on 2026-09-07 `beta=0…3` byte-identical" **move** |
| 5 | 36–45 | 618/819-day (75%) risk analysis and the BROKEN experiment | **move** |
| 6 | 47–57 | `summary` names each track; excerpt | stay |
| 7 | 59–69 | exact key is the channel gate; test name | stay |
| 8 | 71–75 | `release_version` trap | stay |
| 9 | 77–87 | `(?!"platform")` boundary | stay, except "Measured on the real body with that field deleted…" **move** |
| 10 | 89–91 | 14 KB vs 395 B; release date | stay |
| 11 | 93–110 | no one-click | stay, except the installer-stub / `tar.lzma` / "Both are fixable" sentences **move** |
| 12 | 112–124 | CROSS-CHANNEL | **(c)-fixed**: it said `detect()` calls every copy stable and that the gate "would need the channel we cannot read". `WindscribeChannel` now reads it and answers nil only for the Release preference. The stale sentences "(proven by extracting both … bundles)" and "`channel-verify` … reports `UPDATE 2.24.10 → 2.24.12`" are **(c)-deleted**: they are no longer true in the present tense and the audit already records both (Channel 详情, 生产全链上今天的实际行为). |
| 13 | 126–131 | "Verified 2026-09-07 against the real bundle…" | **move** |
| 14 | 133–144 | EXPECTED WARNING | first two sentences stay; rest **(c)-fixed**: it said the warning "goes away once the channel recipes land". They have landed, and `Verify.installedVersions` now keys a copy by its resolved channel, so only a prerelease build on the Release preference still warns. "618 of the last 819 days" is **(c)-deleted** (the audit has it at 更新检测 §为什么不是 `/CheckUpdate`). |
| 15 | 146–151 | moving OS floor | stay |
| 16 | 166–168 | beta / guinea pig header | stay |
| 17 | 170–179 | ladder, not parallel trains | stay |
| 18 | 181–187 | `summary` matches once | first sentence stays; the alternation experiment and "Measured, not reasoned: `findall`" **move** |
| 19 | 189–195 | `ChangeLogs?platform=osx`, `entryStartPattern` | stay |
| 20 | 197–203 | "Simulated on the real 250 KB body (2026-09-07)…" | **move** |
| 21 | 205–211 | cost | first two sentences stay; the `duo verify` 153→155 / 141s→156s measurement **move** |
| 22 | 213–219 | detection only ⇒ no proof entry | stay |
| 23–27 | 222–263 | stable changelog: GitHub not vendor feed, stable-only split, tag shape, `.stable` covers only stable copies | stay |
| 28–29 | 272–284 | `includesPromotedStable` = ladder; why these two exist | stay |
| 30 | 286–293 | what it lists that it should not | first and last sentences stay; the "Cross-referenced … 12 guinea pig, 7 beta, 12…" counts **move** |
| 31–32 | 295–313 | residual ~4% gap; shipped anyway | stay |
| 33–34 | 326–329, 346–348 | `(?![0-9])`; `windscribeReleases` doc | stay |

**Copies found before moving:**
- the audit's own Chinese write-up covers every moved block; left as is
- `WindscribeProbeRecipeTests.swift` retells three of the moved facts (lockstep :21, `tar.lzma` :88, `Bearer 9999` :142)

The test copies were left alone because they document the tests, not the recipe (engine-notes checklist step 4). No other copies exist.

### CapCut (`Recipes/com-lemon-lvoverseas.swift`)

- **Comment lines:** 270 → 198 (−72)
- **Audit:** +93 lines = 68 moved body lines + 25 structure (heading, intro, 2 source lines, 2 labels, 2 fence lines, blank lines)
- **Reconciliation:** 72 removed = 63 whole lines moved + 8 blank separators dropped + 9 original lines rewritten/split − 8 new lines
- **Verbatim:** 12/13 moved paragraphs. The non-verbatim one is the machine-state rewording below, and its label line in the audit says so.

```
swift comment lines: before 270, after 198, removed 72
audit lines added: 93 = moved body 68 + structure 25 (headings, labels, blank lines, fences)
  NOT VERBATIM: One-click: enabled on BOTH tracks, and every gate was checked against the real artifacts rather than assumed (2026-08-27
moved paragraphs: 13, not verbatim: 1
remaining word diff (original minus moved  →  new comments):
  insert: -[]  +[History: docs/app-audits/com-lemon-lvoverseas.md#历史与实测]
  replace: -[which is the only]  +[a]
  delete: -[here]  +[]
  replace: -[SILENT.]  +[SILENT (so is the vendor rolling every bucket back at once, which no pin avoids — see the audit).]
  replace: -[issues…"),]  +[issues…").]
reconciliation: moved body lines 68, of which partial (line split between code and audit) 5; blank comment lines 29->21; code lines not present verbatim in original (pointer/(c)/reflow) 8; original non-blank lines neither kept verbatim nor moved verbatim ((c)/reflow/split) 9
check: removed 72 == full lines moved 63 + blank dropped 8 + gone-only 9 - new-only 8 -> 72
```

(`account.py` also prints the whole One-click paragraph as a `delete`. That output was filtered out above because the paragraph was reworded, so it is not a verbatim substring.)

**Machine-state rewording (the only change to moved text):** "and this machine has no CapCut launch item anywhere" → "and the machine measured on 2026-08-27 had no CapCut launch item anywhere".

| # | original lines | block | class |
|---|---|---|---|
| 1 | 7 | `MARK` | stay |
| 2 | 9–17 | no appcast; this endpoint is the only surface | stay |
| 3 | 19–25 | endpoint reconstructed from binaries; anonymous GET | stay |
| 4 | 27–36 | two-witness check; `update_url` is the per-device pick | stay (the dated sentence is what defines "the pick", which the contract sentence relies on) |
| 5 | 38–54 | `version_code` trap | contract sentences stay; "Measured 2026-08-27…" + table + "stable field was 9.3.0 for every value" **move**; "which is the only failure here that would be SILENT" **(c)-fixed**: the audit's 2026-09-08 section records every bucket rolling back at once, which no pin avoids |
| 6 | 56–64 | what pinning costs: cut off from deltas | stay |
| 7 | 66–69 | `cfg.diff_*` are log format strings | **move** |
| 8 | 71–79 | probed 2026-08-27, no `diff_update` | **move** |
| 9 | 81–84 | "the first version of this comment said…" | **move** |
| 10 | 86–96 | Sparkle binary delta inference | **move** |
| 11 | 98–103 | consuming it would be plumbing; the real blocker | **move** |
| 12 | 105–110 | two components, not three | stay |
| 13 | 112–115 | `channel` trap | stay |
| 14 | 117–124 | 2026-08-27 reason no longer true, re-measured 2026-09-04 | **move** |
| 15 | 126–131 | nibble-packed version, filename capture | stay |
| 16 | 133–155 | the two tracks' plist fields are swapped (incl. table) | stay (the dated intro sentence introduces the table that stays) |
| 17–19 | 157–174 | only `CapCutChannel` classifies beta; token inside pattern; beta suffix | stay |
| 20 | 176–186 | one-click gate verification (2026-08-27) | **move** (machine-state rewording above) |
| 21 | 188–190 | cask as a second witness | **move** |
| 22–26 | 192–228 | no checksum; 1.24 GB; `hostRequirement`; `downloadURL`; MAS copy | stay |
| 27 | 230–234 | no `changelogURL` | headline and closing stay; "capcut.com publishes no desktop release-notes page (… 502, 2026-08-27)" **move** |
| 28 | 239–252 | beta track closed | stay, except "Without this the row went red…" **move** |
| 29–32 | 260–267, 271–283, 300–305, 311–312 | channel proof; `capCutRecipe` doc; transient envelope; host pin | stay |

**Copies found before moving:**
- the audit's Chinese sections cover every moved block
- `CapCutProbeRecipeTests.swift` :138–142 still says `capcutpc_beta` "returns NOTHING here"
- `CHANNEL_COVERAGE_TODO.md` :116 summarises the one-click dmg contents
- `.claude/skills/app-audit/SKILL.md` :602 uses `diff_update` as a generic lesson

All left alone. The test's line is a test retelling that is itself stale since the 2026-09-04 re-measure, and it is flagged below rather than fixed here, since it's out of scope.

## 4. Carried items

- **Alcove.** `git log -S` shows a16c5be1 (2026-08-22) added the `api.tryalcove.com/changelog` `ChangelogRecipe` but left three comments contradicting it. **The recipe is current; all three comments were false.**
  - (c)-fixed: the probe's "Notes come from `changelogURL` in a WebView rather than a ChangelogRecipe" now names the recipe and says what `changelogURL` is still for (`ChangelogRecipeSelection.fallbackPage`).
  - (c)-deleted: the changelog array's "(No Alcove recipe … Re-add a recipe here only if Alcove publishes notes in a parseable form)". Its one still-true part, why the web page isn't scraped, is already in the probe comment.
  - (c)-fixed: the GitHub-rule tombstone named `update.tryalcove.com` as the public fallback (NXDOMAIN since e278050d) and said every public surface lags; it now names `download.tryalcove.com/latest`.
  - The note was about the changelog kind, not the probe, and a changelog recipe exists, so there was nothing to make explicit.
- **WorkBuddy.** Split into two files, linked both ways, and replaced the one index row with two.
  - `com-workbuddy-workbuddy.md`: the CN site, plus everything both sites share (identical updater code, the endpoint and its three traps, changelog markup traps, the host-pinning rationale for one-click, verification).
  - `com-workbuddy-workbuddy-ai.md`: the intl site's own facts, its matrix row, `TuringShield.bundle`, and the lagging changelog page.
  - Each doc's coverage matrix now has only its own row, so `AppAuditCoverageTests` checks each against its own bundle id.
- **VSCodium.** The filename rule would have accepted a rename (the doc mentions `` `com.vscodium` ``). But `com-vscodium-VSCodiumInsiders.md` is genuinely about Insiders only ("这份审计和对应的 recipe 只管 VSCodium 这一支"), and the new family rule needs `Recipes/com-vscodium.swift` pointers to land in `com-vscodium.md`. So instead of renaming, I **created** `com-vscodium.md` as an explicitly unaudited family doc that links to the Insiders audit, and listed it unchecked under 「仅迁出历史（未审计）」.

## Gates

- `python3 scripts/check_app_audits.py` → `✓ app audits consistent — 124 docs, all indexed, no machine inventory, no pointers into untracked evidence, history pointers resolve (799 Swift files, 226 families)`
- Executable Swift lines changed: `git diff origin/main -- '*.swift' | grep '^[-+]' | grep -v '^[-+][-+]' | grep -vE '^[-+][[:space:]]*(//|$)' | wc -l` → `0`
- `scripts/run-with-hang-report.sh 1800 make test > log 2>&1` (foreground) → `exit=0`, 212 s:
```
Ran 45 tests in 0.004s
OK
Ran 7 tests in 2.717s
OK
Ran 11 tests in 0.007s
OK
✓ no machine-population claims in code comments — 1 files
Ran 32 tests in 0.059s
OK
✓ no blocking calls left on the cooperative pool — 5 call(s) in 1 files
✓ no blocking calls left on the cooperative pool — 6 call(s) in 1 files
Ran 9 tests in 0.044s
OK
Ran 4 tests in 1.245s
OK
Ran 5 tests in 0.000s
OK
✓ downloaded Amp-1.0-355.dmg (5048 KB)
✓ EdDSA signature valid
✓ extracted Amp.app
✓ code signature valid; Team ID match: PZT9BJUAA5 == PZT9BJUAA5
━ Test run with 2994 tests in 258 suites passed after 20.208 seconds with 1 known issue.
✔ Test run with 334 tests in 42 suites passed after 6.367 seconds.
✓ localizable keys agree — 567 in the catalog, all reachable
✓ version comparisons discriminate — 495 files, 2 ledger call(s) keyed on the build, 9 marketing-first pick(s), all display-only
Ran 11 tests in 0.000s
OK
✓ app audits consistent — 124 docs, all indexed, no machine inventory, no pointers into untracked evidence, history pointers resolve (799 Swift files, 226 families)
✓ engine-notes pointers resolve — 799 Swift files scanned, 3 doc(s) indexed
✓ skill docs consistent — 3 files mirrored, documented parameter counts (25 / 27) match the initializers
✓ no machine-population claims in code comments — 799 files
✓ no blocking calls left on the cooperative pool — 9 call(s) in 495 files
```
(The "1 known issue" in the Core run was already in the suite; it is not from this branch.)

## Not done / worth a look

- `docs/app-audits/com-windscribe-client.md` still has stale **audit** text of the same kind as the comments I fixed: 已知问题 says beta / guinea pig can't be detected, and the "接之前必须先做的一件事" section predates `WindscribeChannel`. Audit prose was out of scope for this PR.
- `docs/app-audits/com-vscodium-VSCodiumInsiders.md` says the Insiders `installAssetPattern` is `(?:arm64|x64)`; the recipe comment says it was pinned to arm64 on purpose. Not re-checked or changed.
- `CapCutProbeRecipeTests.swift` :138–142 still asserts the pre-2026-09-04 reason for `capcutpc_0`.
- `App/Resources/Localizable.xcstrings`: the Alcove licence explainer says that without a key DuoUpdater "can't show what changed". Given the public changelog recipe, this looks stale. Not verified in the app.
- The guard's "families" count includes `AppRecipeIndex.swift` and `AppRecipeSet.swift` (226 files, 224 families). It only feeds the ≥100 vacuity floor.

<details><summary>account.py</summary>

```python
"""Comment accounting for one migrated family.
usage: account.py <swift before> <swift after> <audit path> <git rev of audit before>
1. line counts;  2. every moved paragraph is a verbatim (whitespace-normalised)
substring of the ORIGINAL comments;  3. original comments with the moved
paragraphs cut out, diffed word-by-word against the new comments — whatever
remains is exactly the (c) fixes, the pointer and seam punctuation."""
import re, sys, subprocess, difflib
before, after, audit_path, rev = sys.argv[1:5]
def comments(text):
    return [re.sub(r"^\s*//\s?", "", l) for l in text.splitlines() if l.lstrip().startswith("//")]
norm = lambda s: " ".join(s.split())
o, n = comments(open(before).read()), comments(open(after).read())
audit = open(audit_path).read().splitlines()
base = subprocess.run(["git", "show", f"{rev}:{audit_path}"], capture_output=True, text=True).stdout.splitlines()
hist = audit[audit.index("## 历史与实测"):]
META = ("## ", "### ", "从 recipe 注释迁出", "转引自")
paras, cur = [], []
for l in hist + [""]:
    if l.startswith(META) or l.strip() in ("", "```"):
        if cur: paras.append(cur); cur = []
    else:
        cur.append(l)
body_lines = sum(len(p) for p in paras)
print(f"swift comment lines: before {len(o)}, after {len(n)}, removed {len(o)-len(n)}")
print(f"audit lines added: {len(audit)-len(base)} = moved body {body_lines} + structure {len(audit)-len(base)-body_lines} (headings, labels, blank lines, fences)")
O = norm(" ".join(o))
rest = O
bad = 0
for p in paras:
    t = norm(" ".join(p))
    if t not in O:
        bad += 1; print("  NOT VERBATIM:", t[:120])
    rest = rest.replace(t, " ", 1)
print(f"moved paragraphs: {len(paras)}, not verbatim: {bad}")
a, b = norm(rest).split(), norm(" ".join(n)).split()
print("remaining word diff (original minus moved  →  new comments):")
for tag, i1, i2, j1, j2 in difflib.SequenceMatcher(None, a, b, autojunk=False).get_opcodes():
    if tag != "equal":
        print(f"  {tag}: -[{' '.join(a[i1:i2])}]  +[{' '.join(b[j1:j2])}]")
full = set(l.strip() for l in o)
partial = [l for p in paras for l in p if l.strip() not in full]
blank_o = sum(1 for l in o if not l.strip()); blank_n = sum(1 for l in n if not l.strip())
newonly = [l for l in n if l.strip() and l.strip() not in full]
goneonly = [l for l in o if l.strip() and l.strip() not in set(x.strip() for x in n) and l.strip() not in set(x.strip() for p in paras for x in p)]
print(f"reconciliation: moved body lines {body_lines}, of which partial (line split between code and audit) {len(partial)}; "
      f"blank comment lines {blank_o}->{blank_n}; code lines not present verbatim in original (pointer/(c)/reflow) {len(newonly)}; "
      f"original non-blank lines neither kept verbatim nor moved verbatim ((c)/reflow/split) {len(goneonly)}")
full_moved = body_lines - len(partial)
print(f"check: removed {len(o)-len(n)} == full lines moved {full_moved} + blank dropped {blank_o-blank_n} + gone-only {len(goneonly)} - new-only {len(newonly)} -> {full_moved + (blank_o-blank_n) + len(goneonly) - len(newonly)}")
```
</details>

---

## Review round 1

Three additive commits on top of `ef17ffc6`:
- `db6fab27` checker
- `e99cac9d` convention
- `97093bf3` pilots

HEAD is `97093bf3`. The pilot numbers and tables in §3 above describe round 0; the current state is below.

### Finding 3 was only partly right

The finding says `Verify.installedVersions` falls back to `app.releaseChannel`, "which is `.stable` for every Windscribe copy". The fallback is real, but that is not the value it falls back to:
- `AppScanner.readApp` applies `ChannelBinding` to every scanned copy (`DuoUpdaterCore/Sources/DuoUpdaterCore/Scan/AppScanner.swift:723-726`).
- For Windscribe that means `WindscribeChannel.resolveCurrent`. When the preference reads as Beta or Guinea Pig, the scan sets `releaseChannel` to that track and `channelIsAuthoritative = true`.
- `Verify.swift:1338-1341` then skips `ResolvedChannelStore` and uses that channel.
- The store is consulted only for non-authoritative copies. For Windscribe those are copies on the Release preference (the resolver returns nil for `0`, `WindscribeChannel.swift` `channel(forRawValue:)`) and copies whose preference could not be read. The resolver also returns nil when its plist is missing or undecodable, or when the 2 s `ChannelBinding.resolveTimeout` expires.

What was wrong in round 0 was the missing condition, not the mechanism. The comment now says:
- a copy whose preference `WindscribeChannel` reads as Beta/Guinea Pig is filed under its own track;
- every other copy is filed under stable unless `ResolvedChannelStore` holds a proven channel for it;
- that includes an unreadable or timed-out preference.

The CROSS-CHANNEL paragraph now carries the same condition.

### What changed per finding

1. **CapCut `channel`**
   - Corrected in place, as a (c)-fix: the token is currently inert, `capcutpc_0` is kept for its longer measured history, re-measure before changing it.
   - I re-measured instead of relying on the 09-04/09-08 notes. On 2026-09-13T23:16Z, the same URL with `channel=capcutpc_0` and with `channel=capcutpc_beta` both returned HTTP 200. The two `update_reminder` objects have 26 fields each and match field for field.
   - That re-measure is recorded in History as a separate `复测` line.
   - `CapCutProbeRecipeTests.swift:141` was a comment-only fix, no assertion changed.
2. **Windscribe cost:** "`duo verify` pays all three, because it walks the registry rather than the installed apps" plus "Worth stating both ways round…" are back in code. The 500 KB, 153→155 and 141s→156s numbers stay in History.
3. **Windscribe EXPECTED WARNING:** condition stated, as above.
4. **Windscribe alternation:** back in code. "Measured, not reasoned: `findall`…" stays in History.
5. **CapCut bucket exposure:** "The stable field was 9.3.0 for every value that answered at all, so only the beta recipe is exposed to this" is back in code. The table and its date stay in History.
6. **Checker:** `RECIPES_INFRASTRUCTURE = {"AppRecipeSet.swift", "AppRecipeIndex.swift"}`, mirroring `AppRecipeIndexTests.infrastructure`. The success line now says 224 families.

The standalone re-read the new convention asks for turned up more instances of the same pattern. All fixed in the same commit:
- **Windscribe:**
  - "`/ChangeLogs/summary` sidesteps the question" dangled once the question moved. The question and its conclusion (a prerelease leads for most of each cycle, so a CheckUpdate-based recipe would be BROKEN for weeks) are back in code, reworded to drop the date and counts. History keeps the original paragraphs.
  - The "the bug would look like a pass" platform-boundary warning is back.
  - The regression-test dates sentence is back.
  - "no `versionIsBuild` because both plist fields hold one value" is back as a conclusion.
- **CapCut:**
  - The `cfg.diff_*` "these are log format strings, not response fields" caveat is back, because the paragraph above it cites those strings as evidence.
  - "The pin costs nothing today" and "un-pinning needs a way to put the installed version on the wire" are back.
  - "One-click … the bundle swap IS the whole update" is back.
  - "capcut.com publishes no desktop release-notes page" is back.
  - "Without this the row went red, with a Retry that could not work" is back.
  - **(c):** "`9.3.0` would trip that the moment stable is 9.3.0 — which it is today" is false now: stable was `9_4_1_4574` in today's response. It now reads "as it was when this was written".

### Accounting (rerun, base audit rev `7b41abf5`)

| | comment lines (orig → round 0 → round 1) | audit lines added (round 1) | verbatim |
|---|---|---|---|
| Windscribe | 289 → 237 → **274** (37 lines restored, incl. conditions) | 67 = 43 moved body + 24 structure | 8/8 |
| CapCut | 270 → 198 → **220** (22 lines restored) | 84 = 58 moved body + 26 structure (incl. the `复测` line) | 10/11; the other is the machine-state rewording |

Windscribe:
```
swift comment lines: before 289, after 274, removed 15
audit lines added: 67 = moved body 43 + structure 24 (headings, labels, blank lines, fences)
moved paragraphs: 8, not verbatim: 0
remaining word diff (original minus moved  →  new comments):
  insert: -[]  +[History: docs/app-audits/com-windscribe-client.md#历史与实测]
  replace: -[below)]  +[History)]
  insert: -[]  +[The problem is that "the parameter selects the release track" and "the parameter is ignored" cannot be told apart from its answers while release leads: every value's correct answer is then the same (History has the byte-identical responses). That ambiguity is not academic, and its risky branch is the NORMAL state of this vendor: a prerelease outranks the newest release for most of each cycle. If the parameter is ignored, then throughout those windows `CheckUpdate` names a `…_beta_universal.dmg` and a recipe reading that filename would report `versionPatternNoMatch` — the app `.unknown` and `duo verify` BROKEN, for weeks at a time, four times a day.]
  replace: -[prevented and is what actually happens —]  +[prevented, by design:]
  replace: -[string (proven by extracting both the 2.24.12 stable and the 2.24.10 beta bundles),]  +[string,]
  replace: -[signal]  +[signal,]
  replace: -[calls every copy stable. `channel-verify`]  +[`WindscribeChannel` answers nil for the Release preference — the default, which installing a beta dmg does not change. A prerelease build]
  insert: -[]  +[that preference is therefore compared against]
  replace: -[real beta bundle reports `UPDATE 2.24.10 → 2.24.12` through the full production chain.]  +[release track.]
  insert: -[]  +[it is the vendor's own ladder (a client on Release is served the release track),]
  insert: -[]  +[only]
  delete: -[Windscribe's own client on the Beta channel offers the same build (its API answers "this track or better"),]  +[]
  replace: -[It cannot be gated, either, since]  +[A copy on]
  replace: -[gate would need the channel we cannot read.]  +[Beta or Guinea Pig preference is bound to its own track recipe below, when `WindscribeChannel` can read that preference (see EXPECTED WARNING for when it cannot). The bundle's `CFBundleShortVersionString` == `CFBundleVersion` (the plist template writes one value into both, so no `versionIsBuild`).]
  replace: -[A machine carrying]  +[`duo verify` files an installed copy under the channel the scan gave it (`Verify.installedVersions`), and `AppScanner.readApp` applies `ChannelBinding` to every copy, so a copy whose preference `WindscribeChannel` reads as Beta or Guinea Pig is compared against its own track. Every other copy is filed under stable unless an earlier check stored a proven channel for it (`ResolvedChannelStore`):]
  insert: -[]  +[left on the Release preference, and any copy whose preference could not be read (missing or undecodable plist, or the resolver's time bound expiring). A prerelease copy filed under stable]
  delete: -[therefore]  +[]
  replace: -[build — measured, that state holds on 618 of the last 819 days —]  +[build,]
  replace: -[here; it goes away once the channel recipes land and such a copy is compared against its own track.]  +[here.]
  replace: -[apart,]  +[apart (stable, beta and guinea pig give the same answer while release leads),]
  replace: -[apps:]  +[apps.]
reconciliation: moved body lines 43, of which partial (line split between code and audit) 8; blank comment lines 28->28; code lines not present verbatim in original (pointer/(c)/reflow) 48; original non-blank lines neither kept verbatim nor moved verbatim ((c)/reflow/split) 31
lines present both in the audit and still in code (a conclusion kept in code, its paragraph kept whole in History): 3
check: removed 15 == full lines moved 35 - also still in code 3 + blank dropped 0 + gone-only 31 - new-only 48 -> 15
```
CapCut (the One-click paragraph's `delete` is filtered out, as in round 0):
```
swift comment lines: before 270, after 220, removed 50
audit lines added: 84 = moved body 58 + structure 26 (headings, labels, blank lines, fences)
  NOT VERBATIM: One-click: enabled on BOTH tracks, and every gate was checked against the real artifacts rather than assumed (2026-08-27
moved paragraphs: 11, not verbatim: 1
remaining word diff (original minus moved  →  new comments):
  insert: -[]  +[History: docs/app-audits/com-lemon-lvoverseas.md#历史与实测]
  replace: -[bucket.]  +[bucket (History has the measured buckets).]
  replace: -[which is the only]  +[a]
  delete: -[here]  +[]
  replace: -[SILENT.]  +[SILENT (so is the vendor rolling every bucket back at once, which no pin avoids — see the audit).]
  insert: -[]  +[pin costs nothing today: the]
  replace: -[now.]  +[now (History has how that was checked).]
  replace: -[the moment]  +[whenever]
  replace: -[9.3.0 — which]  +[9.3.0, as]
  replace: -[is today.]  +[was when this was written.]
  delete: -[TRAP,]  +[]
  delete: -[deliberately]  +[]
  insert: -[]  +[The token is currently inert — both values answer with the same `update_reminder` (History has the measurements) — so `capcutpc_0` is kept for its longer measured history, not because `capcutpc_beta` is known to be harmful. Re-measure before changing it.]
  replace: -[tracks, and every gate was checked against the real artifacts rather than assumed (2026-08-27, both dmgs downloaded and mounted).]  +[tracks.]
  replace: -[`LaunchDaemons`/`LaunchAgents`/ `PrivilegedHelperTools`, and this machine has no CapCut launch item anywhere]  +[`LaunchDaemons`/`LaunchAgents`/`PrivilegedHelperTools`]
  replace: -[update. The Homebrew cask agrees independently: `artifacts` is `{"app": ["CapCut.app"]}` and its `uninstall` is a bare `quit`, its `zap` only user data. Both bundles are Developer ID signed by Team `22MMUN2RN5` (BYTEDANCE PTE. LTD.) and `spctl -t install` reports "Notarized Developer ID" — the same Team the installed copy carries, which is `VendorInstaller`'s mandatory gate.]  +[update (History has how each gate was checked).]
  insert: -[]  +[and capcut.com publishes no desktop release-notes page.]
  insert: -[]  +[Without this the row went red, with a Retry that could not work until ByteDance opened the next beta.]
reconciliation: moved body lines 58, of which partial (line split between code and audit) 5; blank comment lines 29->24; code lines not present verbatim in original (pointer/(c)/reflow) 23; original non-blank lines neither kept verbatim nor moved verbatim ((c)/reflow/split) 15
lines present both in the audit and still in code (a conclusion kept in code, its paragraph kept whole in History): 0
check: removed 50 == full lines moved 53 - also still in code 0 + blank dropped 5 + gone-only 15 - new-only 23 -> 50
```

`account.py` gained one reconciliation term: lines present both in History and still in code. Windscribe has 3 of them: History keeps a paragraph whole while code keeps its reworded conclusion.

### Item 6 mutation

```
### M6 pointer in AppRecipeSet.swift — checker BEFORE the fix (HEAD ef17ffc6)
✗ app audits disagree with the rules that keep them publishable:
    DuoUpdaterCore/Sources/DuoUpdaterCore/Recipes/AppRecipeSet.swift:20: history pointer names `com-lemon-lvoverseas.md` inside the `AppRecipeSet` family — a family's history goes to `docs/app-audits/AppRecipeSet.md`
exit=1
### M6 same pointer — checker AFTER the fix
✓ app audits consistent — 124 docs, all indexed, no machine inventory, no pointers into untracked evidence, history pointers resolve (799 Swift files, 224 families)
exit=0
### restored
✓ app audits consistent — 124 docs, all indexed, no machine inventory, no pointers into untracked evidence, history pointers resolve (799 Swift files, 224 families)
exit=0
### M4 re-run after the fix (wrong family in a real family file)
✗ app audits disagree with the rules that keep them publishable:
    DuoUpdaterCore/Sources/DuoUpdaterCore/Recipes/com-lemon-lvoverseas.swift:10: history pointer names `com-windscribe-client.md` inside the `com-lemon-lvoverseas` family — a family's history goes to `docs/app-audits/com-lemon-lvoverseas.md`
exit=1
```

### Gates

- `python3 scripts/check_app_audits.py` → `✓ … history pointers resolve (799 Swift files, 224 families)`
- Executable Swift lines vs `origin/main`: `0`
- `scripts/run-with-hang-report.sh 1800 make test > log 2>&1` (foreground) → `exit=0`, 116 s:
```
━ Test run with 2994 tests in 258 suites passed after 20.362 seconds with 1 known issue.
✔ Test run with 334 tests in 42 suites passed after 6.317 seconds.
✓ localizable keys agree — 567 in the catalog, all reachable
✓ version comparisons discriminate — 495 files, 2 ledger call(s) keyed on the build, 9 marketing-first pick(s), all display-only
✓ app audits consistent — 124 docs, all indexed, no machine inventory, no pointers into untracked evidence, history pointers resolve (799 Swift files, 224 families)
✓ engine-notes pointers resolve — 799 Swift files scanned, 3 doc(s) indexed
✓ skill docs consistent — 3 files mirrored, documented parameter counts (25 / 27) match the initializers
✓ no machine-population claims in code comments — 799 files
✓ no blocking calls left on the cooperative pool — 9 call(s) in 495 files
```

### Convention (`docs/app-audits/README.md`)

The "one sentence at a time" rule now has four overriding rules under it, each with its pilot example:
1. A measurement's conclusion stays; only the measurement moves.
2. A correction travels with the claim it corrects, including test comments that repeat the claim.
3. A warning that a wrong change passes today stays.
4. A (c) rewrite is a new claim: name the symbol in the comment, give file:line in the PR, and state its conditions. The rule asks for the symbol in the comment and file:line in the PR, not file:line in the comment, because line numbers in comments go stale (`ChangelogRecipeSelection.swift`'s doc records that happening).

Added the batch step: after moving, re-read the Swift file on its own. Every sentence must still be true, including anything left saying "today", and nothing may dangle.

Also: a sentence restored whole to code is deleted from History, so it lives in one place. A reworded conclusion keeps its original paragraph whole in History.


jizhi0v0:
## Review round 1 — not mergeable yet

**Verified fine:** 0 non-comment Swift lines changed; `check_history_pointers` rules and vacuity floor read correctly; Alcove resolution holds (`ChangelogRecipeSelection.fallbackPage` exists, no GitHub rule in the family, the deleted 1.7.4 lag measurement's canonical copy is `AlcoveUpdateSource.swift:7`); WorkBuddy split links resolve.

**Blocking — the pilots moved current contract because it carried a date or a number.** The batches will copy whatever the pilots do, so this is fixed in the pilots *and* written into the convention:

1. `com-lemon-lvoverseas.swift:69` — the correction ("the `capcutpc_beta` hazard is no longer true, re-measured 2026-09-04") moved out; the stale `TRAP, channel … deliberately send capcutpc_0` framing stayed, and `CapCutProbeRecipeTests.swift:141` still says the beta token "returns NOTHING here".
2. `com-windscribe-client.swift:161` — kept "An APP pays one of the three", moved "`duo verify` pays all three" — the half the original author wrote is misleading on its own.
3. `com-windscribe-client.swift:103` — the (c)-rewrite says `duo verify` compares a Beta-preference copy against its own track; `Verify.installedVersions` does that only when `ResolvedChannelStore` holds a proof, and falls back to `.stable` otherwise.
4. `com-windscribe-client.swift:148` — "an alternation looks like it works today and fails the day a beta leads" is a wrong-fix-passes-today warning; it stays in code.
5. `com-lemon-lvoverseas.swift:41` — "only the beta recipe is exposed to the bucket" is the conclusion of the moved table; it stays.

**Non-blocking:** `check_app_audits.py:245` counts `AppRecipeIndex.swift` / `AppRecipeSet.swift` as families (success line says 226, index has 224) and would demand `AppRecipeSet.md` for a pointer written there.

Convention additions requested: a measurement's conclusion stays and only the measurement moves; a correction travels with the claim it corrects; wrong-fix-passes-today warnings stay; a (c)-rewrite is a new claim and needs file:line evidence with its conditions.

Issue #409, opened 2026-09-07
重整核心文件注释：将事故经过与测量历史迁入文档，保留当前契约


jizhi0v0:
近期核心路径把当前规则、事故经过、旧实现反驳、个人环境测量重复写进代码，增加阅读与维护成本，且新修改容易只更新其中一份说明。

本次审查基于提交 e0143bd804e5f7f293cb00e5f881fc9002bcda7a，按去除前导空白后以 // 开头的整行统计：
- App/Sources/AppListModel.swift：6,427 行，其中 3,296 行为整行注释。
- DuoUpdaterCore/Sources/DuoUpdaterCore/Sources/AppStorePageCache.swift：227 行，其中 156 行为整行注释。
- DuoUpdaterCore/Sources/DuoUpdaterCore/Engine/PreInstallGate.swift：116 行，其中 83 行为整行注释。

这些数字可直接从固定提交重算，不是本机应用数量测量；比例只是定位入口，不是要求压到某个数值。

示例：
- [DuoUpdaterCore/Sources/DuoUpdaterCore/Sources/AppStorePageCache.swift:3](https://github.com/jizhi0v0/duo-updater/blob/e0143bd804e5f7f293cb00e5f881fc9002bcda7a/DuoUpdaterCore/Sources/DuoUpdaterCore/Sources/AppStorePageCache.swift#L3)
- [DuoUpdaterCore/Sources/DuoUpdaterCore/Engine/PreInstallGate.swift:3](https://github.com/jizhi0v0/duo-updater/blob/e0143bd804e5f7f293cb00e5f881fc9002bcda7a/DuoUpdaterCore/Sources/DuoUpdaterCore/Engine/PreInstallGate.swift#L3)
- [App/Sources/AppListModel.swift:2797](https://github.com/jizhi0v0/duo-updater/blob/e0143bd804e5f7f293cb00e5f881fc9002bcda7a/App/Sources/AppListModel.swift#L2797)

建议按模块小步整理：代码旁保留当前契约、非直观约束、边界、以及对应测试/设计文档入口；将事故时间线、完整测量与旧方案讨论迁入维护文档或引用已有 issue。迁移时核对历史结论是否仍成立，不要只把旧长注释原样复制成新的规格书。

验收应以信息可找回、约束不丢失、当前行为不变为准；不按删行数评价，也不增加专门检查注释措辞的测试。先从上述小文件开始，再处理 AppListModel，避免一次巨大的功能与文档混合重构。

[...]

PR #609, opened 2026-09-14
app: delete the dead beginTestFlightLoad helper

jizhi0v0:
Deletes `AppListModel.beginTestFlightLoad(timeout:)` and its doc comment. 19 lines removed, one file, nothing else touched.

`#409` deliberately left this out of scope as a comments-only slice.

## Why it's dead

Its two call sites went away in 599e8dde (2026-06-04, "feat: broaden update coverage"); the definition survived. `performRefresh` has inlined the same logic since — the `tfLoader` detached load (`AppListModel.swift:2501`), the two-second `firstResult` race (`:2608`), the same `"TestFlight: read pending (TCC prompt unanswered) — proceeding without it"` log line (`:2617`), and the handoff to `reapplyTestFlightWhenGranted` (`:2618`).

## What I checked before deleting

- **Nothing reaches it.** Grep repo-wide, not just `App/Sources CLI/Sources` — every file type, excluding `.git`/build dirs. One hit: the definition.
- **The commit claim holds.** This checkout arrived shallow (50 commits, back to 2026-09-12), so `599e8dde` resolved to nothing at first — that is clone depth, not a bad citation. After `git fetch --unshallow`, `git log -S'beginTestFlightLoad('` names 599e8dde as the last commit to touch a call site, and its diff shows both `await Self.beginTestFlightLoad()` lines removed with the definition left as context.
- **`firstResult` is not orphaned.** It was the helper this function depended on; it keeps ~15 other callers in this same file (`:503`, `:1495`, `:2215`, `:2332`, `:2538`, `:6716`, `:7152`, …), so removing its one dead caller retires nothing.

## Verification

**`make test` passes on this head — see [the tail in this comment](https://github.com/jizhi0v0/duo-updater/pull/609#issuecomment-5657667094), taken from CI.**

I could not run `scripts/run-with-hang-report.sh 1800 make test` in the session that wrote this: it runs on Linux, and the repo is macOS-only (`AppListModel.swift` imports AppKit; no `swift`, `xcodebuild`, or `xcrun` on the box). It died at step 5 of 17 on `FileNotFoundError: 'xcrun'` in `test_check_swift_backdeploy.py` — the environment, not this diff. So I ran every step that doesn't need the Xcode toolchain individually (all green), and CI then ran the whole gate on exactly this commit (`head_sha` `be32c803` == PR head), including the five steps I couldn't.

The one that matters here is `check_localizable_keys.py`: it does a full Release `xcodebuild -scheme DuoUpdater build` to harvest string metadata, which compiles `AppListModel.swift` as a side effect. **That is the compile this deletion needed, and it passed** — an earlier revision of this description said nothing had compiled the file yet, which CI has since retired.

Also green: `✓ App tests — 48 of 48 cases executed` (the anti-vacuity gate reporting a real run, not an empty one), and the CLI suite at 334 tests in 42 suites. `AppListModel.swift` is in no test target — it pulls in SwiftUI via `SettingsView` — so no test *executes* this code either before or after; the deletion's correctness rests on it being unreferenced, which is the grep above.

https://claude.ai/code/session_0125HFpwv2pTqmFWSdJkYXsx

---
_Generated by [Claude Code](https://claude.ai/code/session_0125HFpwv2pTqmFWSdJkYXsx)_


jizhi0v0:
## `make test` tail — from CI, since this session can't run it

The PR body explains why I couldn't run `scripts/run-with-hang-report.sh 1800 make test` locally (Linux session, no `swift`/`xcodebuild`/`xcrun`). CI runs that exact gate — `ci.yml` invokes `make test` through `run-with-hang-report.sh` — so here is the real tail, from [run 34794504451](https://github.com/jizhi0v0/duo-updater/actions/runs/34794504451/job/103824893231), `test` ✅ success in 9m50s.

Verified it's this tree, not a stale green: the run's `head_sha` is `be32c803315db0f01a8c6bb798e92320c257edb3`, equal to the PR head.

```
✔ Test run with 334 tests in 42 suites passed after 6.308 seconds.
python3 scripts/check_localizable_keys.py
✓ localizable keys agree — 567 in the catalog, all reachable
✓ version comparisons discriminate — 495 files, 2 ledger call(s) keyed on the build, 9 marketing-first pick(s), all display-only
python3 scripts/test_check_staged_version_use.py
Ran 11 tests in 0.000s
OK
✓ app audits consistent — 124 docs, all indexed, no machine inventory, no pointers into untracked evidence, history pointers resolve (799 Swift files, 224 families)
python3 scripts/check_engine_notes.py
✓ engine-notes pointers resolve — 799 Swift files scanned, 3 doc(s) indexed
python3 scripts/check_skill_docs.py
✓ skill docs consistent — 3 files mirrored, documented parameter counts (25 / 27) match the initializers
python3 scripts/check_prose_claims.py
✓ no machine-population claims in code comments — 799 files
python3 scripts/check_offpool.py
✓ no blocking calls left on the cooperative pool — 9 call(s) in 495 files
```

### This closes the gap the PR body flagged

The body said "**nobody has compiled this file since the edit**" — that is now stale, and this is what retires it. `check_localizable_keys.py` runs a full Release `xcodebuild -scheme DuoUpdater build` (`CODE_SIGNING_ALLOWED=NO`, `SWIFT_EMIT_LOC_STRINGS=YES`) to harvest string metadata, which compiles `App/Sources/AppListModel.swift` as a side effect. It passed. So the deletion is a real compile, not just a balanced-brace count.

`✓ App tests — 48 of 48 cases executed` is the anti-vacuity gate in `app-tests.sh` reporting a non-zero, fully-executed run rather than a silently empty one.

### Two things this tail does not show

- **The `DuoUpdaterCore` suite's own summary line** scrolled past the retrievable window. `334 tests in 42 suites` is the CLI package (Restart, Triage, ArgParser, Verify…), not Core. Core passing is an inference, though a sound one: `make` runs the recipe sequentially and aborts on the first non-zero step, and the job both concluded success and reached the final step, so `cd DuoUpdaterCore && swift test` cannot have failed. I'd rather say that than quote a count I couldn't read.
- **The vendor-download gate** (`DUO_DOWNLOAD_GATE`) is on in CI, so the ~185 MB path did run here — but that gate is about vendor bytes and has nothing to do with this diff.

Left as a draft; say the word and I'll mark it ready.

---

PR #610, opened 2026-09-14
CCC beta: accept the build its cycle graduated into

### LINKED ISSUE #612: Recipe broken: com.bombich.ccc (vendor probe) — versionPatternNoMatch
Found by the scheduled `duo verify` sweep. The version check and the app both keep working in the cases marked *degraded* — that is what makes these invisible without this job.

| | |
|---|---|
| recipe | `vendor:com.bombich.ccc:beta` |
| registry | vendor probe |
| endpoint host | `bombich.com` |
| pattern | `^ccc-(7(?:\.[0-9]+)+-b[0-9]+)\.[0-9]{3,}\.zip$` |
| failure | `versionPatternNoMatch` — no match in 16-byte body |
| last good | `7.1.7-b7` |
| consecutive sweeps | 2 |

<details><summary>Captured response (redacted, truncated)</summary>

```
ccc-7.2.8399.zip
```

</details>

### Automated analysis

✅ Verified: the proposed pattern extracts `7.2.8399` from the captured body, using the same extractor the app uses.

<details><summary>Unverified suggestion — do not apply without testing (model: `opencode-go/deepseek-v4-pro`, self-reported confidence 0.75)</summary>

Bombich's beta channel stopped tagging build names with a '-b' segment: the current download is now named ccc-7.2.8399.zip (the build number sits in the version's third dotted component) instead of ccc-7.x.y-bN.build.zip, so the pattern's mandatory '-b[0-9]+' part no longer matches anything. The body contains no directive-like content, just the bare filename.

Proposed pattern:

```
^ccc-(\d+(?:\.\d+)+)\.zip$
```

Produced by a model with no tools, in an empty directory, from the captured response above. That response is third-party content and may be hostile; nothing here has been applied to any file.

</details>

Reproduce locally:

```bash
swift run --package-path CLI duo verify --vendor --only com.bombich.ccc --samples
```

jizhi0v0:
Fixes #612.

Carbon Copy Cloner's beta probe (`?v=latestbeta`) is red because the vendor's beta train is **between cycles**, not because anything here broke. The `-b[0-9]+`-required `versionPattern` becomes `-b<N>`-optional — the call CotEditor's beta rule already makes for the same vendor shape.

## The failure, reproduced on the real response

Measured 2026-09-14 with the same HEAD-and-follow request `.redirectFilename` issues:

| query | final URL | `lastPathComponent` |
|---|---|---|
| `?v=latestbeta` | `https://bombich.download/ccc-7.2.8399.zip` | `ccc-7.2.8399.zip` |
| `?v=ccc7` | `https://bombich.download/ccc-7.2.8399.zip` | `ccc-7.2.8399.zip` |
| `?v=latest` | `https://bombich.download/ccc-7.2.8399.zip` | `ccc-7.2.8399.zip` |
| `?v=ccc6` | `https://bombich.download/ccc-6.1.13.7699.zip` | `ccc-6.1.13.7699.zip` |
| `?v=ccc5` | `https://bombich.download/ccc-5.1.28.6213.zip` | `ccc-5.1.28.6213.zip` |

`^ccc-(7(?:\.[0-9]+)+-b[0-9]+)\.[0-9]{3,}\.zip$` matches none of that → `ProbeFailed` / `versionPatternNoMatch`. A 7.1.7-b7 copy shows a failed check and is never offered the 7.2 its own train graduated into; `duo verify` walks **recipes, not installs**, so the miss is a red finding on every machine for a recipe working exactly as written. The 02:26Z sweep reached the threshold and filed #612.

(The CDN host also moved off `bombich.scdn1.secure.raxcdn.com` since the 2026-08-30 captures. `.redirectFilename` reads the final URL's last path component, so the recipe never saw it. The old records keep their host verbatim under their own date.)

## The decision, and the evidence for it

**Yes, the beta pattern should accept a plain `ccc-7.x.NNNN.zip`.** That the 7.1.7 betas graduated as **7.2** is the vendor's own text, not an inference — two independent witnesses:

1. `ccc7_rn.html` (last-modified 2026-09-13) has **no 7.1.7 section at all**: the headings go `CCC 7.1.6: July 15, 2026` → `CCC 7.2`. The train never landed as `7.1.7`.
2. 7.2's "What's new" is `ccc7_rn_beta.html`'s "What we're testing in this beta cycle" item for item — the bootable-backup hybrid policy, both "When files are modified on the source" relaxations, the Settings redesign, the locked-screen `NSFileProtectionComplete` handling, the `/.streams` exclusion, the SAMBA hidden-flag fix, "Sign in with Microsoft" OAUTH, `ccc --watch`, "Removed TouchBar support" — and it says so outright: *"During our latest beta testing cycle"*. The beta page has been frozen on `CCC 7.1.7-b7` since 2026-08-18 while stable moved on.

So this is the vendor's ordinary resting state between cycles, and anchoring to `-b` fails on it in exactly the two ways CotEditor's rule spells out: the copy is never offered its own graduation, **and** the miss is a red sweep finding rather than a quiet nil.

`^ccc-(7(?:\.[0-9]+)+(?:-b[0-9]+)?)\.[0-9]{3,}\.zip$`

### Not the pattern #612's automated analysis suggests

#612 proposes `^ccc-(\d+(?:\.\d+)+)\.zip$` behind a ✅. That check only confirms the pattern extracts *something*; what it extracts is `7.2.8399` — marketing and build glued together, a string no CCC bundle's `CFBundleShortVersionString` ever equals, i.e. a permanent phantom update. Ported to Python over every filename, it is wrong three ways:

| filename | #612's suggestion | this PR |
|---|---|---|
| `ccc-7.2.8399.zip` | `7.2.8399` (marketing+build glued) | `7.2` |
| `ccc-7.1.7-b7.8389.zip` | **nil** — the beta stops resolving while a cycle is open | `7.1.7-b7` |
| `ccc-6.1.13.7699.zip` | `6.1.13.7699` (**cross-generation**) | nil |
| `ccc-5.1.28.6213.zip` | `5.1.28.6213` (**cross-generation**) | nil |
| `ccc-8.0.1.9000.zip` | `8.0.1.9000` (**cross-generation, paid upgrade**) | nil |

It removes the generation anchor the whole three-recipe split exists for. The issue's own "Unverified suggestion — do not apply without testing" is doing real work there.

## What this means for the two things asked about

**`ReleaseChannel` gating — no change, and now a checked decision.** Step 0.8 is untouched: a `-b<N>` copy still detects `.beta` and still selects this recipe. What changes is downstream and one-way — taking the graduation puts the copy on `7.2`, `detect` finds no `-b<N>`, answers `.stable`, and the stable recipe serves it from then on. **Unlike CotEditor this is not recoverable here**: `CotEditorChannel` reads that vendor's prerelease checkbox and outranks the version string; CCC has the same checkbox ("Inform me of beta releases") and **nothing in this repo reads it** — there is no `ChannelBinding` for `com.bombich.ccc`. So this is WhatCable's shape, not CotEditor's. Taken anyway: the alternative is a permanently red row that is offered nothing at all, and the recipe is detection-only — `downloadURL` sends the user to the vendor's own beta endpoint, which serves the same artifact this probe just read. `detectRecognizesTheShortBetaSuffix` now pins `7.2 → .stable` so the cost is asserted, not just remarked on.

**`ChannelProofRegistry` — no entry, and the reason is worth having written down.** `proofs` must cover `channelRecipesWithInstall`; this recipe has no `install`, so `RecipeSanity.crossChannelArtifact` returns nil at its first guard. But `-b[0-9]+` *was* the second, independent statement of which train an answer came from, and it is now satisfiable by a stable filename — so `url`'s `v=latestbeta` carries that alone. The day an install spec is added here, the mandatory proof is `.recipeAnchor(#"latestbeta"#, in: ["url"])` (endpoint-keyed, like IntelliJ EAP's and Alfred beta's), **never** an `.artifact` anchored to `-b`: that one passes every day the vendor has a cycle open and fails on exactly the release it exists to judge — the mistake CotEditor's own proof entry records having shipped for a day.

## Costs, all four

1. **One-way channel move** — see `ReleaseChannel` above.
2. **Changelog mismatch between cycles** — `ccc7_rn_beta.html` is the closed cycle's page, so the offered version and the notes beside it disagree. Not switched by shape: that would mean guessing the vendor's state from a filename, the inference this recipe avoids everywhere else.
3. **A retired beta rail is now invisible** (found by `/code-review`). Requiring `-b` had a side effect: if Bombich drops or renames `?v=latestbeta`, the probe goes red and somebody looks. It now resolves the stable zip and stays green forever — "resting" and "gone" are the same bytes, and nothing else would notice (no `install`, so no proof questions it; the cask is `auto_updates: true`, so no Homebrew cross-check). `trackClosedPattern` is not the fix: it wants the vendor to *say* the track is empty, and this endpoint says nothing.
4. **A copy ahead of the graduation reads a lower "latest"**, and `remoteBehindInstalled` raises an advisory. Reachable, not observed: the review claimed it had just happened (2026-08-18 → 09-13), but the baseline shows beta at `lastGoodVersion 7.1.7-b7` as late as `2026-09-13T14:24Z` — the notes page froze on 08-18, the endpoint did not.

## Independent recomputation

Per CLAUDE.md's rule for regex/version-comparison changes, both rules were ported to Python and run against the real responses rather than re-read off the Swift.

| filename | ccc5 | ccc6 | ccc7 | beta (old) | beta (new) |
|---|---|---|---|---|---|
| `ccc-5.1.28.6213.zip` | 5.1.28 | nil | nil | nil | nil |
| `ccc-6.1.13.7699.zip` | nil | 6.1.13 | nil | nil | nil |
| `ccc-7.1.1234.zip` | nil | nil | 7.1 | nil | 7.1 |
| `ccc-7.1.6.8368.zip` | nil | nil | 7.1.6 | nil | 7.1.6 |
| `ccc-7.1.7-b7.8389.zip` | nil | nil | nil | 7.1.7-b7 | 7.1.7-b7 |
| **`ccc-7.2.8399.zip`** (today's live answer) | nil | nil | 7.2 | **nil** | **7.2** |
| `ccc-8.0.1.9000.zip` | nil | nil | nil | nil | nil |
| `ccc-8.0.2-b1.9012.zip` | nil | nil | nil | nil | nil |

The major-7 anchor is unchanged and **now carries the whole weight**: requiring `-b` used to reject a CCC 8 *stable* filename as a side effect, and only the `7` does now. Both CCC 8 shapes still fail closed.

Ordering (`VersionComparator.tokenize`/`compare` ported): `isNewer("7.2", "7.1.7-b7") = true`, reverse false. **This pair is decided at the second component (1 vs 2)** and never reaches the padded-`.number(0)`-over-`.text` rule CotEditor's `7.1.0` / `7.1.0-beta.6` depends on; that rule only carries a future `7.3` / `7.3-b1` shape, recomputed separately as true.

## Tests

`betaPatternDoesNotMatchAStableFilename` asserted the bug, so it is replaced rather than kept. Each new case with the mutation it is written against:

- `betaPatternReadsTheGraduatedStableFilenameOnPurpose` — the graduated filename reads `7.2`, an open cycle still reads `7.1.7-b7`, the historical two-segment form still reads `7.1`. The second `#expect` catches its own mutation: dropping `-b[0-9]+` entirely leaves the first passing while that one goes nil — measured, because the `\.[0-9]{3,}` that follows stops lining up with `-b7.8389`.
- `theGraduationOutranksTheLastPrereleaseOfItsCycle` — the ordering, asserted rather than assumed, both directions plus the future-cycle shape.
- `aBetaCopyIsOfferedTheBuildItsCycleGraduatedInto` — the user-visible bug end to end through `VendorProbeSource` (channel gate → `installedVersionPattern` → comparison). Stub-served so it keeps asserting between-cycles behaviour once the vendor opens the next cycle.
- `betaEndpointStillResolvesAVersionLive` — the live tripwire, written to survive the vendor changing state: it asserts *a* CCC 7 version resolves, not which one. Its doc comment states what it cannot see (a retired rail looks identical), rather than reading as a stronger guarantee than it is.
- `aFutureCCC8ArtifactResolvesToNothingThroughTheCCC7Recipes` gains the beta-against-a-CCC-8-**stable**-filename half (newly reachable now that the major anchor is the only guard) plus a positive control from the same stub.
- The registry-derived matrix gains the graduated filename and widens the beta row — every CCC 7 filename readable through the beta pattern, no CCC 5/6/8 one.

## Verification

`make test` and `duo verify` cannot run in the session that wrote this — it is Linux, there is no Swift toolchain, and `DuoUpdaterCore` is `platforms: [.macOS(.v14)]`. What was run locally: the live endpoint probes and release-notes fetches above, the Python recomputation against those responses, and the Python gates from `make test` (`check_app_audits`, `check_prose_claims`, `check_offpool`, `check_staged_version_use`, `check_engine_notes`, `check_skill_docs` — all pass).

The Swift halves were verified by CI on `811231f` (pre-rebase), which ran green and, checked against the job log rather than the check mark, actually executed the new cases:

```
Test run with 2997 tests in 258 suites passed (36.7s)
betaPatternReadsTheGraduatedStableFilenameOnPurpose        started=1 passed=1
theGraduationOutranksTheLastPrereleaseOfItsCycle           started=1 passed=1
aBetaCopyIsOfferedTheBuildItsCycleGraduatedInto            started=1 passed=1
betaEndpointStillResolvesAVersionLive                      started=1 passed=1
aFutureCCC8ArtifactResolvesToNothingThroughTheCCC7Recipes  started=1 passed=1
```

The live one passing means CI reached bombich.com and resolved a 7.x through the fixed pattern. CI re-runs on the rebased head.

**`verify/baseline.json` is deliberately untouched** — it is the sweep's output, and hand-editing it would claim a result this session did not produce. The next healthy sweep clears `vendor:com.bombich.ccc:beta` on its own.

## Rebase onto #608

#608 has merged, and it rewrote this family's beta comment (dated history moved to `docs/app-audits/com-bombich-ccc.md#历史与实测`, contract left in code). This branch is rebased onto `da53817`; both conflicts were in that comment block and were resolved toward #608's shape — its trimmed opening and its audit-pointing CCC 5/6 fixture doc are kept, and the dated `?v=ccc7beta` probing it moved out is not restated in code.

One thing could not be kept as #608 wrote it: it describes the failure in the **present** tense ("the probe throws `ProbeFailed`, so the row shows a failed check"), which is what this PR stops being true. It is rewritten as the old behaviour, per CLAUDE.md's rule that a correction has to carry the statements it corrects. Its "likely whenever none is" hedge is also gone, since the 2026-09-14 reading measured it.

## Not changed

Only the CCC beta recipe. No other recipe, no `ReleaseChannel` rule, no `ChannelProofRegistry` entry, no `CHANNEL_COVERAGE_TODO.md` (its 2026-08-29 notes are accurate for their date), no `CHANGELOG.md` (this repo writes release notes at bump time, across seven language files, not per feature PR).

https://claude.ai/code/session_01RT675xfUMwv8Yv8JoaAPuD


jizhi0v0:
`/code-review` round (CLAUDE.md's every-PR rule), pushed as `811231f`. **Comments and docs only — no executable line changed**, so the regex, the tests and their assertions are exactly what CI already ran green on `c5f5470`.

Two costs the original cost list did not name. Both are now in the registry comment and the audit rather than left for the next person:

1. **A retired beta rail is now invisible.** Requiring `-b` had a side effect I hadn't accounted for: if Bombich drops or renames `?v=latestbeta`, the answer stops being a beta filename and the probe goes red, so somebody looks. It now resolves the stable zip and stays **green forever** — "resting between cycles" and "the rail is gone" are the same bytes, and nothing else here would notice (no `install`, so no proof questions it; the cask is `auto_updates: true`, so no Homebrew cross-check). `trackClosedPattern` is not the fix: it wants the vendor to *say* the track is empty, and this endpoint says nothing — it just hands you a different file. Accepted knowingly, the same trade CotEditor's beta rule makes by accepting a stable tag.
2. **A copy ahead of the graduation reads a lower "latest".** If a cycle closes back to a stable older than the installed prerelease (on `7.2.1-b1`, endpoint answers `7.2`), no update is offered — correctly — but the row names a smaller number and `RecipeSanity.remoteBehindInstalled` starts raising an advisory.

On (2) the review said this window had just happened, 2026-08-18 → 09-13. **Checked rather than forwarded, and it did not**: `verify/baseline.json` has `vendor:com.bombich.ccc:beta` at `lastGoodAt 2026-09-13T14:24Z` / `lastGoodVersion 7.1.7-b7`, while stable ccc7 only reads `7.2` in the 20:25 sweep the same day — the endpoint served `-b7` right up to 7.2's release, so no fallback occurred in that window. The beta *notes page* froze on 2026-08-18; the *endpoint* did not. The mechanism is real, the cited instance is not, and it is recorded that way.

`betaEndpointStillResolvesAVersionLive` now states what it **cannot** see rather than reading as a stronger tripwire than it is: `hasPrefix("7.")` is satisfied by the stable fallback, so it cannot tell a retired rail from a resting one, and it would also pass if the recipe's `url` were repointed at `?v=ccc7`. Neither is fixable by a stronger assertion — it is the same structural gap as (1), and the URL is pinned by `betaRecipeExistsAndUsesTheLatestbetaEndpoint`, not here.

The review also found a pre-existing contradiction sitting next to the rows I edited: **three** places in the audit still described the CCC 7 stable recipe as reading `?v=latest` — what the 2026-08-29 generation split replaced with `?v=ccc7`, and what `stableRecipeUsesTheGenerationScopedEndpoint` now forbids outright. Counted before changing, per CLAUDE.md's "count the copies first"; the 2026-08-30 measurement that was actually taken through `?v=latest` keeps its own wording, since both resolved the same file that day.

Nothing else survived: no issues in the regex, the version ordering, the channel/host/installed gates, the stub-server tests, or the matrix test's exhaustiveness.

---

PR #654, opened 2026-09-15
recipes step 3 S2: load .json5 recipe families from a resource bundle; convert 1Password

jizhi0v0:
Step 3, slice S2: a spike that converts exactly one recipe family, `com-1password-1password`, to a `.json5` resource file loaded at runtime, and makes every checker, test and install route understand it. Goldens are unchanged. **Verdict at the bottom: (a), the resource bundle, with no blocker found. Five items still need a real signed build, listed in the coordinator checklist.**

## Round 4: the reproduced Bundle.main bypass, and four smaller items

Two commits, `d1f5db62` and `17bc5436`, on top of the round-3 HEAD the coordinator checked (`c7c3378a`: `make cli` passed including the self-test, one `.ips`, no staging left, `duo verify --only 1password` 2 ✓).

### 1. HIGH — the digest is read from the executable's own section, not `Bundle.main`

The bypass, reproduced here with a swift-built binary: `Bundle.main.object(forInfoDictionaryKey:)` returned a sidecar `Info.plist`'s value (`ATTACKER-WINS`) while `getsectiondata` on the main image returned the embedded one (`EMBEDDED-abc123`). For a bare command-line executable `Bundle.main` treats the executable's own (writable) directory as the bundle, so a dropped `Info.plist` — or `Contents/Info.plist`, `Resources/Info.plist`, `en.lproj/InfoPlist.strings` — wins over `__TEXT,__info_plist`.

**Fix:** `RecipeFamilyFile.embeddedDigest()` reads the section straight from the main executable's image, and `AppRecipeIndex.dataFamilies` passes that instead of `Bundle.main.object(...)`.

**Main-image identification — citation.** I take the header from `dlsym(RTLD_MAIN_ONLY, MH_EXECUTE_SYM)`, not `_dyld_get_image_header(0)`:
- `dlsym(3)` (and `<dlfcn.h>`): *"RTLD_MAIN_ONLY … searches only the main executable."* `RTLD_MAIN_ONLY` is `(void *)-5`.
- `<mach-o/ldsyms.h>`: the symbol `_mh_execute_header` (`MH_EXECUTE_SYM`) *"is the address of the mach header in a Mach-O executable file type. It does not appear in any file type other than a MH_EXECUTE file type."*
- `<mach-o/dyld.h>` documents `_dyld_get_image_header(index)` only as "the mach header of the image indexed", with no statement that index 0 is the main executable — which is why the reviewer flagged it and I did not use it.

Then `getsectiondata(header, "__TEXT", "__info_plist", &size)` (`<mach-o/getsect.h>`) and `PropertyListSerialization`.

**"No key → skip" preserved.** No section, or a section without `DuoRecipeDigest`, reads as nil and skips. Verified with `otool -s __TEXT __info_plist` on the built app: the app binary has **no** such section (the helper, which sets `CREATE_INFOPLIST_SECTION_IN_BINARY`, does). The app's Info.plist is a file in `Contents/`, not a section, so the app reads no key — as before.

**Seam:** `digest(fromInfoPlistSection:)` takes `Data`, for the unit test; `dataFamilies(in:bundle:expectedDigest:)` still takes the digest, for the exit tests.

**Gate.** `build-cli.sh`'s self-test now plants, on the tampered run, all of these at once, each naming the tampered digest or none, and the binary must still trap with `are not the ones this executable was built with`:
- a sidecar `Info.plist`, a `Contents/Info.plist`, a `Resources/Info.plist` (each `CFBundleExecutable=duo-cli`);
- an `en.lproj/InfoPlist.strings` setting `DuoRecipeDigest`;
- `CFPROCESSPATH` pointed at a decoy.

**Mutation (real-mode harness, real swift-built `duo`):**

| Run | Result |
|---|---|
| Fixed tree (`embeddedDigest`) | exit 0; installed `drwxr-xr-x`, stamp written, no staging left; one `.ips` |
| `expectedDigest: Bundle.main.object(...)` restored, rebuilt | **exit 1 at the self-test, before install**: `✗ self-test: with a tampered recipe file the binary exited 2 but stderr lacks "are not the ones this executable was built with" — is AppRecipeIndex.dataFamilies still reading the embedded __info_plist digest (not Bundle.main)?`. No `.ips` (no trap), scratch libexec empty |
| Reverted, rebuilt | exit 0 |

The mutant exits 2 ("nothing to verify") on the tampered run because the sidecar's tampered digest matched the tampered data — exactly the accepted-tampered-data bypass, which is why the fixed binary must trap instead.

**Unit tests.** `theDigestIsParsedFromAnInfoPlistSection` drives the parser (key present, empty, absent, wrong type, non-plist). `embeddedDigestIsNilInTheTestProcess`: in `swift test` the digest is nil (the test runner's executable carries no key) and the check is skipped — which is why a test process could never catch this bypass, and the gate for it is in `build-cli.sh`. It cannot prove the positive case (a section whose key is read); the real-mode self-test does.

### 2. NFC duplicate keys

`check_recipe_json5.py`'s `pairs` compares keys after `unicodedata.normalize("NFC", key)`, because Swift `String` hashing is on canonical equivalents, so a `[String: …]` table keeps one of `"com.example.Keka"` with U+212A KELVIN vs ASCII K, or decomposed vs precomposed `café`. The message names both spellings and says "NFC". Tests: `test_canonically_equal_keys_are_refused` (Kelvin and café, both distinct byte strings). Mutation "compare byte-for-byte, not NFC" → red.

### 3. Global `$STAMP.new`

The stamp is written through `STAMP_NEW="$(mktemp "$STAMP.XXXXXX")"`, cleaned only when set, not a fixed `"$STAMP.new"`. `$STAMP` is global (`~/.local/libexec/duo.built-from`), so a failing `make cli` in another worktree used to delete a concurrent run's `.new` on its own EXIT trap, and that run then died on a bare `mv` after it had already installed. No other trap path is a fixed name (`BUNDLE_TMP`, `TMP`, `SELFTEST` are all `mktemp`).

### 4. Two under-discriminating tests

- Added a `.zz.json5` stray fixture (leading-dot slug with the right extension), in both the Swift `aStrayEntryIsRefused` and the Python stray test. The `.DS_Store` case was refused on its extension, so it never exercised the slug's leading-char rule.
- `recipeFilesThatDoNotMatchTheBuiltDigestAreRefused`: dropped the "compare before reading" mutation from the doc — no single-threaded test can stage a writer between the hash and the decode — and said the read-once property rests on the single read in `loadAll`.

### 5. Prose

- `AppRecipeIndexTests`' symlink-refusal comment was wrong ("a regular file when followed"). Measured `URLResourceValues` on a symlink to a regular file: `isSymbolicLink=true`, `isRegularFile=false` — it does not follow the link. The comment now says that, and the test still pins that the reason given is "a symbolic link".
- Both fragile-recipe mirrors: the escaping caveat now also covers a `"""…"""` literal (not raw: `\` is an escape, a trailing `\` joins lines), pointing at `com-google-GeminiMacOS.swift:42`.

### `make test` (round 4)

```
$ scripts/run-with-hang-report.sh 1800 make test > make-test-5.log 2>&1; echo exit=$?
exit=0
━ Test run with 3101 tests in 267 suites passed after 22.621 seconds with 1 known issue.
✔ Test run with 352 tests in 44 suites passed after 6.398 seconds.
✓ recipe .json5 files in dialect, no duplicate keys — 1 files; family files reconcile with 225 goldens
(every other checker ✓)
```
One green run, first try. Checker mutations: 22 of 22 red (the 21 from round 2 plus the NFC one).

### Coordinator note

`make cli` behaves as in round 3 (one new `duo-cli-*.ips`, no staging left). The self-test now also plants the sidecar `Info.plist` attack on its tampered run; it is removed with the staging directory.
---

## Round 3: a gate for S7, and a stale comment

One commit, `c7c3378a`. The coordinator had checked round 2 on `1f058bdf` (`make test`, `make cli`, embedded digest == `recipe_digest.py`, tamper → 133, restore → 2 ✓, `zz-extra.json5` → 133).

### 1. `build-cli.sh` self-test: the built binary must refuse recipe data it was not built with

It runs after the build products are verified and before anything is installed:
1. `mktemp -d "${TMPDIR:-/tmp}/duo-cli-selftest.XXXXXX"`; the EXIT trap removes it.
2. The staging directory gets `cp -p $PRODUCT` plus `ditto $PRODUCT_BUNDLE`, and a `home/` directory.
3. The staged binary runs twice: from cwd = staging, with `HOME`/`CFFIXED_USER_HOME` = staging `home/`, and `PACKAGE_RESOURCE_BUNDLE_PATH`/`_URL` unset:
   ```
   ./duo-cli verify --only zz-duo-recipe-digest-selftest
   ```
   - **untouched** → must print `nothing to verify — no recipe matches zz-duo-recipe-digest-selftest`. That shows the staged binary found its bundle and the digest matched.
   - **one byte appended to a family file** → must exit non-zero with `are not the ones this executable was built with`.
   - Otherwise `die`, naming the missing line and quoting the last 3 lines.

**Why this command:**
- `CLI/Sources/duo/main.swift` only parses flags before calling `Verify.run`, and `verify` is a known subcommand with `--only` a known flag.
- `Verify.run` (`CLI/Sources/DuoKit/Verify.swift:89`) first calls `SourceStamp.verdict()`. That walks up from cwd looking for `DuoUpdaterCore/Sources` + `CLI/Sources`; the staging directory is not in a checkout, so it returns `.notApplicable` without reading anything else.
- The next statement is `filtered(VendorProbeRegistry.recipes, options)`, the first access to `AppRecipeIndex.all`.
- With nothing matching, `guard total > 0 else { die("nothing to verify …", code: 2) }` exits before `installedVersions()` or any sweep, so there is no network. `EventStore`/`GitHubConditionalCache` flush only if used, and neither is.
- Measured with the real binary: **0 files written** under the staged `home/`.

**Crash reports:** each passing run leaves exactly one `~/Library/Logs/DiagnosticReports/duo-cli-<timestamp>.ips`, from run 2's deliberate trap; the script comment says so. The count, from the harness in real mode, taken from the real `~/Library/Logs/DiagnosticReports`:
- passing run: before 1 → after 2;
- passing run after reverting the mutant: 3 → 4;
- the mutant run: 3 → 3 (no trap, so no report).

**Raw output of the two runs** with the real swift-built `duo`:
```
== untouched
exit=2
nothing to verify — no recipe matches zz-duo-recipe-digest-selftest
== one byte appended
exit=133
DuoUpdaterCore/AppRecipeIndex.swift:42: Fatal error: recipe data: the recipe files in …/duo-cli-selftest-show.KM5RtH/DuoUpdaterCore_DuoUpdaterCore.bundle/Contents/Resources/Recipes are not the ones this executable was built with (built with e37cd8efefc0782b7693f0cdfe84a3b5934dab98f1646ddd0760d20cfa8e668b, found ffddabfc…). Reinstall with `make cli`; …
== files written under staging home: 0
```

### Mutation verification, with the real loader

**Real mode.** The harness gained `MODE=real`. There, the shimmed `xcodebuild` runs `swift build -c release --product duo` of the worktree's real `CLI` package, linked with `-Xlinker -sectcreate __TEXT __info_plist <plist with the digest>`. It copies the real `duo` and its real `DuoUpdaterCore_DuoUpdaterCore.bundle` into the products directory. The scratch repo now carries copies of `DuoUpdaterCore/Sources` and `CLI/Sources`, so the real `verify --source-digest` stamps work. `codesign`/`xcodegen` stay shimmed; `check_swift_backdeploy.py` and `launchctl plist` run for real on the real Mach-O. Release SwiftPM `duo` passed the back-deploy check.

| Run | Result |
|---|---|
| Unmodified tree, `scripts/build-cli.sh` at `c7c3378a` | **exit 0**; installed bundle `drwxr-xr-x`, stamp written, no leftover staging |
| `expectedDigest: nil` in `AppRecipeIndex.dataFamilies` (applied to the worktree, rebuilt) | **exit 1 before install**: `✗ self-test: with a tampered recipe file the binary exited 2 but stderr lacks "are not the ones this executable was built with" — is AppRecipeIndex.dataFamilies still passing the Info.plist digest? Last lines: nothing to verify — no recipe matches zz-duo-recipe-digest-selftest`. Scratch libexec empty, no staging leftover |
| Mutation reverted (`git checkout --`), rebuilt | **exit 0** again |

**Fake mode.** The swiftc stand-in, now answering `verify --only` against its own embedded digest, still passes scenarios 1–9 with the same outcomes as round 2. The new scenario 10 (stand-in ignores its digest) fails at the self-test. Fake mode says nothing about the real loader; real mode is the evidence.

### 2. Stale name

`scripts/recipe_families.py` now says "Must match `RecipeFamilyFile.isFamilyFileName`".

### `make test` (round 3)

```
$ scripts/run-with-hang-report.sh 1800 make test > make-test-4.log 2>&1; echo exit=$?
exit=0
━ Test run with 3099 tests in 267 suites passed after 20.904 seconds with 1 known issue.
✔ Test run with 352 tests in 44 suites passed after 6.272 seconds.
✓ recipe .json5 files in dialect, no duplicate keys — 1 files; family files reconcile with 225 goldens
(every other checker ✓, as in round 2)
```

### Coordinator: additional expectation for `make cli`

`make cli` now prints `→ Verifying the binary refuses recipe data it was not built with` and passes. It adds one `duo-cli-*.ips` to `~/Library/Logs/DiagnosticReports`, and leaves no `duo-cli-selftest.*` directory in `$TMPDIR`.

---

## Round 2: review fixes

Round 1 was checked by the coordinator on `c31cfe2c`, with no failures: `make test`, `make cli` + `duo verify --only 1password`, `make install` + `codesign --verify --deep --strict` (and a tampered `.json5` in a copy of the app failed verification), `make gallery`, and a real notarization. Round 2 is four new commits on top of that (no amend, no force-push):

| Commit | Items |
|---|---|
| `952798c0` | 1, 5: checkers |
| `0d9059d6` | 2, 5, 9: loader, digest, plist |
| `9f8af58a` | 2, 4: `build-cli.sh` |
| `1f058bdf` | 3, 6, 7, 8: docs |

`origin/main` moved (#644, #645: Homebrew and Sparkle source code, no recipe types or families), so there was no merge and no re-record.

### 1. `\r` and friends hiding data from `check_recipe_json5.py`

**Measured first.** I fed Foundation's JSON5 reader directly (`swift json5probe.swift`) with this line:
```
{"sparkleFeeds": { // note<SEP>"com.x": "https://evil…", "com.x": "https://good…" }}
```
Results per separator:
- `CR`: decoded `evil`, so the comment ends at `\r`.
- `LS U+2028`, `PS U+2029`, `NEL U+0085`, `VT`, `FF`: all decoded `good`, so the comment runs on to `\n`.

And per character placed before `//`:
- `NBSP`, `\x1c`, `\x1f`, `U+2003`, `BOM`: each gives `ERROR Unexpected character … at beginning of JSON5 unquoted key`.
- `VT`, `FF`: skipped as whitespace.

Escapes:
- `\u0000` escape: `ERROR Unsupported escaped null`.
- Lone `\uD800`: `ERROR Missing low code point in surrogate pair`.
- Lone `\uDC00`: `ERROR Invalid unicode scalar value`.
- Paired surrogates: decoded.

**Fix** in `check_recipe_json5.py`:
- A comment line is exactly `^[ \t]*//`.
- Refused anywhere in the file: `\r`, every other C0/C1 control except tab and newline, DEL, U+2028, U+2029, U+FEFF.
- Refused in a line's indentation: any whitespace other than space and tab.
- Refused after decoding: any string or key holding NUL or a lone surrogate. This was cheap: a walk of the decoded value. Python accepts both escapes; Foundation refuses both, so without this the file would pass here and trap at runtime.

The reviewer's exact input now gives `[control-character]`; the same input with `\r` → `\n` gives `[duplicate-key]`.

### 2. Recipe data beside a granted `duo`: closed for `make cli` builds, as proposed

**Mechanism:**
- `scripts/build-cli.sh` computes `scripts/recipe_digest.py` over the checkout's `Resources/Recipes` before building: SHA-256 over each file in name order of `"<name>\n<byte count>\n"` + bytes.
- It passes the result as `DUO_RECIPE_DIGEST`.
- `App/project.yml` gives `duo-cli` `INFOPLIST_FILE: duo-cli-Info.plist`, a partial plist holding `DuoRecipeDigest = $(DUO_RECIPE_DIGEST)`. It is merged with the generated plist, which `CREATE_INFOPLIST_SECTION_IN_BINARY` embeds in `__TEXT,__info_plist`.
- At load, `AppRecipeIndex.dataFamilies` passes `Bundle.main.object(forInfoDictionaryKey: "DuoRecipeDigest") as? String` to `RecipeFamilyFile.loadAll`.
- `loadAll` reads each file once, hashes those bytes, and decodes the same bytes, so there is no window between check and decode.
- A mismatch traps with `the recipe files in <dir> are not the ones this executable was built with (built with <a>, found <b>). Reinstall with make cli; …`. An empty key traps too ("built without DUO_RECIPE_DIGEST").
- No key (the app, `swift test`, `swift build`) skips the check; the app's copy is sealed by its signature (your measurement).

**Evidence for the mechanism.** I could not build duo-cli, so I used a scratch XcodeGen `type: tool` target with duo-cli's plist settings and signing off (`.build/s2/plistspike`, not committed, not the app/cli/gallery):
- Built with `DUO_RECIPE_DIGEST=abc123`: the binary prints `DuoRecipeDigest = "abc123"` via `Bundle.main`, also from a copy elsewhere. `launchctl plist __TEXT,__info_plist` shows the key.
- Without the setting: `DuoRecipeDigest = ""`.
- An incremental rebuild changing only the setting: the new value lands.
- The real `App/duo-cli-Info.plist`, XML comment included, merges.
- Ad-hoc signed, then one byte of the digest patched: `codesign --verify` gives `invalid Info.plist (plist or signature have been modified)`, and running it exits 137 (killed).
- The generated project has `INFOPLIST_FILE = "duo-cli-Info.plist"` for both duo-cli configurations, and still has `DEVELOPMENT_TEAM` (8 matches).

**Two implementations of one hash.** Swift (`RecipeFamilyFile.digest`) and Python (`recipe_digest.py`) are both pinned to one known answer, `8fdc639d…8692`, for the same two-file fixture. The answer was also computed by `printf 'a-app.json5\n3\n{}\nb-app.json5\n20\n…' | shasum -a 256`.

**Tests:**
- `theRecipeDigestHasOneKnownAnswer`.
- `recipeFilesThatDoNotMatchTheBuiltDigestAreRefused`: the right digest loads, an empty one throws, and a one-byte change traps in an exit test whose stderr names `built with <a>, found <b>`.
- `scripts/test_recipe_digest.py`: 6 tests, in `make test`.

**Not covered by any test:** mutating the `Bundle.main` read to `expectedDigest: nil` leaves `swift test` green (mutation S7 below). A test process carries no key; only your `make cli` check sees that line.

**Also stated in `RecipeFamilyFile`'s doc:** Debug builds honour `PACKAGE_RESOURCE_BUNDLE_PATH` for the bundle location. The digest check is on content, so a redirected bundle is still checked in any binary that carries the key.

### 3. The app-audit skill search

`.claude/skills/app-audit/SKILL.md` now greps `--include="*.swift" --include="*.json5"` and lists the family file in either directory. For `com.1password.1password`:
- `origin/main`, Swift only: 9 hits.
- This branch, Swift only: 1 (the Dropbox comment).
- With `.json5`: 6.

The 3 still missing are Swift scaffolding: the `AppRecipeIndex` line, `enum com_1password_1password` and `family:`.

A repo-wide grep of `.claude`, `.agents`, `CLAUDE.md`, `README.md`, the coverage ledger and the audit/engine-notes READMEs for `*.swift`, `-t swift` and `Recipes/*.swift` found only this, the channel-discovery awk (paired with the `.json5` command in round 1) and the ledger command (fixed in round 1).

### 4. `build-cli.sh` order

**New order:**
1. Digest the checkout.
2. Build with the digest.
3. Signature and back-deploy checks, as before.
4. **Before installing anything:** the embedded digest (`launchctl plist`) and the built bundle's `Recipes` digest must both equal the checkout digest.
5. Stamp S1, taken from the built binary.
6. The checkout digest must still equal the build digest.
7. Stage the bundle (`chmod 755`, since `mktemp -d` gives 0700) and the binary.
8. Swap back to back: remove the old bundle, rename the new one in, rename the binary in.
9. `codesign --verify`, and the installed bundle digest.
10. Stamp S2 must equal S1, and only then is it written.

A `trap cleanup EXIT` removes both staging paths and `duo.built-from.new`.

**Race choice.** Stamp first, then recheck the recipe digest: an edit before S1 changes the checkout digest and fails at step 6. Stamp again after installing: an edit after step 6 changes S2 and fails at step 10. That is ordering rather than a lock, and it does not catch an A→B→A edit that returns to identical bytes.

**Harness** (`.build/s2/clih/run.sh`, not committed): runs the real `scripts/build-cli.sh` from a scratch copy of the files it reads.
- Shimmed: `xcodebuild` (compiles a `swiftc` stand-in `duo` with the real embedded `__info_plist` section, copies recipes into an Xcode-layout bundle), `xcodegen` and `codesign`.
- Scratch: `HOME` and `DERIVED_DATA`.
- It asserts the shims are first on `PATH`; nothing real is built, signed or installed.

| Scenario | This PR | Round-1 script (`c31cfe2c`) |
|---|---|---|
| 1 clean install | exit 0, bundle `drwxr-xr-x`, installed digest == checkout, embedded digest `e37cd8ef…` | exit 0, bundle **`drwx------`**, no embedded digest |
| 2 reinstall over a bundle holding `zz-deleted.json5` | exit 0, only `com-1password-1password.json5` remains | same |
| 3 binary embeds a stale digest | **exit 1 before install**: `the binary carries recipe digest 'stale0000', not e37cd8…`; libexec empty | exit 0 (nothing checks) |
| 4 built bundle differs from checkout | **exit 1 before install**; libexec empty | exit 1 **after the bundle was swapped in** (bundle left beside no binary) |
| 5 recipe edited just before stamp S1 | **exit 1**: `… changed during the build — run make cli again`; libexec empty | **exit 0, stamp describes the edited tree** |
| 6 recipe edited after install, before S2 | **exit 1**: `the sources … changed while installing …`; no stamp written, so `duo verify` refuses | exit 0 |
| 7 installed copy fails `codesign --verify` | exit 1, previous stamp unchanged | same |
| 8 previous bundle cannot be removed (fails mid-swap) | exit 1, **no staging leftovers** | exit 1, leaves `DuoUpdaterCore_DuoUpdaterCore.bundle.xqjlGs` |
| 9 `.DS_Store` in the checkout's `Recipes` | exit 1 at the first digest; libexec empty | exit 0 |

Each of these mutants of the new script changes its scenario's outcome:
- **Drop `chmod`:** 0700 again.
- **Drop the trap:** leftovers `bundle.QRWp1p`, `duo.ob4ZuR`, `duo.built-from.new`.
- **Drop the step-6 recheck:** scenario 5 exits 0.
- **Drop the S1 == S2 check:** scenario 6 exits 0.
- **Drop the embedded check:** scenario 3 exits 0.
- **Drop the built-bundle check:** scenario 4 dies only after the swap.

### 5. Stray entries and counts that reconcile

**Loader** (`RecipeFamilyFile.readFamilyFiles`): lists every entry, hidden ones included. Anything but a regular (non-symlink) file named `[A-Za-z0-9][A-Za-z0-9.-]*\.json5`, exactly, fails with `recipe directory … holds entries that are not family files, which would be shipped and never loaded: <name> (<reason>)`.

**Checkers:**
- New `scripts/recipe_families.py` gives `check_recipe_json5`, `check_recipe_snapshots`, `check_app_audits` and `check_engine_notes` one definition of a family file.
- `check_recipe_json5` refuses the same entries the loader does.

**Invariant chosen:** the goldens. `RecipeGoldenTests` writes one per family `AppRecipeIndex.all` loads and holds the two sets equal, so the golden count is an independent count of loaded families. Each checker's family count must equal it:
- `check_recipe_json5`: `.swift` + `.json5` files;
- `check_recipe_snapshots`: guarded families;
- `check_app_audits`: family files scanned;
- `check_engine_notes`: Swift family files + `.json5` scanned.

This replaces the floors of 1. It needs a new witness when the goldens retire at S9; the module docstring says so.

**`.DS_Store` measured:** SwiftPM (Swift Build) **does not copy `.DS_Store`** into the bundle but **does copy other dotfiles** (`.hidden-other` was shipped). So a Finder `.DS_Store` would not crash a SwiftPM build; `.hidden-other` does. Xcode's copy was not measured. `make test` refuses both.

### 6. D7 and History

The two `### Recipes/com-1password-1password.swift — …` headings in the audit's `## 历史与实测` are restored verbatim. Only the present-tense line 3 keeps the new path; `git diff origin/main -- docs/app-audits/com-1password-1password.md` is now that one line. `docs/app-audits/README.md` says existing provenance headings are not rewritten when a family changes format.

**D7 note for bulk conversion:** rewrite present-tense path references; never rewrite provenance headings inside `## 历史与实测`.

### 7. "Compiled in"

`CLAUDE.md:25` and `CLI/Sources/DuoKit/Verify.swift:86` are fixed. Re-grepping `编译进|compiled in|compiled into|compiled-in|baked in(to)?` leaves only statements about other things: other apps' binaries, Tauri, `MASHelperProtocol.swift`, the S1 SourceStamp explanation that json5 is *not* compiled in, and History.

### 8. Escaping

Both fragile-recipe mirrors now carry the raw-string caveat: `\#(…)`, `\#n`/`\#t`/`\#"`, `##"…"##`. Write the value Swift produces; the golden shows it. They appear in `com-lemon-lvoverseas` (3 lines), `dev-kdrag0n-MacVirt` (2) and `com-windscribe-client` (1). The same section now also states the control-character rule and the family-files-only directory rule.

### 9. Test holes

- The no-`Recipes`-directory trap has its own exit test (`dataFamilies(in: nil, …)` → `recipe data: no \`Recipes\` directory in /fixture/X.bundle`).
- Exit tests now receive fixture paths from the parent process (typed captures) and the parent removes them. After the run there were no `duo-recipe-fixture-*` directories in `$TMPDIR`. I deleted the round-1 leaks: 5 `duo-recipe-empty-*` and 5 `duo-recipe-trap-*`.
- The channel-discovery command, run on the tree plus a fixture family, prints exactly the `probes`/`githubRules` channel entries and ignores `changelogs`:
  ```
  --- on the tree plus a fixture:
  …/Resources/Recipes/zz-probe-cmd.json5: probes zz.a channel: beta
  …/Resources/Recipes/zz-probe-cmd.json5: githubRules zz.b channel: nightly
  --- on the tree alone:
  (end, exit=0)
  ```
  The fixture held `probes: [{zz.a, beta}, {zz.a}]`, `githubRules: [{zz.b, nightly}]` and `changelogs: [{zz.a, nightly}]`.

### Load cost and threading (not asking for a change)

Debug build, scratch probe, three runs:
- `readFamilyFiles` on today's 1 file: 0.4 ms. Digest: 0.2 ms.
- A synthetic directory of 158 copies of the 1Password file (780,520 B, today's total family bytes): read 13.1–13.8 ms, digest 0.57–0.63 ms, **read + decode 23.8–24.3 ms**. This is a projection with one file's shape repeated, not a real family mix.

**Threading in the app, read not traced exhaustively:**
- The first scan runs in `Task.detached` (`App/Sources/AppListModel.swift:2551`).
- `AppScanner.readApp` consults `SparkleFeedCatalog.feed` (`Scan/AppScanner.swift:577`), so in the normal launch path the index's `static let` is first touched **off the main thread**, on a cooperative-pool thread, blocking it for the load.
- Main-actor sites exist (`AppListModel.swift:1790` `ChangelogRecipeSelection.recipe`, `:1905` `ChangelogRecipeRegistry.recipes`), but they render or invalidate results that the scan has already produced.

I did not prove that nothing on the main actor reaches a registry before the first scan.

### Round 2 mutations

**On the real tree**, each reverted:

| # | Mutation | Result |
|---|---|---|
| R1 | Inside the probe object: `// highest\r"selectHighest": false,` before `"selectHighest": true,` | **new checker red:** `…json5:50:17: [control-character] U+000D`. **Round-1 checker (`c31cfe2c`): exit 0**, `✓ … no duplicate keys`. Golden red at line 23: the hidden `false` is what Foundation kept |
| R2a | `zz-other.JSON5` beside the family | checker `[not-a-family-file] is not named <slug>.json5`; `swift test` traps `holds entries that are not family files … zz-other.JSON5 (not named <slug>.json5)` |
| R2b | `zz-other.json` | same pair |
| R2c | `sub/zz-deep.json5` | checker `…/sub: [not-a-family-file] is not a regular file`; loader trap names `sub (not a regular file)` |
| R2d | symlink `zz-link.json5` → the family | checker `is a symbolic link`; loader trap `zz-link.json5 (a symbolic link)` |
| R2e | `.DS_Store` | checker red; **`swift test` green, because SwiftPM did not copy it** (see 5) |
| R2f | `.hidden-other` | checker red; loader trap `.hidden-other (not named <slug>.json5)` |
| R3 | Extra golden `zz-stray.txt` | `check_recipe_json5`: `224 .swift + 1 .json5 family files: 225, but … holds 226 goldens`; `check_recipe_snapshots`: `guarded recipe families: 225, but … 226`; `check_app_audits` and `check_engine_notes`: the same shape; golden: `orphan golden zz-stray.txt` |

**Swift loader mutants** (`AppRecipeIndexTests` + `RecipeGoldenTests`, each reverted):

| # | Mutant | Red |
|---|---|---|
| S1 | Drop the symlink branch | `aStrayEntryIsRefused(symlink)`: still refused as "not a regular file" (URL resource values do not follow links), so the test pins the reason |
| S2 | Case-insensitive `.json5` | `aStrayEntryIsRefused` |
| S3 | Refusal guard disabled (skip strays) | `aStrayEntryIsRefused` |
| S4 | Skip the digest comparison | `recipeFilesThatDoNotMatchTheBuiltDigestAreRefused` |
| S5 | Digest without the name | `theRecipeDigestHasOneKnownAnswer` |
| S6 | No `Recipes` directory → `[]` | `aDataFamilyProblemTraps` |
| S7 | Ignore the embedded key (`expectedDigest: nil`), or read `Bundle.main` (round 4) | no unit test; **`build-cli.sh` self-test fails** (real-mode harness) |

**Checker mutants** (copies of the scripts):
- `check_recipe_json5.py` + `recipe_families.py`: **21 of 21 red**. That is the 10 from round 1 plus: `\r` allowed; FORBIDDEN narrowed to `\r`; `is_comment` back to `lstrip`; indentation unchecked; `string_problems` dropped; strays skipped; symlinks accepted; case-insensitive extension; dotfiles accepted; reconcile ignored; reconcile counts `.json5` only.
- `check_recipe_snapshots.py`: 3 of 3 red (Swift-only listing, reconcile removed, data directory existence).

### `make test` (round 2)

- **First run: exit 2.** `CarbonCopyClonerProbeRecipeTests` hit a vendor timeout (`ProbeFailure.transport(urlErrorCode: -1001, "The request timed out.")`) in two live tests.
  - The CCC family is Swift and untouched here.
  - `curl` to the same endpoints answered in 0.5 s.
  - The suite failed once more when rerun alone, then passed (`27 tests … passed after 2.448 seconds`).
- **Second full run** on `1f058bdf`:
```
$ scripts/run-with-hang-report.sh 1800 make test > make-test-3.log 2>&1; echo exit=$?
exit=0
━ Test run with 3099 tests in 267 suites passed after 20.356 seconds with 1 known issue.
✔ Test run with 352 tests in 44 suites passed after 6.375 seconds.
✓ localizable keys agree — 569 in the catalog, all reachable
✓ app audits consistent — 199 docs, … history pointers resolve (817 Swift files, 225 families, 1 of them .json5)
✓ no dated snapshots in recipe comments — 225 families guarded (1 as .json5)
✓ recipe .json5 files in dialect, no duplicate keys — 1 files; family files reconcile with 225 goldens
✓ engine-notes pointers resolve — 817 Swift files and 1 recipe .json5 files scanned, 3 doc(s) indexed
✓ skill docs consistent — 3 files mirrored, documented parameter counts (27 / 27) match the initializers
✓ no machine-population claims in code comments — 817 files
✓ no blocking calls left on the cooperative pool — 9 call(s) in 499 files
```

### Coordinator commands for item 2

Run from the repository root of this branch.

```sh
make cli
launchctl plist __TEXT,__info_plist ~/.local/libexec/duo | grep DuoRecipeDigest
python3 scripts/recipe_digest.py DuoUpdaterCore/Sources/DuoUpdaterCore/Resources/Recipes
ls -ld ~/.local/libexec/DuoUpdaterCore_DuoUpdaterCore.bundle
```
Expected:
- `make cli` passes the new step "Verifying the recipe data the build carries".
- The `grep` line shows a 64-hex digest identical to the `recipe_digest.py` output (today `e37cd8efefc0782b7693f0cdfe84a3b5934dab98f1646ddd0760d20cfa8e668b`).
- The bundle is `drwxr-xr-x`.

```sh
# 2a. tamper: append one byte to the installed family file
R=~/.local/libexec/DuoUpdaterCore_DuoUpdaterCore.bundle/Contents/Resources/Recipes/com-1password-1password.json5
cp "$R" /tmp/1p.json5.bak
printf ' ' >> "$R"
duo verify --only 1password; echo "exit=$?"
```
Expected: no sweep. The process dies with exit **133**, and stderr has `Fatal error: recipe data: the recipe files in …/Recipes are not the ones this executable was built with (built with e37cd8ef…, found <other>). Reinstall with \`make cli\`; …`.

```sh
# 2b. restore
cp /tmp/1p.json5.bak "$R"
duo verify --only 1password; echo "exit=$?"
```
Expected: passes as in round 1, with 2 ✓ and exit 0.

```sh
# 2c. (optional) the app carries no key, so the check is CLI-only
/usr/libexec/PlistBuddy -c "Print :DuoRecipeDigest" /Applications/DuoUpdater.app/Contents/Info.plist; echo "exit=$?"
```
Expected: `Print: Entry, ":DuoRecipeDigest", Does Not Exist` with a non-zero exit.

---

_Round 1 (below) is the original description. Where round 2 changed something, the round-1 text is struck through or annotated._

## Design

- **Resource.** `DuoUpdaterCore/Package.swift` declares `.copy("Resources/Recipes")`. Families live at `DuoUpdaterCore/Sources/DuoUpdaterCore/Resources/Recipes/<family>.json5`. They cannot sit beside the `.swift` families, because a `.copy` of `Recipes/` would take the Swift files out of compilation.
- **Hybrid index.** `AppRecipeIndex.all = (swiftFamilies + dataFamilies).sorted { lowercased slug }`.
  - `swiftFamilies` is the old hand-kept literal, renamed, and is still required to be sorted.
  - `dataFamilies` lists `Bundle.module`'s `Recipes` directory and decodes every `.json5` in it.
  - Both are `static let`, so loading happens on first access.
- **File format (`RecipeFamilyFile`).**
  - The top level is `AppRecipeSet`'s ten kinds, each optional and defaulting to empty.
  - Unknown keys and `null` are refused (S0 helpers). Proof tables use `ChannelProofTable`.
  - `JSONDecoder` + `allowsJSON5`.
- **Slug.** The file name is the slug. **The file does not carry `family`**, so there is no second copy to disagree with and no mismatch to catch. What stays checkable:
  - a renamed file is a renamed family (the golden reports a missing golden plus an orphan; mutation 7b);
  - a family in both forms is refused by `AppRecipeIndexTests.familySlugsAreTheFileNames`.
- **Traps (D4)** happen on first access, with `fatalError` naming the file and the coding path. Four cases:
  - bundle not found: the accessor's own trap, `unable to find bundle named DuoUpdaterCore_DuoUpdaterCore`;
  - no `Recipes` directory in the bundle;
  - a directory with **zero** `.json5` files (a floor);
  - any decode error.

  All four were reproduced with a swift-built probe (exit 133 each):
  ```
  alone:  DuoUpdaterCore/resource_bundle_accessor.swift:44: Fatal error: unable to find bundle named DuoUpdaterCore_DuoUpdaterCore
  broken: DuoUpdaterCore/AppRecipeIndex.swift:38: Fatal error: recipe data: Recipes/zz-broken.json5: probes[0].url: required key is missing
  empty:  … Fatal error: recipe data: recipe directory …/DuoUpdaterCore_DuoUpdaterCore.bundle/Contents/Resources/Recipes holds no .json5 family files
  nodir:  … Fatal error: recipe data: no `Recipes` directory in …/DuoUpdaterCore_DuoUpdaterCore.bundle. The resource bundle is incomplete; …
  ```
- **Duplicate keys** are not caught by the decoder, which keeps the first value. They are caught by `scripts/check_recipe_json5.py` in `make test` (below).

Commits:
1. loader + index
2. conversion + path rewrites
3. checkers + tests
4. packaging script + "compiled in" claims
5. docs
6. a checker message fix

Commit 1 alone does not build, because the resource directory arrives in commit 2.

## Goldens unchanged

```
$ git diff origin/main -- DuoUpdaterCore/Tests/RecipeGoldens | wc -c
       0
$ swift test --package-path DuoUpdaterCore --filter RecipeGoldenTests --filter AppRecipeIndexTests --filter RecipeCodableTests
✔ Test theRegistryMatchesTheGoldens() passed after 0.155 seconds.
✔ Test run with 28 tests in 3 suites passed after 0.419 seconds.
```

**Escaping.** The golden quotes every pattern with escapes, so an unchanged, passing `com-1password-1password.txt` means the three regexes decode byte-identical to the Swift raw strings. As a second witness that shares no code with the golden, a scratch script reads the Swift raw strings from `git show origin/main:…` and compares them with `json.loads` of the new file:
```
$ python3 patterns.py old.swift …/com-1password-1password.json5
5 raw strings in the Swift file
same bytes  probes[0].versionPattern  (59 B, 3 backslashes in Swift, 1 exact escaped occurrence in the file)
same bytes  changelogs[0].entryPattern  (156 B, 3 backslashes in Swift, 1 exact escaped occurrence in the file)
same bytes  changelogs[0].itemPatterns[0]  (33 B, 0 backslashes in Swift, 1 exact escaped occurrence in the file)
```

## Comments at the same positions

A one-off, uncommitted script (`comment_sequence.py`) extracts the ordered sequence of `(kind, entry index, entry-level or field, comment block text)` from the old Swift file and from the JSON5 file:
- A block is a run of whole-line `//` comments, taken as text from `//` to end of line.
- It is attached to the next non-comment line: an entry start, a field, or other.

```
$ git show origin/main:DuoUpdaterCore/Sources/DuoUpdaterCore/Recipes/com-1password-1password.swift > old.swift
$ python3 comment_sequence.py old.swift DuoUpdaterCore/Sources/DuoUpdaterCore/Resources/Recipes/com-1password-1password.json5
swift: 2 blocks, 60 comment lines: ('probes', 0, '<entry>') x40; ('changelogs', 0, '<entry>') x20
json5: 2 blocks, 60 comment lines: ('probes', 0, '<entry>') x40; ('changelogs', 0, '<entry>') x20
OK: identical (kind, entry, position, text) sequence
```
- **Field-level attachment** was checked on a synthetic pair: a comment before `selectHighest` in both files gives OK; before `changelogURL` in the JSON5 gives `FIRST DIFFERENCE … 'selectHighest' vs 'changelogURL'`.
- **History pointer.** `// History: docs/app-audits/com-1password-1password.md#历史与实测` is byte-identical from `//` to end of line. Only its indentation changed, from 8 spaces to 4.

## Path references rewritten (D7)

`grep -rn "1password-1password"` outside the goldens found four references to the Swift file. All now point at `Resources/Recipes/com-1password-1password.json5`:
- `DuoUpdaterCore/Sources/DuoUpdaterCore/Recipes/com-getdropbox-dropbox.swift:33`. This is a comment, not data; the Dropbox golden is unchanged.
- `docs/app-audits/com-1password-1password.md:3`, the opening line.
- ~~`docs/app-audits/com-1password-1password.md:9` and `:18`~~: **reverted in round 2.** These are provenance headings inside `## 历史与实测`, which are dated snapshots. D7 for bulk conversion: rewrite present-tense references, never the History provenance headings.

## Checkers and tests that now understand `.json5`

_Round 2 replaced the data-file floors of 1 in this table with a reconcile against the golden count, and added the stray-entry and control-character rules (see Round 2, items 1 and 5)._

| What | Change | Guard against scanning nothing |
|---|---|---|
| `scripts/check_recipe_json5.py` (**new**, in `make test`) | Refuses `/* */`, end-of-line `//`, trailing commas (including across a comment line), single quotes, unquoted keys, bare words (`NaN`), strings running past the end of a line, anything else `json.loads` rejects, and a duplicate key at any depth (`object_pairs_hook`, after blanking whole-line comments). The duplicate report names the lines. | ≥1 file; a missing directory fails. `scripts/test_check_recipe_json5.py`: 15 tests; 10 mutations applied to a copy, all red |
| `check_recipe_snapshots.py` | Globs `Resources/Recipes/*.json5`; the paragraph logic is unchanged, since JSON5 comments are whole `//` lines | Total ≥100, and its own data floor ≥1; a missing data directory fails. 3 new tests, each mutation red |
| `check_app_audits.py` | `registry_bundle_ids` reads `"bundleID": "…"`; history pointers are read in `.json5` families with the family rule; count reported | `families ≥ 100` and `.json5 families ≥ 1` |
| `check_engine_notes.py` | Scans `.json5` family comments for pointers. Verified: a bogus `docs/engine-notes/nope.md` pointer in the JSON5 went red naming `…json5:19` | Swift ≥100 and `.json5` ≥1 |
| `check_prose_claims.py` | `rglob` also takes `*.json5`; those files do not count toward the Swift floor | New test; mutation "swift only" red |
| `check_skill_docs.py` | No change needed: counts init parameters and mirrors | n/a |
| channel-discovery skill | The awk never sees `.json5`, so a python one-liner was added for `probes`/`githubRules` `channel` entries. Run: prints nothing on the tree (1Password has no channel); on a scratch file it printed `zz.a channel: beta` and ignored `changelogs` | n/a |
| `CHANNEL_COVERAGE_TODO.md` §2c extraction command | Reads `.json5` too. **Without it the command silently reports `probes 133 ids` instead of `134`** (both measured, before and after) | n/a |
| `AppRecipeIndexTests` | Detailed in the list below | n/a |
| `AppAuditCoverageTests` | No change: it asks the registries, not the files | n/a |
| `RecipeGoldenTests` / `RecipeCodableTests` | Doc lines only ("nothing decodes JSON yet" is no longer true) | n/a |

Changes to `AppRecipeIndexTests`:
- `.swift` files match `swiftFamilies`, and `.json5` files in the **source** directory match what the **bundle** loaded, with a floor.
- No family appears in both forms; `all.count` equals the sum of both halves.
- Both the literal and `all` are sorted.
- `RecipeFamilyFile.CodingKeys` equals `recipeKinds`.
- New:
  - `aDecodeProblemNamesTheFileAndThePath`: unknown key, wrong type, unknown kind, top-level typo.
  - `anAbsentKindIsEmptyAndNullIsRefused`.
  - `aFamilyFileThatDoesNotDecodeTraps`: a swift-testing **exit test** asserting the child process dies and its stderr has `recipe data: Recipes/zz-broken.json5: probes[0].url: required key is missing`, plus the empty-directory floor.

SourceStamp already digested `.json5` (S1). "Compiled into the binary" claims were corrected in:
- `SourceStamp.swift:6` and the complaint text (formerly line 224);
- `duo verify --allow-stale-binary` help (`CLI/Sources/duo/main.swift:201`);
- `SourceStampTests.swift:6`;
- `build-cli.sh`;
- `ChangelogURLPolicy.swift:7,13`, whose "every recipe is compiled into the signed binary" became "ships inside the app's signed bundle".

## Packaging trace

Evidence levels: **R** = read in code, **B** = observed in a real build output from this worktree's `make test`, **P** = reproduced with a scratch probe.

| Executable | Links DuoUpdaterCore / reaches the index? | Bundle at runtime? |
|---|---|---|
| **DuoUpdater.app** | Yes (`App/project.yml` `DuoUpdater.dependencies`). The index is reached through the registries: `Scan/AppScanner.swift:577` `SparkleFeedCatalog.feed` and many others. R | **Yes, in `Contents/Resources`.** B: `check_localizable_keys.py`'s Release build of the DuoUpdater scheme (`/tmp/duo-loc-check-354bf3d8`) has `DuoUpdater.app/Contents/Resources/DuoUpdaterCore_DuoUpdaterCore.bundle/Contents/Resources/Recipes/com-1password-1password.json5`. Found via the accessor's first candidate, `Bundle.main.resourceURL`. That build has signing off, so signed-build behaviour is still open (checklist) |
| Xcode accessor | – | B: Xcode's generated `resource_bundle_accessor.swift` (loc-check DD) is **byte-identical** to SwiftPM's here (same Swift Build engine). Candidates in order: `Bundle.main.resourceURL`, `Bundle(for: BundleFinder.self).resourceURL`, `Bundle.main.bundleURL`, then `fatalError`. No absolute build-path fallback, so a copied binary cannot silently read the build tree |
| **duo-cli** (`make cli`) | Yes via DuoKit. R | Xcode puts `DuoUpdaterCore_DuoUpdaterCore.bundle` in `Build/Products/<config>/` beside tool products. B for Debug/Release products dirs of the app-tests and loc-check builds; the duo-cli scheme itself was not built. `build-cli.sh` copied only the binary, and P confirms a lone copy traps with exit 133. **Changed** (below) |
| **DuoUpdaterHelper** | **No.** `App/project.yml` target `DuoUpdaterHelper` has no `dependencies:`; sources are `Helper/` + `Shared/` only. R | Not needed. The post-build script copies it into `Contents/MacOS`, which is irrelevant here |
| **BinaryDelta** | **No.** Sparkle's prebuilt artifact, located under `SourcePackages/artifacts` by the post-build script. R | Not needed |
| **RowStateGallery** (`make gallery`) | Links Core. No direct registry use in its sources (grep); a transitive reach is not ruled out. R | `row-state-gallery.sh` runs `$DD/Build/Products/Debug/RowStateGallery` in place, beside the products-dir bundle, via `Bundle.main.bundleURL`. R, not run |
| **duo-tcc-spike** | Links Core; uses only `TCCPreflight` (grep of `App/Spike/main.swift`). R: does not reach the index | Would find the products-dir bundle if run in place |
| **DuoUpdaterAppTests** (hostless) | Links Core; reach not proven either way | **Yes.** B: `/tmp/duo-app-tests-354bf3d8/…/DuoUpdaterAppTests.xctest/Contents/Resources/DuoUpdaterCore_DuoUpdaterCore.bundle`, via `Bundle(for: BundleFinder.self)`. `make test`: `✓ App tests — 48 of 48 cases executed` |
| SwiftPM `swift test` / `swift build` | Yes | B: `DuoUpdaterCoreTests.xctest/Contents/Resources/…bundle`; an edited `.json5` is re-copied on incremental build, and a deleted one is removed (mutation 7a saw the empty directory) |
| `application-test` package | Builds Core | `make test` builds it; not run |

Scripts:
- **`scripts/build-cli.sh` (changed):**
  - dies if `$DD/Build/Products/Release/DuoUpdaterCore_DuoUpdaterCore.bundle` is missing;
  - under "Installing", stages the bundle beside `~/.local/libexec` (`mktemp -d` + `ditto`), `rm -rf`s the previous bundle, and renames the new one in, so a deleted family's file cannot linger;
  - **dies unless `diff -r` of the installed `Recipes` against the checkout's `Resources/Recipes` is empty**. The stamp digests the checkout and never reads the bundle, so without this a stale bundle beside a fresh binary would pass `duo verify`'s stale-binary gate;
  - `verify-signature.sh`, `check_swift_backdeploy.py "$PRODUCT"` (binary only; the bundle holds no Mach-O), `codesign --verify --strict "$DEST"` and the `verify --source-digest` stamp are unchanged. `--source-digest` is answered at `main.swift:434` before the index is touched, so the stamp step neither needs nor exercises the bundle.
  - The install fragment was run verbatim, extracted by `sed`, against a scratch libexec and the swift-built bundle (see mutation 9).
- **`scripts/install.sh`**: `ditto "$PRODUCT" "$STAGE"` + `rsync -aE --delete` copies the whole `.app`, bundle included; then `codesign --verify --deep --strict "$DEST"`. R
- **`scripts/notarize.sh` / `release-build.yml`**: the workflow runs `notarize.sh`. Signing is `sign-sparkle-helpers.sh`:
  - `codesign --force … "$APP_PATH"`, not `--deep`;
  - nested items are signed by explicit path (Sparkle XPCs, `mas`, helper); nothing enumerates `Contents/Resources`;
  - the resource bundle holds no code, so it is sealed as resources.

  **Not verified with a signed build: whether `codesign --verify --deep --strict` and notarization accept the nested `.bundle`.** `notarize.sh` has no dry-run mode (grep: none).
- **`scripts/publish-release.sh` / deltas**: generate_appcast/BinaryDelta diff whole app trees; the only content enumeration is `unzip -Z1 … Info.plist`. R: nothing enumerates resources by name.

No blocker for (a) was found.

## Measurements (this machine, Swift 6.4, Xcode 27 beta, SwiftPM debug)

**Incremental work after a one-character comment edit**, two rounds each, reverted in between (`timing.sh`):

| Edit | `swift build --package-path DuoUpdaterCore` | `swift test … --filter RecipeGoldenTests` (includes build) |
|---|---|---|
| JSON5 family | 1.6 s, 1.8 s, 1.6 s, 1.5 s | 2.2 s ×4 |
| Swift family (`com-getdropbox-dropbox.swift`) | 4.4 s, 4.3 s, 4.1 s, 4.3 s | 5.0 s, 4.9 s, 4.8 s, 5.0 s |

**Cold start**, debug build, scratch probe linking Core, three runs:
- `dataFamilies` first access (the `Bundle.module` lookup plus one decode): 7.8, 6.4, 5.3 ms.
- `swiftFamilies`: 2.2, 0.8, 1.3 ms.
- Merge and sort: 0.17 ms.
- Decode alone, for the 4,940 B file: **65 µs** (mean of 1000).
- **Projection, not measured:** decode cost linear in bytes over today's 777,315 B of family files gives roughly 10 ms of debug decode for a full conversion.

**What an author sees** at runtime (probe) and from `check_recipe_json5.py`:

| Typo | Runtime trap message | Checker |
|---|---|---|
| Typo'd key `selectHighst` | `recipe data: Recipes/com-1password-1password.json5: probes[0].selectHighst: unknown key \`selectHighst\`; expected one of: bundleID, channel, … requestHeaders` | passes (not its job) |
| Wrong type `"selectHighest": "true"` | `…json5: probes[0].selectHighest: expected Bool — Expected to decode Bool but found a string instead.` | passes |
| Wrong enum value `"kind": "zipp"` | `…json5: probes[0].install.kind: unknown value \`zipp\`; expected one of: zip, dmg, tarGz, pkg` | passes |
| Missing comma | `…json5: <top level>: The given data was not valid JSON. (Unexpected character '"' in object around line 49, column 7.)` | `(49, 7, 'not-json', "Expecting ',' delimiter")` |
| Unbalanced brace | `… (Unexpected character '{' at beginning of JSON5 unquoted key around line 51, column 19.)` | `(51, 19, 'not-json', 'Expecting property name enclosed in double quotes')` |
| Unterminated string | `… (Unexpected end of file)`, **no line** | `(89, 23, 'open-string', 'string runs past the end of the line — JSON strings are one line')` |

**Conversion size**:
- Swift file: 5,512 B / 92 lines; JSON5: 4,940 B / 92 lines. Diff is +92 −92 (git sees no rename).
- Comments: 60 lines / 3,604 B in both.
- Non-comment content: 1,428 B in Swift, 1,096 B in JSON5.
- Escaping: backslashes outside comments go from 6 to 12.
- Longest non-comment line goes from 102 to 184 characters: the three-part `entryPattern` concatenation becomes one line.

**Worse than Swift, and multiplied by bulk conversion:**
1. **A decode error kills the whole test process** rather than failing one test. Mutation 2: `swiftpm-testing-helper … exited with unexpected signal code 5`. The message is good (file and coding path) but has **no line number**. A Swift typo is a compile error at a line. Each of 225 files is one more place this can happen.
2. **Doubled backslashes** in every regex (S2 design measured 583 of 728 with backslashes; quoted, not re-measured). Multi-part patterns become single long lines, because the D1 dialect has no string concatenation.
3. **For `duo`, recipe data now lives outside the code signature**, in `~/.local/libexec/DuoUpdaterCore_DuoUpdaterCore.bundle`, which is user-writable. **Round 2 closes this for `make cli` builds:** a digest embedded in the signed Info.plist section, checked at load. What remains: every build route that makes a CLI must embed the digest, or its binary skips the check.
4. **Every tool that reads families needs a second reader.** In this PR that was 4 checkers, 1 skill command and 1 ledger command. The ledger command had silently lost a family.
5. **A packaging route that drops the bundle crashes the app at launch.** That is loud, but it happens in production rather than at compile time. Only `build-cli.sh` has an install-time guard.
6. **Comment placement is checked by nothing in `make test`** (mutation 5). Nothing checked it in Swift either, but bulk conversion moves 7k+ comment lines, which is where step 1's defect class came from.

**Better:** incremental build and golden-test time roughly halve (1.6 s vs 4.3 s build, 2.2 s vs 4.9 s golden), and the file is 10% smaller.

## Mutations

Each was applied, run, then reverted with `git checkout -- <file>` (7b: `rm` the renamed file first; 10: `mv` back).

| # | Mutation | Result |
|---|---|---|
| 1 | `\s+` → `\s*` in `versionPattern` | **red.** `family com-1password-1password differs from its golden at line 18 … expected (golden): …\\s+… actual (registry): …\\s*…` |
| 2 | Add `"selectHighst": true` | **trap.** `swift test` exits 1: `swiftpm-testing-helper … exited with unexpected signal code 5`, stderr `Fatal error: recipe data: Recipes/com-1password-1password.json5: probes[0].selectHighst: unknown key \`selectHighst\`; expected one of: …`. The in-process message format is pinned by `aDecodeProblemNamesTheFileAndThePath`, and the trap by the exit test. The dialect checker stays green, correctly |
| 3 | Duplicate `"selectHighest"` (second one `false`) | **Checker red:** `[duplicate-key] key \`selectHighest\` appears twice in one object …; a JSON decoder silently keeps the first — \`"selectHighest":\` is on line(s) 50, 51`. **Golden stays green** (exit 0), which shows the decoder kept `true` |
| 4a | Trailing comma after `"newestLast": true` | **Checker red:** `:89:25: [trailing-comma] trailing comma before \`}\` on line 90`. **`allowsJSON5` alone accepts it:** golden run exit 0, 7 tests passed |
| 4b | `/* the feed escapes its markup */` line | **Checker red:** `:88:7: [block-comment]`. **`allowsJSON5` alone accepts it:** golden exit 0 |
| 5 | Swap the probe's and the changelog's comment blocks | **`comment_sequence.py` red:** `FIRST DIFFERENCE at block 0`. **`make test` does not catch this:** golden + index tests pass (17 tests), and `check_recipe_json5`, `check_recipe_snapshots`, `check_app_audits`, `check_engine_notes` and `check_prose_claims` all exit 0 |
| 6 | Append "Verified against the live feed 2026-09-15: 91 entries, …" to a JSON5 comment | **red.** `…json5:19: dated measurement in a recipe comment [measurement-led]` |
| 7a | Delete the `.json5` | **Trap** before any golden check runs: `Fatal error: recipe data: recipe directory …/DuoUpdaterCoreTests.xctest/Contents/Resources/DuoUpdaterCore_DuoUpdaterCore.bundle/Contents/Resources/Recipes holds no .json5 family files`. It is the only data family, so the empty-directory floor fires. Checkers: `check_recipe_json5` red (`only 0 … files`), `check_recipe_snapshots` red (data floor), `check_app_audits` red (`224 recipe family files (0 .json5)`), `check_engine_notes` red. `check_prose_claims` stays green; it has no data floor, by design |
| 7b | Rename the file to `com-1password-1passwordx.json5`: what deleting one of many families looks like | **Golden red:** `no golden for family com-1password-1passwordx` + `orphan golden com-1password-1password.txt`. `check_app_audits` red |
| 8 | Anchor `#历史与实测` → `#历史` | **red.** `docs/app-audits/com-1password-1password.md: has \`## 历史与实测\` but no Swift or recipe .json5 comment points at it` |
| 9 | `build-cli.sh` without the bundle copy (not runnable; the fragment was run by a harness against a scratch libexec) | Deleting `ditto` with a stale bundle present: `DIE: the installed … has no Recipes directory`, exit 1. Deleting the whole install with a stale bundle that differs from the checkout: `DIE: the recipe files in … differ from …`, exit 1. Same with no bundle: `DIE: … has no Recipes directory`. **Not caught:** a stale bundle whose files equal the checkout's (exit 0), which is not stale by content. A bundle replaced after install is not caught by `duo verify`: the stamp does not read the bundle. As written over a stale bundle holding an extra `zz-deleted.json5`, the extra file is removed |
| 10 | `.json5` → `.json` so the checker finds zero files | **red.** `✗ only 0 recipe .json5 files under … — expected at least 1` |

Checker-only mutations (copies of the scripts):
- `check_recipe_json5.py`: 10 of 10 red. The mutations were: drop `/*` branch, drop `//` branch, drop `object_pairs_hook`, floor `<`→`<=`, comment lines reset the comma, forget escapes, drop single-quote branch, stop blanking comments, `main` returns 0, drop open-string.
- `check_recipe_snapshots.py` data half: 3 of 3 red.
- `check_prose_claims.py` json5 glob: red.

## `make test`

```
$ scripts/run-with-hang-report.sh 1800 make test > make-test-1.log 2>&1; echo exit=$?
exit=0
━ Test run with 3096 tests in 267 suites passed after 20.239 seconds with 1 known issue.
✔ Test run with 352 tests in 44 suites passed after 6.165 seconds.
✓ localizable keys agree — 569 in the catalog, all reachable
✓ app audits consistent — 199 docs, … history pointers resolve (817 Swift files, 225 families, 1 of them .json5)
✓ no dated snapshots in recipe comments — 225 families guarded (1 as .json5)
✓ recipe .json5 files in dialect, no duplicate keys — 1 files
✓ engine-notes pointers resolve — 817 Swift files and 1 recipe .json5 files scanned, 3 doc(s) indexed
✓ skill docs consistent — 3 files mirrored, documented parameter counts (27 / 27) match the initializers
```
- That run was on commit 5 (`70fe942f`).
- Commit 6 changed only the duplicate-key message in `check_recipe_json5.py`. Its 15 unit tests and the checker were rerun green after it.
- The "1 known issue" is an existing `withKnownIssue`.

## Coordinator checklist

I was not allowed to run these. Commands run from the repository root of this branch.

1. **CLI**
   ```sh
   make cli
   ls -d ~/.local/libexec/DuoUpdaterCore_DuoUpdaterCore.bundle
   diff -r DuoUpdaterCore/Sources/DuoUpdaterCore/Resources/Recipes \
       ~/.local/libexec/DuoUpdaterCore_DuoUpdaterCore.bundle/Contents/Resources/Recipes && echo same
   duo verify --only 1password
   codesign -dv ~/.local/libexec/DuoUpdaterCore_DuoUpdaterCore.bundle 2>&1 | head -3   # informational
   ```
   Expected:
   - `make cli` ends at "Installed" with no `✗`. If Xcode puts the bundle elsewhere than `Build/Products/Release/`, it dies at "build produced no resource bundle", which is itself the finding.
   - `ls` prints the path, and `diff` prints `same`.
   - `verify` exits without a crash and reports the 1Password probe and changelog as not failing, with no `unable to find bundle`.
   - `codesign` tells us whether Xcode signs the bundle.
2. **App**
   ```sh
   make install
   ls /Applications/DuoUpdater.app/Contents/Resources/DuoUpdaterCore_DuoUpdaterCore.bundle/Contents/Resources/Recipes/
   codesign --verify --deep --strict --verbose=2 /Applications/DuoUpdater.app
   ```
   Expected:
   - `ls` shows `com-1password-1password.json5`.
   - `codesign` prints `valid on disk` and `satisfies its Designated Requirement`, which `make install` also enforces.
   - The app launches (no new `DuoUpdater-*.ips` in `~/Library/Logs/DiagnosticReports`).
   - The 1Password row (on a machine that has a copy) resolves a remote version as before; `duo check 1password` is the CLI equivalent.
3. **Gallery**: `make gallery`. Expected: it completes, and `git status verify/row-states` shows no changes, since no rendering changed.
4. **Notarization**: `notarize.sh` has **no dry-run mode**. The first signed and notarized build (`make notarize`, or `release-build.yml` via workflow_dispatch) is the check that the nested `.bundle` in `Contents/Resources` is accepted. I suggest one run before any bulk conversion merges.
5. **Helper and BinaryDelta do not touch the registry**
   ```sh
   strings /Applications/DuoUpdater.app/Contents/MacOS/DuoUpdaterHelper | grep -c DuoUpdaterCore_DuoUpdaterCore
   strings /Applications/DuoUpdater.app/Contents/MacOS/BinaryDelta     | grep -c DuoUpdaterCore_DuoUpdaterCore
   strings /Applications/DuoUpdater.app/Contents/MacOS/DuoUpdater       | grep -c DuoUpdaterCore_DuoUpdaterCore
   ```
   Expected: `0`, `0`, and ≥1. The accessor's bundle name is linked only where Core is.

## Verdict: (a) or (b2)

**(a).** No blocker turned up:
- The Swift Build accessor used by both SwiftPM and Xcode is identical and has no build-path fallback.
- Xcode places the bundle in `Contents/Resources` of the app (real Release build) and of the xctest.
- Tool products get it beside them.
- The one install route that copies files by hand now carries it and checks it.
- Incremental edits are cheaper than editing Swift.

Still open, and needing the checklist above:
- signed `codesign --deep --strict` and notarization of the nested bundle (items 2 and 4);
- the Xcode products placement for the duo-cli scheme specifically (item 1).

If item 4 fails, that is the evidence for (b2).

Bulk conversion should also weigh the costs listed above. None of them blocks (a), but they grow with 225 files:
- a decode error kills the test run;
- `duo`'s data sits outside its signature;
- every tool that reads families needs a second parser;
- comment placement is guarded only by a one-off script.


jizhi0v0:
Closing without merging: step 3 stops after S1, and recipes stay Swift. The reasoning and measurements are in `docs/engine-notes/recipe-coding.md` (close-out PR). This branch stays as the record of the loader, the CLI recipe digest (read via `getsectiondata`, not `Bundle.main`), and the JSON5 checker hardening.

PR #677, opened 2026-09-16
Correct stale facts in CLAUDE.md (gallery counts, and where the recipe sweep runs)

jizhi0v0:
Doc-only. Three claims in CLAUDE.md's gallery section were checked against the
tree on 2026-09-16 and were measurably wrong.

| Claim | Tree |
|---|---|
| 40 states × 2 = 80 tiles | 49 × 2 = **98** (`verify/row-states/*`, contiguous 01–49) |
| `mayBeBlank` = `30-up-to-date` + three explanation panels | + **six** (38/39/40/44/48/49) |
| "`mayBeBlank` has no dead-exemption check (#271)" | `deadBlankExemptions` in `App/RowStateGallery/main.swift` — same shape as `mayLookAlike`'s, same build failure |

The third is the one worth fixing: the text told the reader to go back and
remove a stale blank-exemption by hand, because nothing would catch it. That
gap is closed, so the instruction was sending people to do work the build now
does. The same parenthetical had been copied into the prose-claims section, so
both were changed together.

Also loosened "有五道闸" — `main.swift` now sets `failed` in about ten places
(fixture missing, unregistered placeholder, tooltip extractor found nothing,
tooltip scope drifted, …), so the enumerated five are a subset. The text now
says to count `failed = true` rather than trust the numbering. The counts
themselves are marked as drifting, for the same reason.

Verification: `scripts/check_prose_claims.py` passes (822 files). No code changed.


jizhi0v0:
Added `e73d17c3`: the Git section said the recipe sweep runs in this checkout and switches your branch four times a day. Verified over ssh on 2026-09-16 — `com.bobby.duo-recipe-verify` is a LaunchAgent on the mac mini with `WorkingDirectory=/Users/bobby/Developer/duo-updater`, a different path on a different machine, so it cannot touch the dev Mac's working copy. The two failure modes it warned about are also fixed in `duo-recipe-verify.sh` itself (branches from `origin/$BRANCH`, returns to $BRANCH at the end), so the text now points at that comment rather than keeping a second copy.

jizhi0v0:
`/code-review high` on `e73d17c3`: 3 findings, 2 blocking, all of them assertions this branch itself introduced.

- **blocking** `CLAUDE.md:88` — named "fixture missing" as a `failed` gate, then told the reader to enumerate gates by counting `failed = true`; that gate uses `exit(1)` in `render()`, so the grep returns 10 and omits it.
- **blocking** `CLAUDE.md:757` — "每轮收尾 `git reset --hard`" is not every run. Measured on the mini: the checkout+reset pair is reached only when `git diff --staged --quiet` fails *and* `pulled` is set; the `! pulled` path commits locally with `failed=1` and never resets.
- non-blocking `CLAUDE.md:752` — `runs=119` is a counter that advances every 6h, stated as a fixed fact.

All three fixed in `58e0890f`. Targeted re-check: `grep -c 'failed = true'` = 10 (matches the new text), the fixture gate's `exit(1)` confirmed inside `render()`, and the sweep's control flow re-read on the mini to confirm both non-resetting paths. `check_prose_claims.py` passes (822 files). No further AI review round — fixes are text-only with measured evidence and a clear blast radius.

PR #724, opened 2026-09-17
Remove a dead Accessibility helper; fix two stale docs

jizhi0v0:
Three things that stopped matching the code.

- **`AppListModel.guideAccessibilityForIncrementalIfNeeded()` removed.** It has no callers (`git grep` finds only the definition). It served the Settings picker for the incremental App Store route, which Settings no longer offers; that route is now set only with `defaults write`. `presentAccessibilityPermissionFlow()` and `guideAccessibilityOncePerSession()` are still used and stay.
- **README: iOS-on-Mac apps are not skipped.** `MacAppStoreSource` trusts the lookup version for native `mac-software` listings and wrapped iPhone/iPad apps. An iOS-on-Mac app with its own Mac release line is read from the Mac product page (`iosOnMacVersion`), and gets no version when that page has none.
- **application-test/README.md: `records/` is not the evidence.** It is no longer tracked (`.gitignore`), the app-audit skill says evidence goes in the audit doc's 如何复验 section, and `scripts/check_app_audits.py` fails on audits that link to `records/`. Updated the section and dropped the `(see records/)` pointer.

Checked for other copies of both doc statements with `git grep`; none.

Local `make test`: running, result to follow in a comment.


jizhi0v0:
Local `scripts/run-with-hang-report.sh 1800 make test` on this tree (same content as `6ab3e12b`): exit 0.

- DuoUpdaterCore: `Test run with 3336 tests in 288 suites passed after 25.235 seconds with 1 known issue.`
- CLI: `Test run with 350 tests in 44 suites passed after 6.241 seconds.`
- `✓ App tests — 51 of 51 cases executed`
- `✓ localizable keys agree — 602 in the catalog, all reachable` (this step does the Release build of the app, so it compiles the `AppListModel` change)
- `✓ app audits consistent — 205 docs, all indexed, no machine inventory, no pointers into untracked evidence, …`
- The other check scripts (`check_staged_version_use`, `check_recipe_snapshots`, `check_engine_notes`, `check_skill_docs`, `check_prose_claims`, `check_offpool`) all ✓.

PR #726, opened 2026-09-17
Review: 13 verified bugs across the installer, the release log, the recipes, the engine and the CLI

jizhi0v0:
A **global bug sweep** of the codebase, then the fixes for what it found. Review only — no
feature work.

Method: 18 read-only review passes over the 866 tracked Swift files (one per module family:
install, sources/parsers, engine, scan/support, events/traffic, recipes, App model, App views,
CLI, scripts, plus two modules nothing else covered — `application-test/` and the root helper),
each hunting only for defects with a **reachable trigger and a concrete wrong outcome**, and
required to name which link it could not verify. Every finding below I then re-read myself at the
cited lines before touching it; anything I could not stand behind is in "Reported, not fixed".

Each fix is small and independent. **13 fixes in 21 files**, plus one documentation-drift
correction. Four have a reproduction or a mutation-verified test in this PR; the rest state plainly
what was read versus measured.

---

## 1. A feed-named version string deleted an arbitrary directory (data loss)

`DuoUpdaterCore/Sources/DuoUpdaterCore/Install/SparkleInstaller.swift:99-101` and
`VendorInstaller.swift:141-143` built the download scratch directory by interpolating the
**remote's** version string:

```swift
let workDir = FileManager.default.temporaryDirectory
    .appendingPathComponent("DuoUpdater-\(result.app.scratchSlug)-\(remote.displayVersion ?? "new")")
try? FileManager.default.removeItem(at: workDir)          // :101 / :143
try FileManager.default.createDirectory(at: workDir, withIntermediateDirectories: true)
```

`appendingPathComponent` keeps an embedded `/` as a separator and the kernel resolves `..`, so a
feed saying `999.0/../../../../Users/me/Documents` made `workDir` resolve to that directory — and
then the cleanup above, the failure path at `:116`/`:174`, and `InstallCoordinator`'s
`defer { await removeItemOffCooperativePool(at: downloaded.workDir) }`
(`InstallCoordinator.swift:489`) **recursively delete it**, as the user.

The version reaches there unvalidated: `SparkleAppcastSource` takes
`sparkle:shortVersionString` raw (only "non-empty" is checked), `displayVersion` is
`shortVersion ?? version`, and the item can still be offered (`comparableMarketingVersion`
accepts a digit-led, whitespace-free string; the newer/older decision rides the attacker-set
`sparkle:version`). So this runs **before** — and independently of — the EdDSA and Team-ID gates
that a hijacked feed's payload could never pass: it is a data-destruction primitive that
escalates around every signature check the repo has.

**Reproduced** (sacrificial directories, `/tmp` only):

```
workDir.path        = /tmp/zz-trav/scratch/DuoUpdater-slug-999.0/../../VICTIM
workDir standardized= /tmp/zz-trav/VICTIM
VICTIM STILL EXISTS AFTER CLEANUP: false
```

**Fix:** the version goes through `String.filesystemSafeToken` — the helper that already existed
in `Support/FilesystemSafe.swift` for exactly this ("a token can never introduce a path
separator"), and which `scratchSlug` already used. Only this half of the interpolated string was
raw. Both sites.

## 2. Cherry Studio offered a five-releases-stale build as its newest

`Recipes/com-kangfenmao-CherryStudio.swift:16`. The vendor renamed its macOS artifacts at
v2.0.10 — `Cherry-Studio-<ver>-arm64.dmg` became `Cherry-Studio-<ver>-mac-arm64.dmg` — so a
pattern anchored on the old spelling matched nothing from that release on. The walk back past
asset-less releases (`GitHubReleasesSource.swift:1335`, tolerating five) then landed on **2.0.9**:

- every copy on 2.0.10–2.0.13 read "up to date" while 2.0.14 was published;
- anything older was offered a one-click of the **2.0.9** dmg, which the Team-ID gate accepts
  (same vendor);
- `settle` records that walk as a success, so diagnostics called the recipe healthy;
- and because the answer is non-nil, `UpdateChecker` never reached `ElectronManifestSource`,
  which would have answered 2.0.14.

Measured against the live release list (2026-09-17): v2.0.14's macOS assets are
`Cherry-Studio-2.0.14-mac-arm64.dmg` / `.zip`, with the CN edition as
`Cherry-Studio-CN-2.0.14-mac-arm64.dmg`. The repo's own sweep agreed with the bug rather than
catching it: `verify/baseline.json` recorded
`github:CherryHQ/cherry-studio:stable → lastGoodVersion "2.0.9"`, `consecutiveActionable 0`.

**Fix:** the pattern admits both spellings —
`^Cherry-Studio-[0-9.]+-(?:mac-)?arm64\.dmg$` — and still rejects the CN edition, the Intel dmg,
the update zip and the Linux/Windows artifacts. The golden is re-recorded (one line), and new
`CherryStudioGitHubRuleTests` pins the live asset list: **three of its four cases fail** with the
old pattern restored.

The engine half of this is *not* fixed — see "Reported, not fixed".

## 3. A silent privileged helper hung every App Store install forever (the timeout could not fire)

`App/Sources/PrivilegedHelperClient.swift:300-318`. `ensureReachable()` raced a 10-second deadline
against the probe **inside a task group**, then called `group.cancelAll()`. A task group does not
return while any child is still running, and the probe is parked in a
`withCheckedContinuation` that no cancellation can resume — so on the state the code itself
documents as measured ("the connection is accepted and the reply simply never comes", `:233-238`,
observed as `launchd: service inactive` while a batch sat frozen) the deadline resumed
`group.next()` with `false` and the call **hung in the group's implicit drain anyway**. The
`guard answered else { … throw .helperUnresponsive }` at `:311` was unreachable in exactly the
case it was written for.

Consequences when it happens: the row sits at 0 % with no error ever surfaced, the single-slot App
Store gate (`AsyncSemaphore(value: 1)`) is held for the life of the process so every queued App
Store update waits forever, and Settings ▸ Diagnostics' probe leaves its button permanently
spinning and disabled.

**Reproduced in isolation** with the same shape (`swiftc`, exact `withThrowingTaskGroup` structure,
1-second deadline, 12-second watchdog):

```
$ ./tg2 & … kill -9 after 12s
exit=137 (137 = HUNG past its own 1s timeout)
SWIFT TASK CONTINUATION MISUSE: probeVersion() leaked its continuation without resuming it.
```

**Fix:** one continuation, with the deadline and the probe racing through a `ResumeGuard` (which
the file already had, for exactly this) — no dependency on a group's drain. The timeout path also
`invalidate()`s the connection, so the call still parked on the silent peer is torn down instead
of holding its continuation.

## 4. Typing an oversized size into the Requests filter killed the app

`DuoUpdaterCore/Sources/DuoUpdaterCore/Events/RequestQuery.swift:284,287` did
`Int64(magnitude * scale)` / `Int64(magnitude)` on a `Double` parsed from user text. `Int64.init(_:
Double)` **traps** — it does not clamp — on infinite, NaN or out-of-range values, and
`Double(String)` accepts all three shapes: `Double("1e19")` is finite and merely larger than
`Int64.max`, `Double("inf")`/`Double("nan")` parse too. `RequestLogPane` builds its query in
`body`, so this ran on the main actor on every keystroke: `size>1e19` took the whole menu-bar app
down, against the file's own "**Parsing never fails.** … anything unrecognised degrades to free
text" contract. The duration bound (`:466`, `minDuration * 1_000_000`) had the same hole via
`took>1e300`.

**Fix:** a single `int64(_:)` that refuses anything without an `Int64` (bound written as `2^63`,
because `Double(Int64.max)` rounds *up* past it and would still trap), plus `.max` for the duration
bound, which means "matches nothing" — the honest reading of an out-of-range `took>`. New test
`outOfRangeMagnitudesAreRefused`; before the fix it does not fail, it **crashes**.

## 5. The Release Log recorded the same release again on every launch

`Releases/ReleaseTimelineStore.swift:116-118` dedupes on
`$0.version == version && $0.publishedAt == publishedAt`, while `save()`/`load()` round-trip dates
through `.iso8601` (`:316`, `:328`) — which writes **whole seconds**. Three shipped recipes report
sub-second publish times (Claude Desktop's real body carries
`"pub_date":"2026-08-14T22:50:24.042387"`, plus Raycast and Cline), and
`ReleaseDate.parseWithPrecision` keeps the fraction, so after any relaunch the stored copy no
longer compared equal to the date the feed keeps reporting: **+1 duplicate per launch, per
release, forever** — inflating `ReleaseStats`, growing `releases.json` without bound, and
eventually colliding `ReleaseLogView.Row.id` (version+date) inside one `ForEach`.

**Measured:** `.iso8601` writes `1786000224.042387` and reads back `1786000224.0` (unequal); a
whole second round-trips equal.

**Fix:** `record` normalises the two dates it keys on to the precision the store actually carries,
so the key and the persisted value are the same quantity. `flush()` also stops clearing the dirty
bit when `save()` failed, so a batch is retried instead of dropped. New test
`aSubSecondPublishDateIsNotLoggedAgainAfterARelaunch`, **mutation-verified**: without the
normalisation it fails with `!recordedAgain`.

## 6. The changelog pane showed the previous release's notes

`ChangelogService.cacheKeyURL` keyed the in-memory cache on `resolved.absoluteString + "#" +
recipeID` — and for a recipe with a fixed `source` (the common case) `resolved` is the same URL
for every version. So a request for a version discovered inside the 15-minute TTL was an
in-memory **hit** of the previous version's fetch, and the pane drew that release's entries beside
a row offering the new one. It did not repair itself: the prewarm path sets
`fetched = changelog != nil` and marks the key revalidated (`AppListModel.swift:2159,2167`), so the
pane's own path returns early, and the disk write lives inside the fetch closure an in-memory hit
skips — so the new version's notes never reach disk either. The doc comment on `load` already
promised the opposite ("Cache on THIS resolved URL so different versions … never serve each
other's notes").

**Fix:** the target version joins the fragment (`<recipeID>@<version>`), and
`ChangelogCache.invalidate(fragment:)` matches the separator-delimited prefix so "drop this
recipe's notes" still drops every version slot. New test
`twoVersionsOfOneFixedSourceGetDistinctKeys`, **mutation-verified in both halves**: with the
version removed from the key it fails on `keys[0] != keys[1]`; with the prefix match reverted to
equality it fails on `invalidateMemoryCache missed a version slot`.

## 7. A timer or FSEvents rescan read TestFlight's container with the attended gate

`App/Sources/AppListModel.swift:3247` (and the unattended-only `clearSettledExternalUpdates`, now
`:7527+`) used `mayReadTestFlightStore` — which admits Full Disk Access `.unknown`
(`TCCPreflight.admitsOtherAppsData` returns true for it) and is documented as "right for a refresh
someone asked for". Both are reached from a 180-second timer and from FSEvents, where the rule is
the opposite, stated in `TestFlightDetection`'s own warning and pinned by
`TestFlightDetectionUnattendedTests` ("a background timer becomes able to put up a TCC dialog").
The model already had `unattendedMayReadTestFlightStore`, and the other two unattended readers
used it. Constructing `TestFlightInventory` *is* the privileged open, so an unattended trigger
could raise macOS's "access data from other apps" prompt that nobody asked for.

**Fix:** the trigger says which it is — `refreshLocal(unattended:)` / `performLocalRescan(unattended:)`,
with `backgroundLocalRescan` passing `true`; the windows' refreshes and `installAll`'s sweep keep
the attended gate.

## 8. `EventStore`: an unreachable buffer cap, and a reset that could resurrect rows

- `commitBuffer()` returned early when `open()` failed **without** trimming, so
  `maxBufferedEvents` was unreachable in the one state its doc names ("A database that stays
  unwritable must not turn a diagnostic log into a memory leak") — at ~1.5 KB an event, in a
  process that runs for weeks.
- `reset()` cleared `buffer` but never drained `staging`, while `stage()` hands events over from a
  delegate callback and only *then* schedules `absorbStaged` on a `Task`. A reset in that window
  cleared the buffer, missed the staged events, and let the Task drop them back into the emptied
  buffer for the next flush — so the log the user was told is cleared came back holding entries
  from before the reset, totals included.

**Fix:** trim on the un-openable path too; `reset()` drains staging first (which `flush()` already
did, for the mirror-image reason).

## 9. The CLI and the harness

- **`duo triage` trapped (uncatchable).** `Triage.swift:380-383` found `{` and `}` independently
  and then built `text[start...end]`: a reply whose last `}` precedes its first `{` (a stray brace
  in a regex example, then a truncated object) tripped "Range requires lowerBound <= upperBound" —
  a precondition failure, not a throw, so the caller's `catch` could not see it and `duo triage`
  died on a signal. Now refused as the schema error it is.
- **`duo events --limit -5` meant "no limit".** SQLite reads a negative `LIMIT` as unbounded, so
  the flag printed the entire retained history — the opposite of its documented meaning and of why
  the default window exists. `ArgParser` refuses only a non-integer, and `-5` is not a flag. Now
  bad input (exit 2), as `duo requests recent` already treats a non-positive count.
- **`duo install --json` polluted the NDJSON stream.** `Checking N apps…` and the human `describe`
  plan printed to stdout *before* `NDJSON.begin`, so `duo install --json | jq` failed on its first
  line; the rest of `apply` already guards its prose with `if !json`. Both now guarded.
- **`channel-verify --expect` with no value asserted nothing and exited 0.** Every failure path in
  that harness is gated on the expectation being present, so `--expect "$CHANNEL"` with `CHANNEL`
  unset (the flag becomes the last argument) turned a verification run into a pass. **Reproduced:**

  ```
  $ channel-verify /tmp/zz-fix/ZZNoUpdater.app --expect beta   → exit 1, "✗ MISMATCH"
  $ channel-verify /tmp/zz-fix/ZZNoUpdater.app --expect        → exit 0, no message   (before)
  $ channel-verify /tmp/zz-fix/ZZNoUpdater.app --expect        → exit 2, "--expect needs a channel"   (after)
  ```

## 10. The app never noticed a refused elevation written by `duo`

`Preferences.observeExternalWrites()` watched `IgnoredApps` and `SkippedVersions`, but
`duo install` writes a third key into the same suite
(`DuoKit/Settings.recordDeclinedElevation` → `UpdateSettings.declinedElevationKeysKey`), and the
CLI tells the user the menu bar will honour it. The app kept its launch-time copy, so it still
offered the one-click for the refused app and hid the row-menu escape hatch — and because the
set's `didSet` writes the whole set back, the app's next write to that key **deleted the CLI's
entry outright**. Now reloaded with the other two.

---

## Documented drift corrected (not a bug)

`docs/app-audits/README.md` still called four apps detection-only after each had gained one-click:
Comet, AionUi and Msty Studio (2026-08-29) and 豆包输入法 (2026-08-28). Each row is contradicted by
that app's own audit doc (`## 一键安装` → 状态: 已启用 / 已接入), by `CHANGELOG.md:392,396`, and by
`install: VendorInstallSpec` in the recipe. That file is the index `/app-audit` and the next recipe
PR read, so a stale row invites re-filing a closed gap — or "fixing" a working recipe back toward
detection-only. Only the claim changed; each row keeps its own trailing verified-date.

## Verification

```
make test
```

Green on this branch, all gates: Core **3343 tests / 289 suites** (7 more than `main`; the four
new Cherry Studio cases and the three Core ones), CLI 350 tests / 44 suites, `✓ App tests — 51 of 51 cases executed`,
`✓ localizable keys agree — 602 in the catalog, all reachable`, and every lint gate
(`check_staged_version_use`, `check_app_audits`, `check_recipe_snapshots`, `check_engine_notes`,
`check_skill_docs`, `check_prose_claims`, `check_offpool`, plus the ten `test_*.py` suites).

Focused evidence for the three test-backed fixes:

```
swift test --filter "outOfRangeMagnitudesAreRefused|aSubSecondPublishDateIsNotLoggedAgain…\
|twoVersionsOfOneFixedSourceGetDistinctKeys|invalidationDropsEveryVersionResolvedSlot…"
✔ 7 tests in 2 suites passed

# the recipe change, against the live asset names
swift test --filter CherryStudioGitHubRuleTests
✔ 4 tests in 1 suite passed
DUO_RECORD_RECIPE_GOLDENS=1 swift test --filter RecipeGoldenTests   # re-record (fails by design)
swift test --filter RecipeGoldenTests                              # ✔ 7 tests, one golden line changed
```

plus the two mutation runs (each reverted, run, and restored):

```
# revert the invalidate prefix match → ✘ "invalidateMemoryCache missed a version slot"
# revert the version in the fragment → ✘ "two versions of one page must not share a slot"
# revert storedPrecision            → ✘ "the same release must not be re-recorded after a relaunch"
# restore the old asset pattern     → ✘ 3 of 4 CherryStudioGitHubRuleTests cases
```

Not covered by a test, and said so rather than implied: the traversal fix, the helper-timeout fix,
the TestFlight gate, the `EventStore` pair of fixes and the CLI/harness fixes have no unit test in
reach (the affected App files are deliberately outside the `DuoUpdaterAppTests` source list, and
`application-test` has no test target) — their evidence is the reproductions above, the code, and
`make test` staying green.

**Baseline for comparison:** on `origin/main` (`dc0a468e`) the same `make test` is green with
3336 Core tests and 1 known issue — the parked `task_id` schema drift (issue #462), which is a
deliberate non-decision and is untouched here.

---

## Reported, not fixed

Found by the sweep, verified to the level stated, and left alone — either because the correct fix
needs a design decision, or because I could not get the premise past "plausible" without the live
service:

- **Android Studio's stable vendor probe offers a phantom update that can never clear**
  (`Recipes/com-google-android-studio.swift:17`). That recipe captures a **three**-segment version
  from the page path (`install/2025.2.3.<build>/…`) and is the one recipe in the file with
  neither `versionIsBuild` nor `displayVersionPattern` (`probes[0]` in
  `RecipeGoldens/com-google-android-studio.txt:24,26` pins both as false/nil, while `probes[1]` and
  `[2]` — its Canary/Beta siblings — set `versionIsBuild = true`). But the installed
  `CFBundleShortVersionString` is **truncated to two segments**, which this file's own comment
  states ("the installed `CFBundleShortVersionString` is truncated (e.g. "2026.1") … so a
  marketing-version compare is useless — we compare on the `build` field … via `versionIsBuild`")
  and `AppScanner.swift:625-626` repeats ("Android Studio "2025.2" for 2025.2.3"; the fixture in
  `AppScannerFilterTests.swift:334-335` uses the same two-segment form). With `versionIsBuild`
  false the captured string lands in `shortVersion` and `UpdateChecker.evaluate` takes the
  marketing fallback, where `isNewer("2025.2.3", "2025.2")` is **true** — the comparator pads the
  missing component with 0 — and the folded-build rescue needs a dotless digit build, which
  `AI-252.28238.7.2523.14688667` is not. So any copy whose page pattern matches is offered an
  update it already has, forever, and its own one-click dmg reinstalls the same build.
  Left unfixed deliberately: the two candidate fixes have a real trade — truncating the capture to
  `major.minor` clears the phantom but silently stops detecting the patch releases that exist
  inside one minor (2025.2.1 → 2025.2.3), while moving the stable recipe onto the same
  `jb.gg/android-studio-releases-list.json` its siblings read needs the live JSON's stable-entry
  shape. Either way it is a recipe change, and this repo requires those to be reproduced against
  the real response body and to re-record the golden — neither of which this offline sweep could
  do. Worth its own small PR with `duo verify --only google.android.studio`.
- **The asset-less walk-back's tolerance is a boundary, not a guard**
  (`GitHubReleasesSource.swift:951,1335`). `maxReleasesWithoutMacOSAsset = 5` with
  `if skippedForMissingAsset.count > Self.maxReleasesWithoutMacOSAsset { break }` tolerates five
  asset-less releases and then keeps walking, so a run of *exactly* five ends on an older release
  that still matches — offering it as the newest — instead of surfacing as the recipe failure the
  comment two lines above says it must be. Cherry Studio sat exactly on that boundary, which is why
  the recipe fix above was needed rather than the guard firing. Left alone because the right answer
  is a design decision (is five the tolerance? should a match reached after N skips be refused, or
  sanity-checked against the newest release the walk passed?) and not a one-character change to make
  silently. Related, same file: `resolveDiagnostic` reports an asset-rename miss as
  `versionPatternNoMatch`, handing the maintainer the **tag** regex — the conflation the file's own
  comment at `:1412-1416` says must not happen.
- **`duo check --source <unknown>` prints "Everything is up to date." and exits 0** (a typo'd
  source name filters every row out; `Check.swift:221-226,505`). A false green, and every sibling
  enumerated flag already refuses an unknown value. Left because validating the name needs a
  canonical registry of source names — `sourceName`s are per-source statics plus two engine-side
  literals ("Toolbox", "TestFlight"), and a hand-written list is exactly the drift this repo
  removes. Worth its own small PR that puts that list in Core next to `SourceStack`.
- **`BundleFacts.readBundle` reads neither the wrapped interior nor a flat bundle root**
  (`Scan/BundleFacts.swift:276-282`), so for a wrapped iOS app (four are installed on this machine;
  `Amp 2.app` has no `Contents/` at all) `duo diff` reports "0 bundles" and skips the whole TRUST
  SURFACE comparison while still printing a normal-looking report. Verified on disk by the reviewer
  and at both consumer sites; the fix is `BundleLayout.infoPlistURL` plus a root-`Info.plist`
  candidate, which I did not want to fold into a 12-fix PR without its own verification.
- **The App Store AX route's product-page identity check is a substring test**
  (`AppStoreAXInstaller.swift:1861`), so "WeChat" can match a page showing "WeChat Work" — the same
  file already fixed this class for the Updates list (`carriesName`) and not here. Medium-high, but
  falsifying it needs a live App Store, so I did not change it on reasoning alone.
- **The notified-version ledger records before it posts** (`AppListModel.swift:7745`), so with
  notifications denied the user is never told and `wasAnnounced` suppresses it forever. Left: the
  notifier documents delivery as best-effort, so whether this is a bug or the accepted trade is a
  judgement call I did not want to make silently.
- **Smaller, each with its unverified link named in the sweep:** `sparkleStagedBundle` returns the
  first `.app` of an unordered walk; the Squirrel branch reads staged version fields raw while the
  Sparkle branch normalises them; the version-less `.relaunchToApplyStaged(to: nil)` row never arms
  a quit hand-off; `duo verify`'s exit code has no `isInstallTransientReportable` arm while
  `reconcile` files issues for it; `_ = NSWorkspace.shared.open(url)` discards the pkg hand-off's
  only failure signal; a resume that never captured a strong validator can splice two builds;
  `MASInstaller`'s log tailer drops a whole read when a UTF-8 character straddles the boundary;
  `Redactor` cannot see a JSON-quoted secret; `isMajorUpgrade` compares a build against a marketing
  version; `VendorProbeSource.isRicher`'s installability axis can never fire
  (`downloadURL != nil` is true on every vendor remote); the appcast heading regex can still
  swallow `<li>` bullets when a heading closer is mistyped; several structured changelog decoders
  emit markdown while declaring `.plain`.


jizhi0v0:
## Review round 1 — head `42495126`

Each claim was first checked against `main` (does the bug exist), then the fix was read. Probes were standalone `swiftc` programs / sacrificial temp dirs; no system UI, no real helper. CI `test` green; `git merge-tree` onto current main (after #724) is conflict-free.

### Does each bug exist, and is the fix right?

| # | Claim | Exists on main | Fix |
|---|---|---|---|
| 1 | feed version → path traversal → recursive delete | yes — reproduced (victim dir deleted; kept with the fix) | correct; no regression test |
| 2 | Cherry Studio asset rename → stale 2.0.9 | yes — live release list: rename at v2.0.10, walk-back skips exactly 5 and lands on 2.0.9; new regex hits exactly the 51 arm64 dmgs of 1,089 asset names | correct; 3/4 new tests fail with the old pattern |
| 3 | helper timeout unreachable inside task group | yes — reproduced (1s timeout, killed by 8s watchdog; fixed shape returns in 1s, no double resume, next call reconnects). "Forever" not proven: 7636a687 suggests XPC eventually errors in the stale-helper state | correct |
| 4 | `Int64(Double)` trap in RequestQuery | yes — `size>1e19`/`inf`/`nan`, `took>1e300` all SIGTRAP | correct; test never calls `sqlPredicate()`, so `took>` is uncovered. **Missed sibling:** `duo events --since 1e14d` / `infd` traps in `EventStore.micros` |
| 5 | Release Log re-records sub-second dates per launch | yes — 4 simulated launches: 1→4 events on main, 1 with fix | correct; existing on-disk duplicates are not cleaned |
| 6 | changelog pane shows previous version's notes | partially — the cache holds the whole page and nothing filters by version, so it only bites when the vendor updated the page inside the 15-min TTL; within-session non-repair confirmed | correct; `ChangelogCache.swift:11-13` and `ChangelogService.swift:29` comments now stale |
| 7 | timer/FSEvents rescans take the attended TF gate | yes, but only differs when FDA status is `.unknown` (rare) | **incomplete:** `WorkbenchWindowView.swift:321-323` 180 s `refreshTimer` still calls `refreshLocal()` with default `unattended: false` |
| 8 | EventStore cap unreachable / reset resurrects staged rows | yes, both, low severity (cap: only when the file cannot be opened at all; resurrect: race probe 9 rows on main, 0 with fix) | correct |
| 9a | `duo triage` brace-range trap | yes — reproduced | correct |
| 9b | `--limit -5` unbounded | yes — SQLite doc: negative LIMIT = no upper bound | correct |
| 9c | `install --json` prose before schema line | yes | **incomplete + regression:** `Nothing to install.` (`:125`), the confirm prompt, `Cancelled.`, `Skipping …` still hit stdout before any schema line; `--json --dry-run` and refusals-only now print nothing at all |
| 9d | `channel-verify --expect` without value exits 0 | yes (code trace) | correct |
| 10 | app ignores CLI-written declined-elevation key | yes, both halves (probe with the real `DefaultsKeyObserver`) | correct, no reload ping-pong |
| docs | four stale detection-only rows | yes | correct (豆包's CHANGELOG ref is `:437`, not 392/396) |

### Verdict
Blocking, being fixed on this branch:
- 9c: `--json` stdout must be valid NDJSON on every exit path of `run`, with nothing the human mode shows silently dropped.
- 7: workbench `refreshTimer` → `unattended: true`; drop the parameter default so a future miss is a compile error.

Also being fixed (same bug class / now-false statements):
- 4: `--since` trap sibling; `took>` test coverage via `sqlPredicate()`.
- 6: the two stale comments.

Not blocking, out of scope: `EventStore` ignores `insert`'s result, so a read-only store silently drops a committed batch (pre-existing, separate issue).

jizhi0v0:
## Review round 2 — fixes pushed, head `0c977e8c`

Follow-up commits for the round-1 findings, each reviewed before landing:

| Commit | Finding | What changed |
|---|---|---|
| `d6ebd8e3` | 7: workbench timer took the attended TF gate | `WorkbenchWindowView` `refreshTimer` passes `unattended: true`; the `= false` default is gone from `refreshLocal`/`performLocalRescan`, so every caller states it (all 8 call sites re-classified; `restart(byID:)`'s only caller is the notification action tap) |
| `cc5fc0ef`, `0c977e8c` | 9c: `install --json` stdout not NDJSON on every path | `NDJSON.begin` moves to the top of `run`; empty plan → schema line only; refusals → `skippedPayload` rows carrying `Decision.refuse`'s route (review caught it being dropped in the first pass); confirm prompt / cancel → stderr in json mode; the defer-recheck skip → a skipped row. Same bug fixed in `duo skip --json` (`Checking N apps…`) and `duo backups restore --json` (prompt, `Cancelled.`). Human output unchanged. `--json --dry-run` emits the schema line and refusal rows but not "would install" rows — no existing row shape fits, and adding one is a product decision |
| `d3123481` | 4 sibling: `duo events --since 1e14d` / `infd` trapped | `EventStore.micros` saturates via `RequestQuery.int64` instead of trapping (both `rawRows` and `--filter` paths go through it). New `outOfRangeTimeBoundsClamp`: SIGTRAP on the unfixed code, passes after. Sweep of other `Int64(Double)` sites found none reachable from input |
| `4e111f90` | 4: `took>` fix untested | `outOfRangeMagnitudesAreRefused` now calls `sqlPredicate()` for `took>1e300/inf/nan/1e300ms/9.3e12`; reverting only the `took>` line → SIGTRAP |
| `8a652941` | 6: stale cache-key comments | `ChangelogCache` header, `ChangelogService.load`/`cacheKeyURL` docs and two DocC links describe the per-version key |

### Verification on `0c977e8c`
`make test`, every step:
- Python suites: all `OK`.
- Core: `3344 tests in 289 suites` — **1 failure, unrelated**: `vendorResolvesInstallPlans()` → `com.termius-beta.mac [beta] resolved no installer — httpStatus503`. Live network test; `curl https://autoupdate.termius.com/mac-beta-universal/latest-mac.yml` → `503` from this machine (behind a proxy; not confirmed from another network). None of these commits touch that code.
- CLI `350 tests in 44 suites passed`; `application-test` builds; `✓ App tests — 51 of 51 cases executed`.
- `check_localizable_keys`, `check_staged_version_use` (+ its test), `check_app_audits`, `check_recipe_snapshots`, `check_engine_notes`, `check_skill_docs`, `check_prose_claims`, `check_offpool`: all ✓.
- `git merge-tree` onto current `main`: clean.

### Status
All round-1 blocking items closed. Remaining, out of scope: `EventStore` ignores `insert`'s result on a read-only store (pre-existing; filed as a separate task).

PR #727, opened 2026-09-17
GitHub: report a renamed macOS asset as an asset miss, not a version miss

jizhi0v0:
A `duo verify` sweep that hits a renamed macOS artifact reports the wrong failure, and hands the
reader the wrong regex. This fixes the report; it deliberately does **not** move the walk-back
boundary that decides whether an older release is offered (see the last section).

Found while reviewing the recipes, chasing a live one: Cherry Studio renamed its macOS artifacts at
v2.0.10 (`Cherry-Studio-<ver>-arm64.dmg` → `Cherry-Studio-<ver>-mac-arm64.dmg`), the install pattern
kept matching only the old spelling, and the source answered **2.0.9** for a repo whose newest
release was 2.0.14 — copies on 2.0.10–2.0.13 read "up to date", and `verify/baseline.json` recorded
2.0.9 as a *good* version. (The pattern itself is fixed separately in #726.)

## The bug

`GitHubReleasesSource.settle` walks back past releases whose assets do not match
`installAssetPattern`, accumulating them in `skippedForMissingAsset`. When the walk ends with no
answer at all it *does* record the distinction correctly — and says why it matters, two lines above
the code:

```swift
// Two different breakages end up here and they need different words: a
// tag-format change (nothing matched the version pattern) versus an
// asset rename (tags matched fine, none carried the macOS file). Reported
// as the same thing, the second reads like the first and gets fixed in the
// wrong place.
```

That distinction was never carried out of `settle`. `Resolution` holds only
`remote`/`tags`/`archIncompatible`, so `resolveDiagnostic` returned
`.versionPatternNoMatch(sampleBytes: resolved.tags.count)` for both cases — literally the byte count
of the *tag list* — and the CLI then passed `pattern: rule.versionPattern` into the finding. So a
maintainer was told "no match in 53-byte body" and pointed at the **tag** regex, which was matching
every one of those releases perfectly. The asset pattern — the thing that actually broke — appeared
nowhere in the report.

This is the reporting half of the Cherry Studio class: had that walk failed instead of landing on an
older match, the sweep would have blamed the tag pattern, and the fix would have gone in the wrong
place. It is the same failure `buildLineagePatternNoMatch` was split out for (`VendorProbeDiagnostics`).

## The fix

- `ProbeFailure.assetPatternNoMatch(walked: Int)` — a recipe problem like `versionPatternNoMatch`,
  with its own `kind` (`assetPatternNoMatch`) and a `detail` that says what happened: *"N release(s)
  matched the version pattern but none carried an asset matching the install pattern — the vendor
  may have renamed the macOS artifact"*. `walked` is the size of the evidence.
- `Resolution` carries the count (0 everywhere except that exit), so the caller can tell the two
  pattern failures apart.
- `resolveDiagnostic` maps it to the new kind instead of the version one.
- `CLI/Sources/DuoKit/Verify.swift` quotes the pattern the failure is *about*: for
  `assetPatternNoMatch` the finding now carries `installAssetPattern`, not `versionPattern`.

## Tests

`GitHubAssetWalkBackTests` drives the source through a stub session (`GitHubReleasesSource(session:)`,
the idiom `CotEditorChannelTests` already uses) — five cases:

1. **A renamed artifact is reported as an asset miss** — six asset-less releases, no answer, and the
   failure is `.assetPatternNoMatch(walked: 6)`. Mutation-verified: with the plumbing removed it
   fails with the old `versionPatternNoMatch(sampleBytes: 53)`.
2. **A tag that matches no version is still a version miss** — the distinction has to cut both ways,
   or a tag-format change would be blamed on the asset pattern instead.
3. **Five asset-less releases then a match → the older release is offered.**
4. **Four asset-less releases then a match → same.**
5. **A matching asset on the newest release → answered without walking.**

Cases 3–5 exist because this area has two contracts that pull against each other and neither was
tested:

- `Recipes/com-federicoterzi-espanso.swift:12-15` relies on the tolerated side — its asset pattern is
  a literal (`Espanso-Mac-Universal.dmg`) and older releases shipped the same name as a `.zip`, so
  "walk back to the newest release that still carries it" is the behaviour it was written for.
- The walk's own comment says the opposite about a *run*: "a run of them means the pattern stopped
  matching, which is a recipe failure and has to surface as one."

Cherry Studio sat exactly on that seam: five asset-less releases, then a match on the sixth.
`GitHubListPageSizeTests` asserts the walk and page exhaustion "reach the same `recordMiss` exit",
which is true only when no match is found. **This PR does not change the boundary** — moving it
would break one of the two documented contracts above, and that is the maintainer's call, not a
review's. What it does is make the boundary visible and enforced: three cases now pin where it is,
and the six-release case pins what the walk says when it runs past it.

## Verification

```
swift test --filter GitHubAssetWalkBackTests          ✔ 5 tests (3 consecutive runs, no flakes)
# mutation: drop `assetMiss` from Resolution          ✘ "expected an asset-pattern miss,
#                                                         got versionPatternNoMatch(sampleBytes: 53)"
make test        ✔ all gates green — Core 3341 tests / 289 suites (the 5 new ones are this
                 suite), CLI 350 / 44, ✓ App tests 51 of 51, every lint gate green
```

The stub is keyed by owner/repo slug and immutable, because the first version used one shared
`static var` and swift-testing's parallelism made the cases read each other's fixtures — the suite
flipped between green and red run to run until that was fixed.

## Not in this PR

- **The walk-back boundary itself** (`maxReleasesWithoutMacOSAsset = 5`, and the answer standing when
  a match is found after a run). Left alone for the reason above; now pinned by tests.
- **Android Studio's stable probe**, the other live instance of this class: its `versionPattern`
  captures a three-segment version from `install/<year>.<minor>.<patch>.<build>/` while the installed
  `CFBundleShortVersionString` is truncated to two (`AppScanner.swift:625-626` documents exactly this,
  and its Canary/Beta siblings use `versionIsBuild` to avoid it), so `isNewer("2025.2.3", "2025.2")`
  is true forever — a phantom update that can never clear. Both candidate fixes need live data:
  truncating the capture to `year.minor` silently stops seeing the patch releases inside one minor,
  and moving the recipe onto the siblings' `jb.gg/android-studio-releases-list.json` needs that
  JSON's stable-entry shape. Neither `developer.android.com` nor `jb.gg` resolved from the sandbox
  this was written in, so it is not fixed here rather than guessed at.


jizhi0v0:
## 复审第 1 轮

- reviewer：Claude Code 会话里的 `/code-review 727`（medium）。**不是独立会话**：第 2 轮的修复也是同一个会话写的，见第 2 轮的说明
- 审的 head：`0ad8695192b6fc98b72c57234308f29acedf9bc0`（范围 `main..0ad86951`，1 个提交，4 个文件）

### reviewer 最终回复原文

````
I reviewed #727 and found 2 issues. The main change is sound: `settle` now carries the asset-miss count out, `resolveDiagnostic` checks it before falling back to a version miss, and the report and CLI quote the install pattern. Nothing else in `CLI/`, the core sources, `scripts/` or `.github/` treats `versionPatternNoMatch` specially, and `verify/baseline.json` has no GitHub rows under that name whose label would change.

1. [Verify.swift:872](CLI/Sources/DuoKit/Verify.swift:872) — **Triage can post a "✅ Verified" fix for the wrong regex.** Findings with the new `assetPatternNoMatch` failure still go to `duo triage`. Triage asks the model for a *version* regex, quotes the install pattern as "current pattern", and checks the answer against the captured sample, which for a GitHub rule is only the tag list. Any tag regex the model proposes matches that list, so the issue would say "✅ Verified: extracts `2.0.14`" right under a line saying the version pattern is fine. That is the misdirection this PR is meant to remove. Two ways out: skip triage for this failure type, or capture the asset names so a fix can be checked against them.
2. [GitHubAssetWalkBackTests.swift:136](DuoUpdaterCore/Tests/DuoUpdaterCoreTests/GitHubAssetWalkBackTests.swift:136) — **The tests depend on the machine's CPU.** They call the public `resolveDiagnostic(rule)`, which reads the host architecture, and every fixture asset is arm64-only. On an Intel Mac, the three tests that expect an answer get "not applicable" and fail. Using the internal overload with `preferring: .arm64, allowingIntelTranslation: false` fixes this.

**CI:** the PR's `test` check is red. The new `GitHubAssetWalkBackTests` suite passed there. The failure is `VendorInstallTests.vendorResolvesInstallPlans`, which makes live network calls: `com.termius-beta.mac [beta]` got an HTTP 503 from Termius. That looks unrelated to this PR, but I haven't re-run it, so it needs a re-run before merge.
````

### 作者裁定

- **Finding 1：确认，阻塞。** 不修的话，合并后 `assetPatternNoMatch` 的 issue 会带一条「✅ Verified」的 tag 正则建议，正好把人引向本 PR 要避免的那个错误正则。修法选「对这种失败跳过 triage」：tag 列表里没有 asset 名，建议没法校验。
- **Finding 2：确认，非阻塞。** 只在 Intel Mac 上跑测试时才会红，CI runner 不受影响；改动只有一行调用，按用户要求一并修了。
- **CI 红：不是本 PR 引起的。** 失败的是 `vendorResolvesInstallPlans` 里 Termius beta 的实时请求，`autoupdate.termius.com` 的 stable 和 beta feed 在本机 curl 也都是 503；本 PR 没碰任何 recipe。

修复提交：`23e12acd`。

jizhi0v0:
## 复审第 2 轮（针对性复查）

- reviewer：与第 1 轮同一个会话。⚠️ **这个会话也写了修复提交 `23e12acd`，所以这一轮不独立**；合并前仍需另开 reviewer 会话，按固定提示词复查 `0ad86951..23e12acd`
- 审的 head：`23e12acde3e08b03c863f96dc41b8bf6ed36c2f6`；增量 `0ad86951..23e12acd`（线性，1 个提交，3 个文件）

### reviewer 最终回复原文

````
I rechecked only what the fix commit `23e12acd` changed, not the whole PR again. Both findings are fixed and I found nothing new.

**Triage fix** (skipping asset-pattern misses):
- **It matches the real findings.** A GitHub finding's failure name is the one `resolveDiagnostic` sets, and the check compares against that same name from the code rather than a typed-out copy.
- **Nothing downstream needed a triage result.** Suggestions are optional when the issue is filed, so a skipped finding just gets no suggestion block. It never records a "triaged" marker either, so each sweep prints the skip reason again, which costs no model call.
- **The other skip checks are untouched.** The new check sits after the streak check and before the empty-body check, and existing tests still use the default failure name. 25 of 25 pass.
- **Mutation proof:** with the check disabled, the new test fails.

**Test fix** (pinning arm64):
- The tests now call the internal version of `resolveDiagnostic`, which is what the public one calls with the host's values. On an Apple Silicon Mac nothing changes (5 of 5 pass). Setting the host to Intel fails exactly the three answering cases, so the pin is what decides the result.

**Tests:** I ran all of `make test` on this exact code before committing. Everything passed except `vendorResolvesInstallPlans`.

**CI:** the `test` check is still running. It will probably fail on that same test: Termius's beta update feed still returns 503 as of just now. This PR doesn't touch the Termius recipe, so re-run the check once their server is back.

Two caveats:
- **This isn't an independent review.** This same session wrote the fix, and your process has a separate reviewer session do the recheck. Before merging, give that session the SHA `23e12acd` and your standard recheck prompt.
- **Neither round is on the PR yet.** Your process leaves a PR comment per round (head SHA, reviewer reply, verdict). Want me to post one covering both rounds?
````

### 作者裁定

- **两条已确认的问题都已闭环。** 本轮没有新 finding。
- **证据**（本机实跑）：
  - Triage 检查关掉时，新测试 `anAssetPatternMissIsNotAsked` 失败（`Expectation failed: why?.contains("asset-pattern miss") == true`）；恢复后 `TriageTests` 25/25 通过。
  - host 设成 `.x86_64` 时，`GitHubAssetWalkBackTests` 里正好三个会给出答案的用例失败（7 个 issue，都是 `remote == nil`）；钉回 `.arm64` 后 5/5 通过。
  - `make test` 在这份代码上跑到 core 套件停下，唯一失败是 Termius 那条（3341 个测试，另有 1 个 known issue）。后面的步骤手动跑完：CLI 351/351，App tests 51/51，各检查脚本全过。
- **还没闭环的：**
  - 独立 reviewer 会话复查 `23e12acd`（见上方 ⚠️）；
  - CI `test` 通过（等 Termius 恢复后 re-run）。

PR #733, opened 2026-09-17
Claude review: take the reviewer's model from a repository variable

jizhi0v0:
## What

The Claude review workflow now reads the reviewer's model from the repository variable `CLAUDE_REVIEW_MODEL`. When the variable is set, `claude_args` gets `--model <value>`. When it is unset, the flag is omitted and the action's default model applies, as before.

## Why

Changing the model no longer needs a pull request. The variable is set in Settings → Secrets and variables → Actions → Variables.

## Testing

- `actionlint` is clean.
- This PR's own review run is the live check. The run log's `"model"` line should show the value of the variable.


jizhi0v0:
Author verdict on round 1 (`dc0f687`). No blocking findings. The three unverified external claims were checked locally against run logs and the pinned action:

| # | Claim | Verdict | Evidence |
|---|---|---|---|
| 1 | `claude-code-action@v1.0.227` accepts `--model` through `claude_args`, with no conflicting `model` input | **confirmed** | The pinned `action.yml` (`3b8197d3`) has no top-level `model` input. This PR's run 35250311329 logs `claude_args: --model <variable value> --allowedTools …` and `"model"` equal to that value. |
| 2 | With the variable unset, the action's default model applies | **confirmed** | Run 35240997493 (#731, no `--model` in `claude_args`) logs `"model"` as the action's default, which differs from the variable's value. |
| 3 | The run log prints a `"model"` line | **confirmed** | Both runs above print it in the Review step's init output. |

---

PR #740, opened 2026-09-18
verify: fix the six open recipe-broken issues, and one that never reported

### LINKED ISSUE #467: Recipe degraded: com.qoder.ide (changelog) — newest changelog entry (1.28.0) trails the detected version (1.29.0)…
Found by the scheduled `duo verify` sweep. The version check and the app both keep working in the cases marked *degraded* — that is what makes these invisible without this job.

| | |
|---|---|
| recipe | `changelog:com.qoder.ide:-` |
| registry | changelog |
| endpoint host | `docs.qoder.com` |
| pattern | `data-component-part="update-label">(?<date>[^<]{3,40})</div>(?:(?!data-component-part="update-label").)*?data-component-part="update-description">(?:Qoder\s+)?(?<version>[0-9]+(?:\.[0-9]+)+)</div>(?:(?!data-component-part="update-label").)*?data-component-part="update-content">(?<body>(?:(?!data-component-part="update-label").)*?)</div></div></div>` |
| resolved | `1.28.0` |
| entries parsed | 40 |
| last good | `1.28.0` |
| consecutive sweeps | 2 |

**Warnings**

- newest changelog entry (1.28.0) trails the detected version (1.29.0) by a whole release — the entry pattern may be reading a stale section

<details><summary>Captured response (redacted, truncated)</summary>

```
…[sample centred on the pattern's anchor 'data-component-part="update-description">']…
flex size-6 items-center justify-center rounded-md bg-white text-stone-400 shadow-sm ring-1 ring-stone-400/30 hover:ring-stone-400/60 group-focus/link:border-2 group-focus/link:border-adoc-primary dark:bg-adoc-bg dark:text-white/50 dark:ring-1 dark:ring-stone-700/25 dark:brightness-[1.35] dark:group-focus/link:border-adoc-primary-light dark:hover:ring-white/20 dark:hover:brightness-150"><svg aria-hidden="true" fill="currentColor" height="12px" viewBox="0 0 576 512" xmlns="http://www.w3.org/2000/svg"><path d="M0 256C0 167.6 71.6 96 160 96h72c13.3 0 24 10.7 24 24s-10.7 24-24 24H160C98.1 144 48 194.1 48 256s50.1 112 112 112h72c13.3 0 24 10.7 24 24s-10.7 24-24 24H160C71.6 416 0 344.4 0 256zm576 0c0 88.4-71.6 160-160 160H344c-13.3 0-24-10.7-24-24s10.7-24 24-24h72c61.9 0 112-50.1 112-112s-50.1-112-112-112H344c-13.3 0-24-10.7-24-24s10.7-24 24-24h72c88.4 0 160 71.6 160 160zM184 232H392c13.3 0 24 10.7 24 24s-10.7 24-24 24H184c-13.3 0-24-10.7-24-24s10.7-24 24-24z"></path></svg></div></a></div><div class="flex grow-0 cursor-pointer items-center justify-center rounded-lg bg-adoc-primary/10 px-2 py-1 font-medium text-adoc-primary text-sm" contentEditable="false" data-component-part="update-label">September 2, 2026</div><div class="wrap-break-word mt-3 max-w-[160px] px-1 text-adoc-text-secondary text-sm dark:text-adoc-text-tertiary" contentEditable="false" data-component-part="update-description">1.28.0</div></div><div class="max-w-full flex-1 overflow-hidden px-0.5"><div class="prose-sm" data-component-part="update-content"><h3>Improvements</h3><ul>
<li>Increased the tool execution limit for single tasks in Qoder IDE to 500 rounds, making complex long-chain tasks less likely to be interrupted prematurely.</li>
<li>Tasks interrupted due to depleted credits can now be manually resumed by clicking continue after credits are restored.</li>
</ul></div></div></div>
<div id="1270-2026-08-29"></div>
<div class="update update-container relative flex w-full flex-col items-start gap-2 py-8 lg:flex-row lg:gap-6" id="august-29-2026"><div class="group top-(--scroll-mt) flex w-full shrink-0 flex-col items-start justify-start lg:sticky lg:w-[160px]"><div class="absolute"><a aria-label="Navigate to changelog" class="group/link -ml-10 flex items-center border-0 opacity-0 focus:opacity-100 focus:outline-0 group-hover:opacity-100" href="#august-29-2026">​<div class="flex size-6 items-center justify-center rounded-md bg-white text-stone-400 shadow-sm ring-1 ring-stone-400/30 hover:ring-stone-400/60 group-focus/link:border-2 group-focus/link:border-adoc-primary dark:bg-adoc-bg dark:text-white/50 dark:ring-1 dark:ring-stone-700/25 dark:brightness-[1.35] dark:group-focus/link:border-adoc-primary-light dark:hover:ring-white/20 dark:hover:brightness-150"><svg aria-hidden="true" fill="currentColor" height="12px" viewBox="0 0 576 512" xmlns="http://www.w3.org/2000/svg"><path d="M0 256C0 167.6 71.6 96 160 96h72c13.3 0 24 10.7 24 24s-10.7 24-24 24H160C98.1 144 48 194.1 48 256s50.1 112 112 112h72c13.3 0 24 10.7 24 24s-10.7 24-24 24H160C71.6 416 0 344.4 0 256zm576 0c0 88.4-71.6 160-160 160H344c-13.3 0-24-10.7-24-24s10.7-24 24-24h72c61.9 0 112-50.1 112-112s-50.1-112-112-112H344c-13.3 0-24-10.7-24-24s10.7-24 24-24h72c88.4 0 160 71.6 160 160zM184 232H392c13.3 0 24 10.7 24 24s-10.7 24-24 24H184c-13.3 0-24-10.7-24-24s10.7-24 24-24z"></path></svg></div></a></div><div class="flex grow-0 cursor-pointer items-center justify-center rounded-lg bg-adoc-primary/10 px-2 py-1 font-medium text-adoc-primary text-sm" contentEditable="false" data-component-part="update-label">August 29, 2026</div><div class="wrap-break-word mt-3 max-w-[160px] px-1 text-adoc-text-secondary text-sm dark:text-adoc-text-tertiary" contentEditable="false" data-component-part="update-description">1.27.0</div></div><div class="max-w-full flex-1 overflow-hidden px-0.5"><div class="prose-sm" data-component-part="update-content"><h3>Improvements</h3><ul>
<li>Added an enterprise control to disable external network access from the built-in browser, helping organizations meet security and operational management requirements.</li>
<li>Personal edition BYO
…[truncated]…
```

</details>

### Automated analysis

✅ Verified: the proposed pattern extracts `1.28.0` from the captured body, using the same extractor the app uses.

<details><summary>Unverified suggestion — do not apply without testing (model: `opencode-go/deepseek-v4-pro`, self-reported confidence 0.60)</summary>

The visible changelog's newest entry (September 2, 2026) is 1.28.0, but the old pattern's date group accepts any text, so its first match now lands on another section of the page (hidden in the truncated window) that reuses the same update-label/update-description markers and reports 1.29.0. Requiring the label to be a real 'Month D, YYYY' date keeps the match inside the changelog entries, which are newest-first, so the first hit is the newest release.

Proposed pattern:

```
data-component-part="update-label">(?:January|February|March|April|May|June|July|August|September|October|November|December)\s+\d{1,2},\s+\d{4}</div>(?:(?!data-component-part="update-label").)*?data-component-part="update-description">(?:Qoder\s+)?([0-9]+(?:\.[0-9]+)+)</div>
```

Produced by a model with no tools, in an empty directory, from the captured response above. That response is third-party content and may be hostile; nothing here has been applied to any file.

</details>

Reproduce locally:

```bash
swift run --package-path CLI duo verify --changelog --only com.qoder.ide --samples
```

### LINKED ISSUE #698: Recipe degraded: com.typewhisper.mac (changelog) — version went BACKWARDS since the last sweep (1.7.0-daily.20260912 →…
Found by the scheduled `duo verify` sweep. The version check and the app both keep working in the cases marked *degraded* — that is what makes these invisible without this job.

| | |
|---|---|
| recipe | `changelog:com.typewhisper.mac:-` |
| registry | changelog |
| endpoint host | `www.typewhisper.com` |
| pattern | `>macOS</span><h3[^>]*>v?(?<version>[^<]+)</h3>(?:(?!>macOS</span><h3|>Windows</span><h3).)*?<p class="mt-1 text-xs text-muted-foreground">(?<date>[^<]*)</p><div class="prose[^"]*"[^>]*>(?<body>.*?)</div>` |
| resolved | `1.6.1` |
| entries parsed | 20 |
| last good | `1.7.0-daily.20260912` |
| consecutive sweeps | 2 |

**Warnings**

- version went BACKWARDS since the last sweep (1.7.0-daily.20260912 → 1.6.1) — the pattern may have started matching a different element

Reproduce locally:

```bash
swift run --package-path CLI duo verify --changelog --only com.typewhisper.mac --samples
```

### LINKED ISSUE #713: Recipe degraded: com.bjango.istatmenus (vendor probe) — version went BACKWARDS since the last sweep (7.50.1 → 7.50)
Found by the scheduled `duo verify` sweep. The version check and the app both keep working in the cases marked *degraded* — that is what makes these invisible without this job.

| | |
|---|---|
| recipe | `vendor:com.bjango.istatmenus:stable` |
| registry | vendor probe |
| endpoint host | `download.istatmenus.app` |
| pattern | `iStatMenus([0-9]+\.[0-9]+(?:\.[0-9]+)?)\.zip` |
| resolved | `7.50` |
| last good | `7.50.1` |
| consecutive sweeps | 2 |

**Warnings**

- version went BACKWARDS since the last sweep (7.50.1 → 7.50) — the pattern may have started matching a different element

Reproduce locally:

```bash
swift run --package-path CLI duo verify --vendor --only com.bjango.istatmenus --samples
```

### LINKED ISSUE #736: Recipe degraded: com.figma.Desktop (changelog) — version went BACKWARDS since the last sweep (Publish Weave tools to…
Found by the scheduled `duo verify` sweep. The version check and the app both keep working in the cases marked *degraded* — that is what makes these invisible without this job.

| | |
|---|---|
| recipe | `changelog:com.figma.Desktop:-` |
| registry | changelog |
| endpoint host | `www.figma.com` |
| pattern | `<entry>\s*<title type="html"><!\[CDATA\[(?<version>.*?)\]\]></title>\s*<id>[^<]*</id>\s*<link[^>]*/>\s*<updated>(?<date>[^T]+)T[^<]*</updated>\s*(?<body><content type="html"><!\[CDATA\[.*?\]\]></content>)` |
| resolved | `Create on-brand content with your Figma designs in Weave workflows` |
| entries parsed | 20 |
| last good | `Publish Weave tools to the Figma Community` |
| consecutive sweeps | 2 |

**Warnings**

- version went BACKWARDS since the last sweep (Publish Weave tools to the Figma Community → Create on-brand content with your Figma designs in Weave workflows) — the pattern may have started matching a different element

Reproduce locally:

```bash
swift run --package-path CLI duo verify --changelog --only com.figma.Desktop --samples
```

### LINKED ISSUE #737: Recipe degraded: com.workbuddy.workbuddy-ai (vendor probe) — installURLNotFound: HTTP 404 from…
Found by the scheduled `duo verify` sweep. The version check and the app both keep working in the cases marked *degraded* — that is what makes these invisible without this job.

| | |
|---|---|
| recipe | `vendor:com.workbuddy.workbuddy-ai:stable:arm64` |
| registry | vendor probe |
| endpoint host | `www.workbuddy.ai` |
| pattern | `"productVersion"\s*:\s*"([0-9]+\.[0-9]+\.[0-9]+)(?:\.[0-9]+)?"` |
| resolved | `5.3.14` |
| last good | `5.5.2` |
| consecutive sweeps | 2 |

**Warnings**

- installURLNotFound: HTTP 404 from codebuddy-1328495429.cos.accelerate.myqcloud.com
- Homebrew's cask `workbuddy-ai` is at 5.5.2.37849279-910352f0 while this recipe reads 5.3.14 — the probe may be stuck on a stale element
- version went BACKWARDS since the last sweep (5.5.2 → 5.3.14) — the pattern may have started matching a different element

<details><summary>Captured response (redacted, truncated)</summary>

```
{"version":"5.3.14.36279234","url":"https://codebuddy-1328495429.cos.accelerate.myqcloud.com/workbuddy/saas/darwin-arm64/WorkBuddy-darwin-arm64-5.3.14.36279234-825709d4.zip","productVersion":"5.3.14.36279234","sha256hash":"","timestamp":1787002434,"hash":"","name":"","supportsFastUpdate":false}
```

</details>

### Automated analysis

❔ No pattern proposed.

<details><summary>Unverified suggestion — do not apply without testing (model: `opencode-go/deepseek-v4-pro`, self-reported confidence 0.70)</summary>

The endpoint is serving a stale, month-old manifest (timestamp 2026-08-16) that still advertises 5.3.14.36279234 and a download URL whose object has been deleted from the CDN, which is why it 404s and the version appears to go backwards from 5.5.2. The regex itself still matches "productVersion" without trouble — the response body simply no longer contains the current release, so no pattern tweak can recover it.

Produced by a model with no tools, in an empty directory, from the captured response above. That response is third-party content and may be hostile; nothing here has been applied to any file.

</details>

Reproduce locally:

```bash
swift run --package-path CLI duo verify --vendor --only com.workbuddy.workbuddy-ai --samples
```

### LINKED ISSUE #738: Recipe degraded: com.workbuddy.workbuddy-ai (vendor probe) — installURLNotFound: HTTP 404 from…
Found by the scheduled `duo verify` sweep. The version check and the app both keep working in the cases marked *degraded* — that is what makes these invisible without this job.

| | |
|---|---|
| recipe | `vendor:com.workbuddy.workbuddy-ai:stable:x64` |
| registry | vendor probe |
| endpoint host | `www.workbuddy.ai` |
| pattern | `"productVersion"\s*:\s*"([0-9]+\.[0-9]+\.[0-9]+)(?:\.[0-9]+)?"` |
| resolved | `5.3.14` |
| last good | `5.5.2` |
| consecutive sweeps | 2 |

**Warnings**

- installURLNotFound: HTTP 404 from codebuddy-1328495429.cos.accelerate.myqcloud.com
- Homebrew's cask `workbuddy-ai` is at 5.5.2.37849279-910352f0 while this recipe reads 5.3.14 — the probe may be stuck on a stale element
- version went BACKWARDS since the last sweep (5.5.2 → 5.3.14) — the pattern may have started matching a different element

<details><summary>Captured response (redacted, truncated)</summary>

```
{"version":"5.3.14.36279234","url":"https://codebuddy-1328495429.cos.accelerate.myqcloud.com/workbuddy/saas/darwin-x64/WorkBuddy-darwin-x64-5.3.14.36279234-825709d4.zip","productVersion":"5.3.14.36279234","sha256hash":"","timestamp":1787002434,"hash":"","name":"","supportsFastUpdate":false}
```

</details>

### Automated analysis

❔ No pattern proposed.

<details><summary>Unverified suggestion — do not apply without testing (model: `opencode-go/deepseek-v4-pro`, self-reported confidence 0.60)</summary>

The regex didn't stop matching — it still finds `"productVersion":"5.3.14.36279234"` and captures 5.3.14 — but the vendor's update manifest itself is stale: its timestamp (1787002434 ≈ 2026-08-18) is about a month old, it has regressed from 5.5.2.x to 5.3.14.36279234, and its download URL returns HTTP 404. The true newest release (5.5.2.37849279, per Homebrew) appears nowhere in this body, so no regex change can fix the backwards version; the probe's vendor endpoint (or the vendor's feed) needs to be corrected instead.

Produced by a model with no tools, in an empty directory, from the captured response above. That response is third-party content and may be hostile; nothing here has been applied to any file.

</details>

Reproduce locally:

```bash
swift run --package-path CLI duo verify --vendor --only com.workbuddy.workbuddy-ai --samples
```

jizhi0v0:
_Requested by **Jizhi0v0** · [project thread](https://claude.ai/code/project/chan_01XkAMWWU8faP2SRifP8Y4Z1?thread=cmsg_01XkAMWWU8faP2SRifP8Y4Z12zHm2eRkwY2zReobHMz6od)_

Closes #737, closes #738, closes #736, closes #698, closes #467, closes #713.

Five apps, four distinct causes. Two were real detection bugs; three were the `BACKWARDS` check asking a question the recipe cannot answer.

## WorkBuddy (#737, #738) — plus two recipes that were wrong in silence

`/v2/update` answers with the **next hop on an upgrade chain**, not the latest build. Measured on `www.workbuddy.ai` 2026-09-18:

| request | `productVersion` |
|---|---|
| no `version` param | `5.5.2.37849279` |
| `&version=5.3.14` | `5.5.2.37849279` |
| `&version=5.0.0` / `1.0.0` / `0.0.0` | `5.3.14.36279234` |

`version=0.0.0` was pinned to dodge the 204 that an up-to-date version gets, and it read correctly only while the chain's first hop *was* the newest build. Once the vendor shipped above it, **all four** recipes (2 sites × 2 arches) froze on the hop.

Only the two `.ai` ones ever complained, and **not because the version was wrong** — only because that hop's artifact was later deleted from the CDN, producing the 404. The `.cn` pair kept a live artifact and sat green at 5.3.14 while the CN train shipped 5.5.6 — a number the *same app's changelog row* held a few lines away in the same `baseline.json`. Nothing compares those two rows.

Fix: send no `version` at all, which is what Homebrew's `workbuddy-ai` cask livechecks. All four now read 5.5.2 / 5.5.6 live.

Also checked for the same shape elsewhere: ChatWise is unaffected (identical answer with or without the param), and CapCut pins `9.99` — *above* every release, so it cannot land on a hop.

## Figma (#736) — prose compared as versions

The `version` group captures a post **title**; Figma doesn't number its notes. Publishing a post filed an issue, and every post after it would file another.

The identical guard already existed in `Verify.changelogLagComplaint` — whose comment names Figma — and was never carried across to `Baseline`. Carried across now, as **"either side carries a digit"** rather than "leads with a digit":

- Ten baseline rows lead with a non-digit; three are prose (Figma's post titles, Cursor's `Sep 10, 2026`, Kiro's empty string) and **five** are real versions behind a prefix (`Build 4200`, `Build 2130`, `V16.6.0.32198`, `v2.0.11.1`, `Xcode 27`). Leading-digit would have switched the check off for all five, silently. The remaining two are Superconductor's `bca46eb1` twice — commit hashes, which never reach this guard because `ordersByLineage` is checked first at the call site.
- **EITHER, not BOTH**, so Blender's `5.1` → `LTS` stays reported — that is a pattern that has stopped finding versions, not a vendor that never numbered anything. An existing test (`aVersionWithNoDigitKeepsTheCheck`) pins it, and caught the BOTH version of this change.

## TypeWhisper (#698) — a date-ordered page across two trains

The page interleaves macOS stable and daily builds in one list, so shipping 1.6.1 above `1.7.0-daily.20260916` moved the newest entry backwards with nothing wrong (the recipe's own comment already says a stable install sees dailies above its release).

A backwards version is now reported only when the baseline's version is **no longer on the page**: published-above keeps it one row down, a slipped pattern drops it off the top. `Finding` carries `entryVersions` for this.

## Qoder IDE (#467) — the vendor is the stale one

Notes top out at 1.29.0 (Sept 8) while the probe reads 1.30.1. `acknowledgedStaleEntry: "1.29.0"` — named, not switched off, per that field's contract.

## iStat Menus (#713) — no code change

#699's follow-up moved the recipe to reading the archive's own `Info.plist`, whose `7.50` is correct; the baseline kept the filename-derived `7.50.1`. A regressed value is deliberately never written back, so the complaint could never clear. The audit's own closing line called for this edit and it was missed.

## Verification

Against the live endpoints with the installed CLI:

- Swept against a copy of the **pre-fix** baseline: the three changelog complaints are already gone, only iStat's remains (its fix is the datum).
- Against the fixed baseline: **7 rows ok, 0 warnings**, every streak reset to 0.
- `make test` green — 3355 core + 354 CLI.
- Each new guard mutation-tested: deleting it, and flipping EITHER→BOTH, each fail their test.

## Review round 1 (`b654afe`)

Three blocking findings, each reproduced before fixing:

1. **`Finding.entryVersions` was dropped by `observing(_:)` and `adding(warning:)`** — the same silent regression both methods' comments were written about for `entryCount` (every argument there has a default). It hit exactly the findings that carry a complaint, which `Verify.foldingBaselineComplaints` rebuilds through `adding(warning:)`, so `report.json` said "not recorded" for the one field that tells a slipped pattern from a vendor publishing above the baseline. The in-process verdict was unaffected — `reconcile` runs on the original finding, before the fold — so this was lost diagnostics, not a wrong sweep. Both rebuilds now forward it, pinned by `annotatingAFindingKeepsThePageItRead` and mutation-tested by dropping the argument from either call.
2. **The CN WorkBuddy audit still described the `version=0.0.0` pin in the present tense**, as the fix rather than the bug. The `.ai` audit got a dated correction; the `.cn` one — same recipe factory, and the pair that actually sat two releases behind — got none. Corrected in place, with the upgrade-chain explanation and a pointer to the full measurement table.
3. **The digit-guard comment miscounted its own measurement.** Re-counted against `verify/baseline.json`: of the ten non-digit-leading rows, five are versions behind a prefix, not six, and the two it folded into "the rest are versions" are commit hashes `ordersByLineage` already exempts. Comment and the Figma section above both corrected.

https://claude.ai/code/session_01JtBbjUJKy8k6Z7aLZM9UCW


jizhi0v0:
All three addressed in `7a8a09c2` (merging the Action's concurrent fix `b654afe2` with `85185fb7`).

1. **`entryVersions` dropped by both rebuilds** — real, and the finding understates it slightly: `entryCount` *already had* a targeted test (`annotatingAFindingKeepsItsEntryCount`) and a comment saying "every field forwarded", and `entryVersions` was dropped anyway. So the comment and the per-field test were both already in place and neither prevented it. Forwarded in both rebuilds, and added `FindingRebuildTests`, which walks `Mirror` over a fully-populated finding and both rebuilds and compares every child — mutation-tested against four dropped fields including `gatewayRetries`, which no test names. The Action's targeted test is kept alongside it.

2. **Stale `.cn` audit** — correct, and worse than stated: that file already recorded the CN changelog at 5.5.6 and the CN probe at 5.3.14 *in the same paragraph* on 2026-09-14. Sentence repaired in place, and the appended note now carries the transferable part — no check in the sweep compares an app's probe row against its own changelog row, so a recipe that reads wrong *consistently* trips nothing.

3. **Miscount** — confirmed, mine. Recounted programmatically rather than by eye this time: five versions behind a prefix, three prose/empty, two commit hashes already exempted by `ordersByLineage`. "The rest are versions" and "all six" were both wrong.

On the unverified external claims: every measurement cited was taken live in this session, and the end-to-end result is in the PR body — swept against a copy of the pre-fix baseline, then the fixed one, 7 rows ok / 0 warnings. The one I did *not* independently verify is CapCut's `9.99` sitting above every published 9.x build; that sentence rests on the pre-existing claim in the recipe, and only adds that a pin above every release cannot land on a hop.

jizhi0v0:
Round 2's blocking finding is correct, and was already fixed before the review ran — `00ee5fa9` sits on top of `7a8a09c2`, the sha round 2 scoped. I'd caught the same error myself while writing up the follow-up issue: I asserted "nothing compares the probe row against the changelog row" without reading `changelogLagComplaint` first, which is precisely the mistake the repo's own rules exist to prevent. Both audits now say the check exists and is one-directional, and #743 is filed against the direction gap rather than the imaginary missing check.

The `Verify.brewComplaint` pointer at the end of the finding was the most useful line in it, and it led somewhere the review didn't go — `f841b950`:

`brewComplaint` **fired correctly** for the `.ai` pair (#737/#738 carry `Homebrew's cask workbuddy-ai is at 5.5.2… while this recipe reads 5.3.14`). It was silent for `.cn` **not for want of a cask**: `workbuddy-cn` exists and has been right all along, reading `5.5.6.38337834` — exactly the version our probe should have had. It missed on the **key**. Brew's bundle-id table is built from each cask's `uninstall quit:`, which `workbuddy-cn` sets to `com.tencent.workbuddy.mac` against our measured `com.workbuddy.workbuddy`.

So the contrast is sharper than either of us had it: the `.ai` pair tripped three signals, the `.cn` pair tripped zero — one gate pointed the wrong way, the other keyed on an id it could not match. A silent key miss is indistinguishable from "no cask exists", which is its own detectable condition. #743 updated with both halves.

On the unverified claims: the `.cn` endpoint measurements are mine from this session and reproducible with the `curl` lines in the audit. You're right about the test count — 357 was the pre-merge tree; the merged tree is 358, since both the targeted and the `Mirror` test were kept. The Outlook failure is `officecdn.microsoft.com` refusing TLS from this machine (3/3 `unexpected eof while reading`, verified by `curl` outside the test); no file in this diff touches Outlook, and CI is the arbiter there.

jizhi0v0:
Round 3's finding is correct and now fixed in `5d8932c4`. I introduced the contradiction: the section I added in `f841b950` said the CN cask exists while the 覆盖矩阵 in the same document still read `—` / `无 cask`. You're right that in this doc family `无 cask` is a claim backed by a 404 lookup, not a shorthand for "our lookup misses it" — `com-qoder-ide.md` does it properly.

**Both** matrices were wrong, not just the CN one, and as you note the international one was falsifiable from inside the repo: that recipe's `lastSignature` opens with ``Homebrew's cask `workbuddy-ai` is at 5.5``. Both Homebrew columns are now `○` with the correction recorded.

Your `auto_updates` question was the most valuable part — it decides whether fixing the key buys anything. Checked both: **neither cask sets it**, so `!cask.autoUpdates` doesn't block either, and for the CN pair the key mismatch is the sole blocker. Values, for the record:

| cask | version | `auto_updates` | `uninstall quit:` |
|---|---|---|---|
| `workbuddy-cn` | `5.5.6.38337834-5f969292` | not set | `com.tencent.workbuddy.mac` |
| `workbuddy-ai` | `5.5.2.37849279-910352f0` | not set | `com.workbuddy.workbuddy-ai` |

Worth flagging a trap I hit while checking: `uninstall quit:` lives under `artifacts` in the `formulae.brew.sh` JSON, and the top-level `uninstall` key reads `null`. My first pass read that null and would have concluded the cask sets no quit id at all.

The two claims you marked unverified are now reproducible rather than asserted:

- **the 50** — `verify/baseline.json` has 56 bundle ids carrying both a `changelog:` and a `vendor:` row; restricted to `lastGoodVersion` leading with a digit (the precondition for a major.minor comparison to mean anything) it's 50. The filter is now stated in the doc, so it derives from the committed baseline alone.
- **the complaint string** — the full sentence is quoted from the #737/#738 issue body; the baseline keeps `lastSignature` truncated at `is at 5.5`. The doc now says which source each comes from.

The matrix error is folded into #743 as part of its scope, because it's the same fault wearing a different hat: a silent key miss is indistinguishable from "no cask exists" — including to a human writing the audit. That's how `无 cask` got written in the first place.

jizhi0v0:
Round 4: zero new blocking findings, and round 3's is confirmed fixed. Two of the "unverified" items turned out to be wrong, though — both mine — so they're now corrected in `67c78c05`.

**The 5 was a methodology error, not a stale reading.** You were right that it isn't reproducible from the committed baseline, and the reason is worse than staleness: my comparison took an *arbitrary* probe row — the first in an unordered list — so multi-channel apps were compared against whichever row came first. That manufactured two of the five:

- `org.mozilla.thunderbird` — compared against its **ESR** row while its stable row reads `156.0`, matching the changelog exactly.
- `org.mozilla.thunderbirdbeta` — `157.0beta` vs `157.0b2`, one release spelled two ways.

Under "the changelog reading is above **every** probe row", it's **3**: WorkBuddy CN, Claude for Desktop, Obsidian — which is your highest-vs-highest number. That's now the stated rule, so it derives from the committed baseline.

This makes the case *better*, not worse: 1-real-in-3 still argues against a naive reversal, and two of the four discriminators I'd asked #743 to implement were chasing my own artifacts. Multi-channel awareness and spelling normalisation both fall out of the correct rule for free; only the two genuine classes remain (two build namespaces, another train's entries on the page). #743 updated with the corrected table and the shorter rule.

**The quote was not complete.** You were right to push on that — `Verify.swift:1099-1101` does append `— the probe may be stuck on a stale element`, and #738's body carries that ending. I'd truncated it and then explicitly called the truncation "完整的一句". Quoted in full now, with the correction noted inline.

On the remaining external claims: the `workbuddy-cn` fields (`5.5.6.38337834-5f969292`, `quit: com.tencent.workbuddy.mac`, `auto_updates` null) and `workbuddy-ai`'s quit id are live `formulae.brew.sh` reads from this session, reproducible with the `curl` in the audit — and as you note, the `.ai` quit id is corroborated from inside the repo, since that complaint could not exist unless the lookup resolved. The 2026-08-27 `channel-verify` run predates this branch; I've left it attributed as an on-machine result rather than restating it as fact.

jizhi0v0:
Round 5's finding is correct and now fixed in `63dbdeb8`. It caught something I should have caught myself from data I was already holding.

`org.mozilla.thunderbirdbeta` has exactly **one** probe row, so "above an arbitrary probe row" and "above every probe row" are the same rule for it — the rule change I credited *cannot* have removed it. What removed it was a bug in my analysis: I approximated `VersionComparator` in Python, and the approximation truncated both `0beta` and `0b2` to `0` and called them equal. Your trace of the real path is exactly right — `majorMinor` leaves both intact, tokenization gives `[157, 0, "beta"]` vs `[157, 0, "b", 2]`, and `"beta" > "b"` compares as text, so `isNewer` is true.

Recomputed by calling the actual `VersionComparator` over the committed baseline: **4 hits** — WorkBuddy CN, Claude for Desktop, Obsidian, Thunderbird Beta. One in four is a real fault. The non-beta `org.mozilla.thunderbird` *is* correctly excluded by the every-row rule (stable row equals the changelog at `156.0`), so that half of my last correction stands.

You also identified the more damaging half, which I'd have missed: on the strength of that faulty 3 I **removed** 「同一个版本的不同写法」 from the list of classes a reversed check must learn — in the same commit that claimed the list had got shorter. It's restored, and it's now the stated reason Thunderbird Beta is on the table rather than an argument for leaving it off. #743 carries the corrected four-row table and the three required classes.

The generalizable part, recorded in the audit: **don't approximate a production comparator in an analysis script.** Both of my wrong numbers came from modelling `Verify`'s comparison in Python rather than calling it — the first by picking an arbitrary probe row, the second by collapsing `0beta`/`0b2`. To ask whether production would complain, run production's code over the baseline, which is what the corrected table does.

CLI suite still green at 358.

PR #744, opened 2026-09-18
verify: cross-check the changelog against the probe in BOTH directions, and stop the cask lookup missing on the key

### LINKED ISSUE #743: duo verify: the probe↔changelog cross-check is one-directional, so a frozen probe is invisible
## The gap

`duo verify` **does** cross-check an app's changelog against its probe — `Verify.changelogLagComplaint`, fed by `knownVersions`, which is built from *this sweep's own findings* (`Verify.swift:189-194`), so a probe row and a changelog row for the same app are compared whenever both registries sweep together.

That check is **one-directional**. It fires only when the changelog **trails** the probe:

```swift
// Verify.swift:1261
guard entryMM != detectedMM, VersionComparator.isNewer(detectedMM, than: entryMM)
else { return nil }
```

The ignored direction — the changelog reading **ahead** of the probe — is exactly what a **frozen or stuck probe** looks like. Nothing reports it, and none of the history checks can: they all compare a row against *its own* previous value, and a recipe that reads wrong *consistently* never moves.

## What it cost (#737, #738)

WorkBuddy's `.cn` pair sat green at `5.3.14` for weeks while the CN train shipped `5.5.6`. Both numbers were in the same sweep's hands — identical `lastGoodAt` in `verify/baseline.json`:

```
changelog:com.workbuddy.workbuddy:-          lastGoodAt=2026-09-17T20:25:48Z  version=5.5.6
vendor:com.workbuddy.workbuddy:stable:arm64  lastGoodAt=2026-09-17T20:25:48Z  version=5.3.14
vendor:com.workbuddy.workbuddy:stable:x64    lastGoodAt=2026-09-17T20:25:48Z  version=5.3.14
```

`docs/app-audits/com-workbuddy-workbuddy.md` had *also* recorded both contradicting numbers in one paragraph on 2026-09-14. The `.ai` pair of the same four-recipe family was caught only by luck — the stepping-stone artifact it was frozen on was later deleted from the CDN, producing a 404. Had the vendor left that file in place, all four would still be wrong.

## Why the obvious fix is wrong

Do **not** just drop the `isNewer` guard. Measured against `verify/baseline.json` on 2026-09-18 — 50 apps carry both a probe row and a changelog row, and **5** have the changelog ahead at major.minor:

| app | changelog | probe | real? |
|---|---|---|---|
| `com.workbuddy.workbuddy` | 5.5.6 | 5.3.14 | ✅ the bug |
| `com.anthropic.claudefordesktop` | 2.2553.0 | 2.110.1 / 1.46388.3 | ❌ two build namespaces |
| `md.obsidian` | 1.14.2 | 1.13.7 | ❌ insider entries on the page (brew agrees at 1.13.7) |
| `org.mozilla.thunderbird` | 156.0 | 140.16.0esr / 156.0 | ❌ multi-channel; one probe row matches |
| `org.mozilla.thunderbirdbeta` | 157.0beta | 157.0b2 | ❌ same version, different spelling |

**1 real out of 5.** A naive reverse check is ~80% false-positive on today's data, and this sweep's whole value is that it doesn't cry wolf.

## What a sound version needs

The four false positives are each a known, already-modelled situation, so the discrimination exists — it just has to be applied:

- **Channel/namespace awareness** — compare a probe row only against a changelog row for the *same* channel, and skip apps whose changelog page carries another train's entries (the Obsidian-insider and TypeWhisper-daily shape; `ChangelogRecipe.channel`, `minimumAppVersion`/`belowAppVersion` already express this).
- **Skip `vendorBuildVersion` apps** — Claude for Desktop's two numbering namespaces are documented and must never be compared (see `duo-updater-two-build-namespaces`).
- **Normalise spelling before comparing** — `157.0beta` vs `157.0b2` is one release.
- **Require persistence** — a real freeze lasts; a publishing-order skew resolves in a sweep or two. The existing `consecutiveActionable` / `acknowledgedStaleEntry` machinery is the right shape, the latter especially since "the vendor's notes run ahead" is a legitimate steady state for some apps.
- **Corroborate with Homebrew where available** — the existing cask cross-check already resolves ~35% of bundle ids and would have settled all four false positives here.

## Acceptance

The check, run against the current baseline, should flag `com.workbuddy.workbuddy` (using its pre-#740 value of `5.3.14`) and **none** of the other four.

---

Found while fixing #737/#738 in #740. Not blocking that PR — the four WorkBuddy recipes are fixed and verified there; this is about the class of failure they belong to.

jizhi0v0:
Closes #743.

Both halves of the issue: the **direction gap** in `changelogLagComplaint`, and the **key gap** in `brewComplaint`.

## 1. The direction

`changelogLagComplaint` only ever fired when a changelog TRAILED its probe. The ignored direction is what a frozen probe looks like from the outside, and no history check can cover for it — they all compare a row against its *own* previous value, and a recipe that reads wrong *consistently* never moves.

`Verify.changelogLeadsProbeComplaint` is that direction. It compares against **every** probe row for the bundle id (which takes multi-channel apps out for free — Thunderbird's stable row equals its changelog) and excludes the three already-modelled situations a naive reversal misreads:

| class | case | how |
|---|---|---|
| two numbering namespaces on one channel | Claude for Desktop's GA redirect vs its Squirrel rollout endpoint (2.x vs 1.x, both `.stable`) | a channel whose rows disagree with *each other* is not one ordered train → skip the app |
| another train's entries on the page | Obsidian's insider builds | new `ChangelogRecipe.carriesOtherTrainEntries` |
| one release spelled two ways | `157.0beta` vs `157.0b2` | `Verify.numericMajorMinor` cuts each component at its first non-digit |

**Acceptance, run over the committed baseline with production's own code** (not a Python approximation — that is what produced both of the issue's wrong numbers):

```
FLAGGED 1:
  · com.workbuddy.workbuddy:- — newest changelog entry (5.5.6) reads AHEAD of
    every probe row (5.3.14, 5.3.14)
```

Each discriminator is load-bearing — switched off one at a time, each brings back exactly its own case and no other:

| guard off | HEAD baseline | `2095a06c` baseline |
|---|---|---|
| (none off) | 1 | 1 |
| mixed-train skip | 2 (+`md.obsidian`) | 2 (+`md.obsidian`) |
| namespace skip | 1 | 2 (+`com.anthropic.claudefordesktop`) |
| normalisation | 2 (+`org.mozilla.thunderbirdbeta`) | 2 (+`org.mozilla.thunderbirdbeta`) |
| all three off | 3 | **4** — the issue's table, reproduced |

(The namespace case needs `2095a06c`: at HEAD Claude's GA row has caught up with the docs page.)

## 2. The key

`brewComplaint` caught the international WorkBuddy pair and said nothing about the Chinese one — not for want of a cask. `workbuddy-cn` had been right the whole time (5.5.6 against our frozen 5.3.14), but it quits `com.tencent.workbuddy.mac` while our recipe is keyed `com.workbuddy.workbuddy`, so the bundle-id index simply had no such app. A silent key miss is indistinguishable from "no cask exists" — including to a human writing an audit, which is why both of that family's coverage matrices said `无 cask`.

`liveCasks` now falls back to the `.app` filename key (`com.workbuddy.workbuddy` → `workbuddy.app` → `WorkBuddy.app`; the catalog's filename index is case-folded), taking the answer only when it is **unambiguous**. Measured against the live catalog and the committed baseline:

- 138 vendor bundle ids: 60 resolve by id, **17 more** by filename — 43% → 55%.
- All 18 filename hits are the right app (each cask agrees with our reading or trails it). The one ambiguous case, `Telegram.app` (installed by both `telegram` 12.10 and `telegram-desktop` 7.2.9), is dropped by `unambiguousCasks`.
- Running the real `brewComplaint` over every baseline row against the live catalog: **exactly two new complaints, both the WorkBuddy CN rows.** No regressions.

A complaint reached through the guessed key says so, on both the "cask ahead" and the phantom direction.

Also fixed while in there: the upstream direction compared on a bare two-component split, so brew's `version,build` (`librewolf` ships `156.0,1`) and `version+revision` spellings read as a whole release ahead of an identical version. It had never fired only because every cask spelling versions that way sat behind a key this function could not reach — the fallback puts several of them in range.

## Verification

- `make test` — exit 0.
- Red→green on the real path, through the installed `duo`: flipping `carriesOtherTrainEntries` off makes `duo verify --only md.obsidian --vendor --changelog` warn with the new complaint; on it clears.
- Full live sweep (`--vendor --github --changelog`, 358 recipes, 207s): one warning the baseline did not have — `io.dcloud.HBuilderX`, whose notes carry `5.26.2026091702` while its `release.json` has gone back to `5.24.2026081301` (Homebrew's cask agrees with the probe; re-fetched four times, stable). That is real, and it is the *other* reading of this signal, so the complaint now names both: a stuck probe, **or** a release the vendor pulled whose notes it still carries. Worth its own look — not handled here.


jizhi0v0:
### Round 1 verdict — blocking finding accepted, plus one hazard from the unverified list

**Blocking · `caskKeyNote` names a stanza the cask may not have — correct, fixed in `8adf2299`.**

I measured it rather than taking it on reasoning: of the 39 casks the filename fallback reaches across both registries, **36 declare no `uninstall quit:` id at all**; only `mstystudio` (`com.MstyStudio.app`), `headlamp` (`com.kinvolk.headlamp`) and `workbuddy-cn` (`com.tencent.workbuddy.mac`) declare a competing one. So the wrong half of the sentence was the *common* case, not an edge. The note now says the cask "does not declare that bundle id in its `uninstall quit:`" — the only thing the branch actually knows. `BrewCaskKeyTests.aFilenameMatchedCaskComplainsAndSaysSo` now asserts both halves, including `!complaint.contains("declares a different")`, which is the assertion whose absence let this through.

**From the unverified-claims list: `app.app` / `desktop.app` — a real latent hazard, also fixed.**

Confirmed the premise: 35 of the 228 cross-checked bundle ids derive a qualifier rather than a name (`bot.cline.app` → `app.app`, `com.google.Chrome.beta` → `beta.app`, `dev.kiro.desktop` → `desktop.app`, `com.termius-beta.mac` → `mac.app`). Confirmed none resolves today — no cask in the live catalog ships `App.app`, `Desktop.app`, `Mac.app`, `Client.app` or `Dev.app` — but that is luck, not a property worth resting a filed issue on. `caskAppFilename` now returns nil for a qualifier, with the channel half taken from `ReleaseChannel.allCases` rather than a hand-copy.

Writing the test for that caught a second bug in my own guard: the raw values were compared against a lowercased component, so `ReleaseChannel.guineaPig` slipped through. Both sides are folded now.

**Also correct: the measurement missed `GitHubReleaseRule`.** It is `BrewCrossChecked` too, and my numbers came from the baseline's vendor rows alone. Re-taken over both registries against the live catalog:

| | |
|---|---|
| cross-checked bundle ids | 228 (138 vendor probes, 90 GitHub rules, no overlap) |
| resolve by bundle id | 88 |
| resolve by the `.app` fallback | 39 more — **39% → 55%** |
| qualifier stems now refused | 35 (was 0 refused, 0 resolving) |
| complaints over the 259 recipes/rules with a baseline version | **2 before → 2 after**, both WorkBuddy CN; no regressions |

Every doc-comment number in the PR has been replaced with these. The other items on the unverified list (the `156.0,1` spelling, the Telegram ambiguity, HBuilderX, Obsidian's insider entries) I had measured against the live sources before filing; HBuilderX is now [#745](https://github.com/jizhi0v0/duo-updater/issues/745), where the artifact check adds something the PR body got wrong — both dmgs are still on the CDN, so DCloud rolled back the *config*, not the release.

`make test` exit 0.

jizhi0v0:
### Round 2 verdict — both findings accepted, fixed in `df974e1d`

**Finding 1 · a doc still said "withdrawn-release" after the message stopped saying it.** Correct, and the worse version of the mistake: `52b441e2` changed the message *because* I had measured that the release was not withdrawn, then left the sentence above it asserting the opposite. Fixed, with the artifact check itself recorded in the doc so the next reader doesn't have to redo it.

**Finding 2 · the qualifier enumeration was incomplete, and the count was wrong too.** Re-measured over the two registries with `qualifierComponents` as it now stands:

```
TOTAL ids 228; refused 52 across 10 stems
  desktop ×16 [hand-list]   app ×16 [hand-list]   beta ×5 [ReleaseChannel]
  mac ×5 [hand-list]        nightly ×3 [RC]       dev ×2 [RC]
  client ×2 [hand-list]     canary ×1 [RC]        macos ×1 [hand-list]
  preview ×1 [RC]
```

So it is **52 across ten stems**, not 35 across six. Both of my numbers came from the hand-written list in the throwaway measurement script, which I ran *before* deriving the channel half from `ReleaseChannel.allCases` and never re-ran afterwards — the count described a set the code no longer used.

The trigger you named is the real cost, and it's sharper than the comment states: `macos` and `client` are carried by **one bundle id each** (`com.raycast.macos`, `com.spotify.client`), they're in the hand-written half with no `ReleaseChannel` derivation to restore them, and a reader auditing the set against my list would have found nothing at all behind `macos`.

A corrected sentence is not enough protection for that, so `aQualifierIsNotAName` now pins **one real registry id per stem** — Raycast and Spotify among them — and a second test asserts those ids are still in the registries, so the pins can't quietly become vacuous. Trimming `"macos"` now fails a test that names the app it would break, rather than silently putting Raycast back on the `macos.app` key.

One correction to the evidence, not to the finding: `dev.warp.Warp.Stable` isn't in the registries — Warp's ids are hyphenated (`dev.warp.Warp-Stable`, `dev.warp.Warp-Preview`), so `stable` derives nothing and is among the ten entries in `qualifierComponents` that no current id exercises. The other six ids you cited are all real and all now covered.

`make test` exit 0.

jizhi0v0:
### Round 3 verdict — accepted, fixed in `2911a846`

Correct, and it contradicts the table four lines above it in the same comment. `macos` really is ×1 (`com.raycast.macos`); `client` is ×2 — `com.spotify.client` and `com.windscribe.client`, both `VendorProbeRecipe`s and so both cross-checked. My own round-2 measurement printed `client.app ×2 [hand-list] com.spotify.client, com.windscribe.client`; I wrote the prose from memory of the shape rather than from the output I had just produced.

The trigger is the part that matters: that sentence is what a reader follows *when the vacuity test fires*. Told Spotify was the only id behind the stem, they'd drop the pin and `"client"` together, and `com.windscribe.client` would start deriving `client.app` — exactly the cross-app accusation the same paragraph warns about for `macos`.

Fixed in all three copies (`Verify.swift`, `BrewCaskKeyTests`, the app audit). Both `client` ids are now pinned rather than one standing for the stem, since that's the entry whose count is easiest to misread, and `thePinnedQualifierIdsAreStillInTheRegistries` covers both. The doc now also says outright to read the counts from the table rather than from the prose beside it.

Worth naming the pattern, since it's three rounds running: every finding has been a **claim about the measurement**, never the measurement or the behaviour. The numbers have held up each time they were re-derived — what has drifted is the sentence describing them, twice because I re-ran the script and not the prose, once because I wrote the prose without looking at the script's output at all. The pinned-ids tests are the part of this round that survives that; the paragraph is now explicitly subordinate to the table.

`make test` exit 0.

PR #746, opened 2026-09-18
Read an app's binary by its bundle name when it never declared one

jizhi0v0:
## What broke

Meta's **Muse.app** (`com.meta.endo`) ships **no `CFBundleExecutable` at all**:

```
$ /usr/libexec/PlistBuddy -c 'Print :CFBundleExecutable' /Applications/Muse.app/Contents/Info.plist
Print: Entry, ":CFBundleExecutable", Does Not Exist
```

It launches fine, and the system resolves it anyway:

```
$ codesign -dv /Applications/Muse.app
Executable=/Applications/Muse.app/Contents/MacOS/Muse
$ swift -e 'print(Bundle(url: URL(fileURLWithPath: "/Applications/Muse.app"))!.executableURL!.path)'
/Applications/Muse.app/Contents/MacOS/Muse
```

`AppRuntimeDetector.executableURL` required the key, so it found nothing — and
with no binary there are no load commands, so the runtime, the linked frameworks
**and** the SDK all came back empty and the row drew no framework mark.

## The fallback, and what it is not

`CFBundleExecutable` where the packager wrote one; the **bundle directory's base
name** where they did not. That is what CFBundle itself was measured doing on
this Mac (macOS 27, 2026-09-18): given `Dir.app` whose plist says
`CFBundleName = Named` and whose `MacOS` holds *both* `Dir` and `Named`,
`Bundle.executableURL` answers `Dir`; delete `Dir`, leave `Named`, and it answers
nil rather than switching.

Apple's [Bundle Programming Guide](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html)
lists the key as **required and documents no fallback**, so this mirrors observed
CFBundle behaviour rather than a documented promise. Two things keep that from
becoming a guess:

- `fileExists` still has the last word on **both** paths.
- A *declared* executable that is missing still **fails closed** — the name
  fallback does not rescue it. That is the existing "reading a different file
  instead would be the guess it refuses to make" rule, unchanged.

`RuntimeVersion` had a second copy of the same resolution; it now calls the one
in `AppRuntimeDetector` rather than growing a second fallback.

## Verification

**Red → green on the real bundle**, through the production detector:

```
before: /Applications/Muse.app: runtime=nil    frameworks=[]                 binary=nil
after:  /Applications/Muse.app: runtime=native frameworks=[AppKit, SwiftUI]  binary=Muse
```

(`/Applications/QQMusic.app` was already `native / [AppKit, SwiftUI] / QQMusic`
and is unchanged — it is in the probe as the control.)

**Blast radius.** Across the **182** bundles in `AppScanner.defaultLocations` on
this Mac, exactly **one** lacks the key — Muse — and it is the one the fallback
rescues. Four others have no `Contents/` directory at all (`Amp 2`, `Aqara Home`,
`ClaudeUsageApp`, `Nowledge Mem`): wrapped iOS apps, which take the separate
`isiOSAppOnMac` path and never reach this function.

**Mutation-tested**, five new cases against four mutations — every mutation goes
red, and every new case is killed by at least one:

| mutation | tests that go red |
|---|---|
| drop the fallback | `aBundleThatNeverDeclared…`, `anEmptyExecutableKey…`, `theNameFallbackUsesTheBundleName…` |
| drop the empty-string check | `anEmptyExecutableKey…` |
| drop `fileExists` | `theNameFallbackStillRequiresTheFileToBeThere`, `theNameFallbackDoesNotRescueADeclaredExecutableThatIsMissing` |
| use `CFBundleName` instead of the directory name | `aBundleThatNeverDeclared…`, `anEmptyExecutableKey…`, `theNameFallbackUsesTheBundleName…` |

The empty-string check earns its own case: left alone, `""` builds
`Contents/MacOS/` and `fileExists` says **yes** to that directory, so the reading
would describe a folder.

`make test` is green end to end (Core 3360 tests, CLI 358, app tests 51 of 51,
plus every checker).

## Not verified here

The menu row itself was not screenshotted — the popover only opens on a human
click, and this change is a data change (`runtime` nil → `.native`), not a layout
change. Verification is at the detector, on the real bundle.

## Out of scope, found on the way

Muse's Sparkle feed (`https://www.facebook.com/endo/release/appcast.xml?channel=production`)
is anonymous-readable and already drives `duo check` (`source: Sparkle`,
up-to-date at 1.0 / 1070843164), with a dmg `<enclosure>` + `edSignature`, so
one-click is in place. It carries **no `<description>` and no
`<sparkle:releaseNotesLink>`** — the missing changelog is upstream, not ours.
`HatchReleaseChannel` looks like a channel but is not one: `channel=` set to
beta/alpha/internal/dev/canary/staging/test all return a ~68 KB Facebook HTML
page under HTTP 200, not an appcast.


jizhi0v0:
**CI triage — `test` failed on a network timeout, not on this change. Re-run requested.**

Both failures are in `CarbonCopyClonerProbeRecipeTests`, which probes bombich.com live, and both are the same transport error:

```
✘ ccc6InstallResolvesThroughTheCcc6EndpointLiveNotCcc7()
  CarbonCopyClonerProbeRecipeTests.swift:393:6: Caught error:
  ProbeFailed(bundleID: "com.bombich.ccc",
    failure: .transport(urlErrorCode: -1001, "The request timed out."))

✘ betaEndpointStillResolvesAVersionLive()
  CarbonCopyClonerProbeRecipeTests.swift:627:27: Expectation failed: outcome.remote.shortVersion
  ↳ outcome.remote → nil          # the same timeout, arriving as a nil remote
```

Each took **17.3 s** — the timeout, not a parse.

Evidence it is not the diff:

- The whole suite passes here against the same live endpoints: **27 of 27 in 1.8 s**, including both live cases.
- `main` was green at 05:04Z; this run failed at 06:05Z on the same `xcode-27` runner.
- The diff touches `AppRuntime.swift`, `RuntimeVersion.swift` and `AppRuntimeDetectionTests.swift` — nothing in the CCC recipes and nothing on a network path. `make test` was green end to end locally before the push (3360 Core / 358 CLI / 51 of 51 app tests).

Nothing to fix in the branch, so I re-ran the failed job rather than pushing a commit.

jizhi0v0:
**Author adjudication — round 1: 0 blocking findings, nothing to change. The three "unverified external claims" are verified below on this machine (macOS 27.0, build 26A428).**

### 1. `AppRuntime.swift:646` — Muse has no `CFBundleExecutable`, launches, and `codesign` names `Contents/MacOS/Muse`

```
$ /usr/libexec/PlistBuddy -c 'Print :CFBundleExecutable' /Applications/Muse.app/Contents/Info.plist
Print: Entry, ":CFBundleExecutable", Does Not Exist

$ codesign -dv /Applications/Muse.app
Executable=/Applications/Muse.app/Contents/MacOS/Muse
Identifier=com.meta.endo

$ pgrep -lf '/Applications/Muse.app/Contents/MacOS/Muse'
8876 /Applications/Muse.app/Contents/MacOS/Muse
```

All three halves hold. "Launches fine" is not inferred from the bundle — the process is running right now under that path.

### 2 and 3. `AppRuntime.swift:651` / `AppRuntimeDetectionTests.swift:833` — CFBundle picks the directory name, not `CFBundleName`

One claim, restated in two places; measured in both directions on a throwaway `Dir.app` whose plist carries `CFBundleName = Named` and **no** `CFBundleExecutable`:

```
--- A: both present (CFBundleName = Named) ---
MacOS contains: Dir Named
  executableURL -> Dir

--- B: only Named present ---
MacOS contains: Named
  executableURL -> nil
```

A is "prefers the directory name over `CFBundleName`"; B is the direction that matters more — with the directory-named file gone it answers **nil** rather than falling through to `CFBundleName`. That is exactly the rule the fallback implements.

**Standing caveat, unchanged:** Apple documents `CFBundleExecutable` as required and [documents no fallback at all](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html). The above is an observation of this machine's CFBundle, not a promise, which is why both the code comment and the test comment say so and why `fileExists` still has the last word on both paths. If a future macOS changes it, the fallback degrades to "no binary found" — the behaviour that shipped before this PR — rather than to a wrong answer.

### CI

The earlier `test` failure was two live bombich.com probes timing out (`-1001`, 17.3 s each); triaged in the comment above, job re-run, currently in flight.

PR #749, opened 2026-09-18
Add 千问输入法 (QianwenIME): detection + one-click

jizhi0v0:
Adds 千问输入法 (QianwenIME, `com.qianwen.inputmethod.desktopime`) — detection **and** one-click. Fourth input method in the registry, after 搜狗 / 豆包 / WeType.

Chatterfly (the other app this branch is named for) is **not** here: its update host is a runtime-dispatched pseudo-domain and it is still invite-only closed beta, so there is nothing to write yet. Reasoning is parked locally; revisit when it goes public.

## Why this endpoint

Its own update check is a Quark PUDS protobuf, WSG-encrypted in both directions (`puds.qianwen.com/upgrade/index.xhtml?from=pb_query`) — no plaintext shape to regex, and that endpoint is also the per-device allocation channel. So the probe reads what the official download button reads:

```
download.qianwen.com/pcdownload/qwenimemac?ch=pcqwenime@homepage_official&platform=mac
→ {"data":{"fp":"dapi-<uuid>","url":".../QwenimeMac_V1.2.26.41_mac_pf8002_(zh-cn)_release_(Build3192975).dmg"}}
```

Two things a future reader needs:

- **`ch` is load-bearing and falls back silently.** An unrecognised value is not an error, it is `@default`, which names a build from the previous version scheme. The pinned value is the vendor's own homepage channel — what every bundle from that page carries as `QianwenChannelId`. If it is ever retired the probe still returns 200 with that old build; `duo verify` catches that as `remote is BEHIND the installed copy`, so the failure is loud.
- **Not `ime.qianwen.com/api/download-config`.** It looks like the obvious version source and is wrong three ways: its `mac` entry states a version that disagrees with the version in its own `url`, its `updatedAt` is frozen, and the page that serves it ignores the entry entirely whenever buwang is on — which it is.

Version scheme: the filename's four-segment number is `CFBundleVersion`, so `versionIsBuild` compares build to build and `displayVersionPattern` keeps the row on the three-part marketing version. No changelog — the vendor publishes no notes page and the notes ride inside the encrypted payload.

## Why one-click is safe here

This app ships two scripts, and reading only the installer's gives the opposite answer:

| script | where | what | which path |
|---|---|---|---|
| `macos_install_core.sh` | installer stub | `rm -rf` the bundle, `mv`, re-register the input source | first install |
| `update.sh` | **inside the installed bundle**, named in `QianwenIMEService`'s integrity check | rotates `Contents` through `Contents_update` via a helper whose only filesystem import is `renameatx_np` | **update** |

So `InPlaceSwap.rotateContents` is the same shape as the vendor's own update, not a substitute for it. The artifact takes the DoubaoIme step: the DMG's only root app is an installer stub, the real app is at `Contents/Resources/QianwenIME.zip`, so `nestedArchivePath` unwraps one level once the stub's signature and Team are proved, and every gate re-runs on the payload. No `checksumPattern` — the response carries no digest, and the stub's `PayloadSHA256` is the wrong hash type for the wrong object.

## Verification

- `channel-verify` on the installed bundle: detects `stable`, probe answers `1.2.26` / build `1.2.26.41`, winning source `Vendor`
- live download through the whole gate stack: notarized Developer ID, Team and bundle id match the installed copy, universal
- **real red→green**: an older build rotated in with the vendor's own `update.sh`, then `duo install`. Outer `.app` inode unchanged, `com.apple.HIToolbox` input-source registration byte-identical, user data intact, rollback point + its user-data snapshot written beside it. It declined to kill the running input method and printed `duo restart` instead.
- `make test` green on the rebased commit

## One thing left open

After the rotation the top-level `Contents` is `755` while every directory inside it is `775`. That is `InPlaceSwap` behaviour, not this recipe's — it equally affects DoubaoIme and WeType — and I could not localize the cause from outside the code. No known functional effect here (the directory is user-owned, so the vendor's own updater can still rename and clean up). Recorded in the audit and being investigated separately; flagging it rather than letting the commit read cleaner than the evidence.


jizhi0v0:
Round 1 裁定。0 blocking，同意。两条「不够格」的注记里有一条我不同意放着，已修；「未验证外部断言」逐条在本机核过。

## 采纳：被本 diff 改成假话的既有注释（3 处）

复审判为非阻塞（「neither guard depends on the count and no test pins it」）。技术上对，但这三句话现在是**错的**，而且正是这个 diff 弄错的——留着就是让下一个人读到假话。同 PR 修掉：

- `UpdatePolicy.swift:313` — 「Today's two recipes are both `.zip`」→ 三条，且 `.zip`/`.zip`/`.dmg`。改写时保留了这句话的论点（重要的是**没有一条是 `.pkg`**，所以走不到这里），只把数目和类型说对。
- `VendorInstaller.swift:226` — 「the one recipe that declares a nested payload」→ 两条。
- `VendorProbeTests.swift:2006` — 「the one recipe in the registry whose download is NOT the app」→ 两条之一，并点名另一条是 QianwenIME。

## 未验证外部断言 —— 逐条本机核

| 断言 | 怎么核的 |
|---|---|
| 响应形状，URL 是未转义的 JSON 字符串 | 抓下原始 body 存盘，三条 pattern 逐一跑过：`1.2.26.41` / `1.2.26` / 完整 DMG URL。**复审这条提得对**——`\/`-转义的 body 会让三条全废；活证据是 `duo verify --only qianwen --vendor` 用生产代码打真端点 `✓ 1 ⚠ 0 ✗ 0` |
| `ch` 回落 | 四次实测同分钟：`@homepage_official`→`V1.2.26.41`；省略 / `@beta` / `@inner` 三者一律 `V1.2.2.32` |
| 版本方案 | 装机 Info.plist `1.2.26` / `1.2.26.41`，解包 payload 的 Info.plist 同值；端点给四段 = `CFBundleVersion` |
| `normalizePlatformConfig` 忽略 `platforms.mac` | 读官网 JS 源码确认，且 `buwang.enabled` 在 `api/download-config` 的响应里为 `true` |
| changelog 四条 404 | `/changelog`、`/release-notes`、`/updatelog`、`/help` 逐条 curl，均 404 |
| `update.sh` 是 UPDATE 且做 Contents 轮换 | 不止读脚本——**跑了它**：自打印 `atomic swap succeeded current=…/Contents next=…/Contents_update`，外层 inode 不变。`nm -u QianwenIMEAtomicSwap` 与文件系统相关的导入只有 `_renameatx_np` |
| DMG 根只有安装器壳一个 app | 新旧两个 DMG 都挂载确认过。**顺带一个 recipe 不受影响但值得记的事实**：旧包壳名是 `QianwenIMEInstaller.app`，新包是 `千问输入法.app`——壳名在版本间变过，所以 `firstApp(in:)` 按名字排序取第一个这件事，这里安全只因为根下确实只有一个 |
| `PayloadSHA256` 哈希类型与对象都不对 | 读壳里的 `QianwenInstallerMetadata.plist`：是 SHA-256 十六进制，描述的是内层 zip 而非下载物 |
| `qianwen` cask 装的是聊天客户端 | `brew info --cask qianwen` → `Qianwen.app`，`www.qianwen.com/qianwen` |
| `replaceItemAt` 取被替换方的 mode | 独立 Swift 探针，775/755 两个方向各跑一次 |

## 755 那条：同意是既有问题，但它**不是**我先前猜的那个

复审说它在 `rotateContents` 的 `mode & 0o020` 条件上。我实测过这个方向并**排除了**：第二轮特意把 live `Contents` 调成 775 再跑（条件为真，内层确实全变 775 = `chmod -R g+w` 跑过了），顶层仍然是 755。加上独立探针证明 `replaceItemAt` 总取 live 的 mode，按理该是 775。**所以真正的成因还没定位**，需要给 `InPlaceSwap` 加日志——已单开任务，不在本 PR 范围。审计文档里写的是「未解」，没有写成已知原因。

`make test` 在改完这三处注释后重跑全绿。

PR #750, opened 2026-09-18
Assert the rotation's group-write carry against a payload that lacks the bit

jizhi0v0:
The rotation test that guards `Contents`' group-write bit was vacuous: it built
**both** bundles at 775, so its assertion ran against a tree that already had the
bit. This makes the incoming side 755 — what a real vendor payload actually is —
so the assertion becomes load-bearing.

### The vacuity, measured

With `chmod -R g+w staged` deleted from `rotateContents` entirely:

| incoming fixture | result |
|---|---|
| 775 (old) | **passes** — the regression is invisible |
| 755 (this PR) | **fails** — `Contents/Resources lost the group-write bit` |

Under that mutation `Contents` *itself* still passes, which is correct and is the
point: the top level comes back group-writable because `replaceItemAt` carries the
replaced directory's mode, and only everything *below* it depends on the `chmod`.

### Why there is no production change

This started from a report that a real rotation left `Contents` at 755 while its
children were 775 — i.e. that `rotateContents` states an invariant it does not
hold. It does hold. Three instrumented rotations of the real
`/Library/Input Methods/QianwenIME.app` — mode, inode, owner, child modes and the
`chmod`'s exit status logged at every step, plus a 20ms external poller — all end
775 at every level, including across an input-method restart. One of them
reproduced the reported starting state exactly (live top 775, live children 755,
payload 755):

```
ROTPROBE entry: live=775 MacOS=755 Resources=755 | src=755 MacOS=755 Resources=755 | umask=22
ROTPROBE guard: read=775 willChmod=true
ROTPROBE chmod: succeeded=true | staged=775 MacOS=775 Resources=775
ROTPROBE post-replace: live=775 MacOS=775 Resources=775
```

`replaceItemAt`'s rule — top-level mode from the replaced directory, everything
below from the new item — also held 300/300 in isolation, with xattrs, and with the
replaced directory `root:staff` while unprivileged. That last case is the shipped
state for a **first** one-click of DoubaoIme and WeType: the `chown` back to root
fails and the mode is still carried (`Contents=775 owner=bobby`).

The reported 755 could not be reproduced from the state described, and that state
is no longer recoverable — the backup store keeps one slot per app and the snapshot
from that run was overwritten. What *is* reproducible is that nothing in the suite
would have caught it if it had been real. Hence: test only, `InPlaceSwap.swift`
byte-identical to `main`.

### Test

`swift test --filter "InPlaceSwap|ContentsRotation|SwapFailure|InterruptedSwap"` —
23 tests, 4 suites, green; plus the red→green mutation above in both directions.


jizhi0v0:
All three flagged claims re-run on `2dec570` itself (not on the pre-rebase tree), raw output below. No code change — 0 blocking findings, and nothing here contradicts the diff.

**1. `InPlaceSwapTests.swift:173` — "QianwenIME's download measured 755 at every level."**

From the instrumented rotations on 2026-09-18 (`src` is the payload, `live` is the installed `Contents`):

```
17:21:00.625 duo ROTPROBE entry: live=775 MacOS=775 Resources=775 | src=755 MacOS=755 Resources=755 | umask=22
17:25:50.618 duo ROTPROBE entry: live=775 MacOS=755 Resources=755 | src=755 MacOS=755 Resources=755 | umask=22
```

Both are vendor-download payloads: 755 at every level. (Two other runs from the same session show `src=775` — those were `duo backups restore`, whose payload comes from the backup store, already carrying the bit. Only the download is 755.)

**2. `InPlaceSwapTests.swift:198` — the top-level carry vs. the `chmod`-dependence.**

The mutation table, re-run at `2dec570`, `chmod -R g+w staged` deleted from `rotateContents`:

| fixture | `chmod` | result |
|---|---|---|
| incoming 755 (this PR) | live | `✔ passed` |
| incoming 755 (this PR) | **deleted** | `✘ Contents/Resources lost the group-write bit` |
| incoming 775 (old) | **deleted** | `✔ passed` |

Row 3 is the vacuity this PR removes. In row 2, `Contents` *itself* still passes — which is the split the comment describes: the top level comes from `replaceItemAt`, only the subtree depends on the `chmod`. Tree restored to `2dec570` afterwards (`git diff --quiet` clean).

**3. `.claude-review/pr.md` — 300/300 isolation.**

```
$ swift probe3.swift      # live 775/755, staged 775/775, backupItemName set
775/775: 300
```

One bucket, 300 runs, no flakiness. Also measured, and the one case that needs `sudo` to set up: with the replaced `Contents` `root:staff 775` and the process unprivileged, the result is `Contents=775 owner=bobby` — the `chown` back to root fails and the mode is still carried. That is the shipped state for a **first** one-click of DoubaoIme/WeType, so it is the case worth knowing; it is not asserted anywhere in the suite, because a test cannot chown to root.

PR #752, opened 2026-09-18
A backup written before the flag stripping must still be deletable

jizhi0v0:
`4259432` (v0.3.66) stopped `ditto` from freezing our own copies: `save` strips
`uchg` from what it stores, sweeps its staging dirs, and un-flags the generation
it is about to supersede. Nothing written since carries the flag.

The copies written *before* it still do, and the paths that delete a backup were
left as they were — a bare `try? removeItem`, which is refused with `EPERM` on
the one locked file and says nothing:

| path | what happens on a pre-0.3.66 backup |
|---|---|
| `remove(forKey:)` — the delete sheet, and the discard after a staged relaunch | deletes nothing, reports success |
| `pruneOrphans()` — "Clean Up Now" and the auto-prune | deletes nothing, and **counts the bundle as freed anyway** |
| the legacy-key cleanup in `save` | the migrated orphan stays forever |
| the rollback scratch in `restore` | wedges every later rollback of that app |

`save` is the only path that ever un-flagged such a copy, and it only runs while
the app is still installed and still being updated. For anything else — an app
you have stopped updating, one you uninstalled — the backup is permanent.

### The delete is a partial one, which is why it stayed invisible

`removeItem` unlinks its way down to the locked file before it is refused, and
`backup.json` can go with it. After that the store is in a state nothing owns:

* `backup(forKey:)` → nil, so the row loses its rollback affordance and looks deleted;
* `pruneOrphans` skips it — no sidecar to read the original path from;
* `totalSize` keeps counting it, so Settings shows disk the UI cannot reclaim;
* `listing()` keeps showing it, sidecar-less, as a row that reports success every
  time it is ticked and never goes away.

Measured on a fixture (`key/backup.json` + a bundle whose `Contents/advInfo.json`
is `uchg`), `FileManager.removeItem`:

```
removeItem: THREW -> NSCocoaErrorDomain Code=513 … NSPOSIXErrorDomain Code=1 "Operation not permitted"
survivors: key/Fake.app/Contents/advInfo.json, key/Fake.app/Contents/Info.plist
           (backup.json and Contents/MacOS/ already unlinked)
```

### The fix

`removeClearingImmutableFlags(at:)` tries the plain removal first and only on a
refusal clears `UF_IMMUTABLE`/`UF_APPEND` and retries. `lstat`/`lchflags`, never
the follow-the-link pair, so a symlink inside a bundle has its own flag cleared
and a link cannot reach outside the store. In-process rather than the `chflags`
child process `save` uses, because every caller is already on a Dispatch thread
doing the deletion.

Clearing is safe here because the target is always a copy this store made, never
the user's own file. `schg` needs root and still fails — now in the log instead
of in silence. A restore is unaffected: it copies its own bundle out, flags and
all.

`pruneOrphans` also stops counting what it did not delete. The number it returns
is shown as space reclaimed.

### What this does *not* change

The restore path, the save path, and the flag stripping that `4259432` added.
This is only the removal side.

### Verification

Four regression tests, each through the store's own public path rather than a
fixture delete, and each confirmed red before green by reverting the production
change:

| test | red without |
|---|---|
| `aPoisonedBackupIsDeletedWhenTheUserAsksForIt` | the flag clearing |
| `pruningAPoisonedOrphanReclaimsItAndCountsOnlyWhatWentAway` | the flag clearing |
| `aPoisonedRollbackScratchDoesNotWedgeTheNextRestore` | the flag clearing |
| `aPruneThatCannotDeleteReportsNothingFreed` | the accounting change |

The first one asserts the **directory** is gone, not that `backup(forKey:)` is
nil: the partial delete takes the sidecar, so the existing `removeDropsBackup`
assertion passes on a backup that is still entirely on disk.

The last one forces a removal that fails for a reason clearing flags cannot fix,
by making the store root read-only — `schg` would be the faithful reproduction
but setting one needs root. Without it the accounting change is untestable: with
the flag clearing in place every removal succeeds, and counting-before and
counting-on-success agree.

Supersedes #55, which fixed the same class of failure against the pre-`4259432`
`BackupStore` and has conflicted with `main` since.


jizhi0v0:
Round 1: 0 blocking, nothing to change. The three unverified claims are all
confirmed on this machine (uid 501, APFS, macOS 27.0).

**1. `BackupStore.swift:794-797` — the refusal is a partial delete that can take
`backup.json`.** Confirmed. Fixture `key/backup.json` + a bundle whose
`Contents/advInfo.json` is `uchg`, then `FileManager.removeItem(at: key)`:

```
removeItem: THREW -> NSCocoaErrorDomain Code=513 … NSPOSIXErrorDomain Code=1 "Operation not permitted"
survivors: key/Fake.app/Contents/advInfo.json
           key/Fake.app/Contents/Info.plist
gone:      key/backup.json, key/Fake.app/Contents/MacOS/
```

Which of the unlocked siblings survive is `removefile`'s traversal order and not
something the comment claims; that `backup.json` *can* go is the part that makes
the failure invisible, and it does.

**2. `BackupStore.swift:848-852` — the `lchflags` mask covers what
`chflags -R nouchg` cleared.** Confirmed, and it is a strict superset.
`nouchg` clears `UF_IMMUTABLE` alone:

```
set:      uappnd,uchg
after /usr/bin/chflags -R nouchg:   uappnd        ← still there
after lchflags(f, flags & ~(UF_IMMUTABLE|UF_APPEND)):  -
```

And on each flag alone, `lchflags` with that mask clears it: `0x2` → none,
`0x4` → none.

**3. `BackupStoreTests.swift:1013-1014` — a read-only store root refuses the
unlink.** Confirmed, including the ordering caveat you raise. `chmod 0555` on the
parent, then `removeItem(root/gone)`:

```
removeItem: THREW -> NSCocoaErrorDomain Code=513
left behind: root/gone   (its contents — backup.json and the bundle — were unlinked)
```

So the directory survives, which is what the test asserts, and the contents do
not — the under-reporting you filed as conservative-and-not-blocking is exactly
this case, now measured rather than assumed. Non-root is the other precondition:
`id -u` is 501 here and the CI runner is not root either, and the check passed
there on this commit.

PR #755, opened 2026-09-18
Give tests one scratch preferences suite that cleans up after itself

jizhi0v0:
The test targets open throwaway `UserDefaults(suiteName:)` domains so they never write the real `com.duoupdater.app` settings. Nothing removed them, so they accumulated in `~/Library/Preferences`: **10,173 files (39.7 MB)** on my machine, **517** on the mini runner. Nothing a released build does creates these — production code only ever opens `com.duoupdater.app` — so this is developer-machine hygiene, not a user-facing bug.

`scratchDefaults()` now registers each suite with a `.scratchPreferences` scope that empties the domain and unlinks its plist when the test ends, including when the test throws or records an issue. A suite that forgets the trait fails with a message naming the fix rather than leaking quietly.

## The obvious fix does not work

Measured 2026-09-18 on macOS 27.0 (Darwin 27.0.0):

| recipe | files left, per 50 suites |
| --- | --- |
| `removePersistentDomain` + `CFPreferencesAppSynchronize` + unlink | 39 / 50 |
| …the same, with a 50 ms settle before the unlink | 39 / 50 |
| `removePersistentDomain` + `CFPreferencesSynchronize` + unlink | 38 / 50 |
| `CFPreferencesAppSynchronize` + unlink | 0 / 50 |

- **`removePersistentDomain(forName:)` only empties the domain.** The file stays as a 42-byte empty plist, so the count still grows by one per test. `InventoryTestFlightReadTests` already had it in a `defer` and leaked anyway — a good chunk of those 10,173 are its empty shells.
- **Unlinking loses a race with cfprefsd**, which caches every domain the process touched and writes it back on its own schedule.
- **Even both together, plus an exit backstop**, left 0, 7, 14, 0 and 14 files over five runs. The write-back can land after the process is gone — one run's leftovers were observed appearing during the *next* run.

So this empties the domain first, which is what an in-process cleanup can honestly promise: a write-back can then only restore an empty domain, never test data. Collecting the stragglers needs a sweep outside the test process, which is a separate PR so this one stays test-only.

## Why there is no unit test for it

The write-back happens after the test process exits, so an in-process `fileExists` reports "gone" for every recipe in that table — including the broken ones. Such a test would pass while the leak continued. The check that discriminates is counting the plists across a full `make test`, and that is what I ran.

## Verification

`make test` on this branch: **exit 0**, `3364 tests in 291 suites passed`, `371 tests in 47 suites passed`, `App tests — 51 of 51`. The one known issue is the pre-existing `withKnownIssue` in `EventStoreSchemaDriftTests`, untouched here.

The missing-trait guard was mutation-tested — a suite deliberately without `.scratchPreferences` fails with:

```
✘ scratchDefaults() was called outside a .scratchPreferences scope, so
  com.duoupdater.tests.165D39DC-... would be left behind in ~/Library/Preferences.
  Add .scratchPreferences to the enclosing @Suite.
```

## Note on scope

Two things in the original brief turned out not to hold: `AppScannerFilterTests` is not a producer (its `com.duoupdater.tests.…` string is a `CFBundleIdentifier` for a temp bundle, never a defaults suite), and the 117 `com.duoupdater.harness.*` files come from a producer that no longer exists in any branch or worktree — dated only Aug 28 and Sep 1.


jizhi0v0:
Closing in favour of a better design, developed in parallel by another session on `claude/strange-shamir-fbd52b`. Same task, dispatched twice — my fault for not checking for a sibling session before starting.

Worth recording that the two sessions measured the same OS behaviour independently and agreed: `removePersistentDomain(forName:)` only empties the domain, the file survives, and cfprefsd writes the domain back after the process exits, so no ordering of remove/synchronize/unlink from inside the test process makes the file stay gone.

Where that PR is better: it stops fighting the write-back and bounds the *names* instead. One fixed label per suite rather than a UUID per test means one plist per label — two files in total, however often the suite runs — so it needs no sweep, no Makefile change, and nothing that deletes files under `$HOME`. It also handles the failure mode that fixed names introduce (a run that dies mid-test handing its values to the next run) by clearing on the way in, verified by killing a run mid-test.

Supplementary data from this branch, in case it is useful there:

| recipe | files left, per 50 suites |
| --- | --- |
| `removePersistentDomain` + `CFPreferencesAppSynchronize` + unlink | 39 / 50 |
| …with a 50 ms settle before the unlink | 39 / 50 |
| `removePersistentDomain` + `CFPreferencesSynchronize` + unlink | 38 / 50 |
| `CFPreferencesAppSynchronize` + unlink | 0 / 50 in isolation, 5 / 8 leaked in the real suite |

Also measured: the leak is 7 files per run from `VisibilityWriteTests`, and the write-back can land after the process is gone — one run's leftovers were observed appearing during the *next* run.

Two scope notes that may save time there: `AppScannerFilterTests` is not a producer (its `com.duoupdater.tests.…` string is a `CFBundleIdentifier` for a temp bundle, never a defaults suite), and the 117 `com.duoupdater.harness.*` files come from a producer that exists in no branch or worktree — dated only Aug 28 and Sep 1.

PR #756, opened 2026-09-18
Sweep the tests' scratch preferences domains in `make test`

jizhi0v0:
Companion to #755, and independent of it — either can land first, and each helps on its own.

The Swift test targets open throwaway `UserDefaults(suiteName:)` domains (`com.duoupdater.{tests,harness}.<UUID>`) so they never write the real `com.duoupdater.app` settings. macOS backs each with a real plist in `~/Library/Preferences`, and they had accumulated: **10,173 files (39.7 MB)** on my machine, **517** on the mini runner.

## Why this cannot live in the tests

cfprefsd keeps its own cached copy of every preferences domain a process touched and writes it back on its own schedule. Measured 2026-09-18 on macOS 27.0, that write-back lands *after* the test process has exited — in one case only during the next run. So an unlink from inside the test process cannot be the last word on the file, however carefully it is written. With test-side cleanup alone (#755), repeated runs still left 0, 7, 14, 0 and 14 files behind.

`make test` therefore sweeps at both ends: the leading sweep collects the previous run's late write-backs, the trailing one this run's. This PR's own verification run shows exactly that:

```
swept 7 scratch preference domain(s)     <- leading: previous run's stragglers
...
swept 7 scratch preference domain(s)     <- trailing: this run's
```

## Blast radius

Only exact `<prefix>.<UUID>.plist` names are matched, via `find -maxdepth 1`, so nothing that is not a generated scratch domain can be hit. Verified on my machine: 10,173 removed, `~/Library/Preferences` 53M → 13M, and the real `com.duoupdater.app.plist` untouched at 29,845 bytes.

This runs on any machine that runs the test suite — my machine, the mini self-hosted runner, and CI runners (ephemeral, so nothing to collect there). It is not reachable from a released build: production code only ever opens `com.duoupdater.app`, so no user's machine ever grows these files or has anything swept.

Worth a second opinion on: this is a test step that deletes files under `$HOME`, which is unusual for this repo even though the pattern is tightly scoped.

## Verification

`make test` on this branch, with no test-side change present: **exit 0**, `3364 tests in 291 suites passed`, `371 tests in 47 suites passed`, `App tests — 51 of 51`, zero failures. Scratch plist count after the run: **0**. The one known issue is the pre-existing `withKnownIssue` in `EventStoreSchemaDriftTests`.


jizhi0v0:
Closing in favour of a better design, developed in parallel by another session on `claude/strange-shamir-fbd52b`.

This PR existed because I treated the problem as "how do I delete the plist", which is a race against cfprefsd that cannot be won from inside the test process — hence a build step reaching into `$HOME`, which I flagged above as wanting a second opinion.

The other approach reframes it: the lever is not how to delete the file but **how many distinct domain names the suite ever asks for**. With one fixed label per suite instead of a UUID per test, the directory holds one file per label — two files in total, for good — so there is nothing to sweep and no build step touching `$HOME`.

The one-time cleanup this PR was also meant to cover is already done: 10,173 files removed, `~/Library/Preferences` 53M → 13M, the real `com.duoupdater.app.plist` untouched.

The measurements here still stand and are worth keeping for the record:

- `removePersistentDomain(forName:)` empties the domain but leaves a 42-byte plist, so the count grows by one per test regardless.
- Unlinking races cfprefsd's write-back: the file was back in 38-39 of 50 suites per recipe.
- Test-side cleanup plus an exit backstop still left 0, 7, 14, 0 and 14 files over five runs.
- This branch's own run showed the shape exactly: `swept 7` at the start (the previous run's late write-backs) and `swept 7` at the end (this run's, with no test-side cleanup present).

PR #758, opened 2026-09-18
Notes filed under a version the page had not published yet must not be permanent

jizhi0v0:
CleanShot X shipped 5.0.1 on 2026-09-18 and the workbench has been showing 5.0's notes under it ever since — the second time this app has done that from one appcast (5.0, 2026-09-01, was the first).

## What was wrong

Not the recipe: `duo verify --changelog --only cleanshot` is green and the production parser reads 5.0.1 off the live page as the first entry.

| | |
|---|---|
| our only fetch of `cleanshot.com/changelog` in seven days | `2026-09-18T11:52:42Z` |
| that page's `last-modified` | `2026-09-18T12:01:56Z` — 9 minutes later |
| what was stored | `…/changelogs/pl.maketheweb.cleanshotx__default__5.0.1.json`, newest entry `5.0` |

The appcast was offering 5.0.1 nine minutes before the changelog page mentioned it. The parse came back topping out at 5.0 and was filed as 5.0.1's notes, and `ChangelogDiskCache` treats an entry as immutable because a released version's notes never change. That invariant is about a page that has *published* the version; here the key can never change again, so no later fetch is ever made for it and the pane shows the new version's number over the previous version's list — nothing logged, nothing to retry.

The 5.0 occurrence was written off in the audit as self-healing on the next fetch, which by this design never comes.

## What changed

`Changelog.carries(version:)` asks whether a parse says anything about the version it is being filed under, and the two layers that were assuming it now check:

- **`ChangelogDiskCache`** still stores and paints an entry whose page fell short — it is the best notes anyone has, and it is what the vendor's page was showing — but as *provisional*, for `provisionalWindow` (6 h) rather than forever. Past that it is a miss, the next pre-warm or open re-reads, and a page that does carry the version is immutable from then on. A provisional entry never enters the in-memory mirror either: the mirror has no age, and this app is left running for days.
- **`AppListModel`**'s once-per-session revalidation debt is discharged only by a fetch that came back with the version's notes on it. Before, the fetch that poisoned the entry also marked it confirmed.

## Reviewer notes

The predicate is permissive by construction — true unless the page can be *shown* to be behind. Versions that aren't version-shaped are not judged (Figma and Notion title their entries, Cursor dates them, Xcode's is `27.2 beta 1 (27B5019j)`), and neither is a page at or ahead of the key. It is judged against the newest version-shaped entry anywhere on the page, not the first: a page carrying two trains is newest-first within a train, not across them.

It does **not** separate a vendor who publishes late from one who numbers notes more coarsely than builds (Raycast's page says `2.4` where the build is `2.4.1.0`; JetBrains Toolbox `3.8.1` for `3.8.1.88030`). Both read as behind, both cost one re-read per window and nothing else — which is why this must never be surfaced as "the notes are stale". About a quarter of the entries in the cache the window was sized against read as behind; at four windows a day that is roughly a doubling of the app's daily changelog traffic, each read being the one this layer exists to skip rather than an extra request on top.

## Verification

- Full `make test` green (including the app target, claim-lint and the audit checks).
- 10 new tests. Every branch checked by mutation: six reverts (window ignored, provisional mirrored, never-behind, no-version-numbers-counts-as-behind, unreadable-key-counts-as-behind, judge-only-the-first-entry) each failed the test that covers it.
- The real poisoned entry from a machine that hit this was replayed through the production cache: served while fresh, a miss once backdated past the window, and `ChangelogService.loadUncached` against the live page now returns `5.0.1 | "Minor bug fixes and improvements"` first.

The audit gains the 2026-09-18 timeline and corrects the "it will self-heal" conclusion from the 5.0 entry.


jizhi0v0:
### Round 1 verdict

**Blocking finding 1 — accepted, and it was my own doc that said so.** The comment promised the entry is "stored and painted rather than thrown away" and the code three lines down returned `nil` for it. Offline (or on any failed read), that is strictly worse than no window: `prewarmChangelogs` saw `nil`, fetched, failed, and settled `.failed` → web-view fallback, for ~a quarter of entries, most of which were never stale.

Fixed by making the window a **request for a re-read, never a reason to withhold**:

- `ChangelogDiskCache.hit(for:)` returns `Hit { changelog, needsReread }` and never drops an aged provisional entry; `get(for:)` is now `hit(for:)?.changelog`.
- `prewarmChangelogs` paints the disk copy and fetches **in addition** when `needsReread`, keeping the painted copy if the fetch comes back nil. The open path already had this shape.
- Docs on the type, `provisionalWindow` and the test now state the failed-re-read case explicitly.

New mutation checks, each failing the test that covers it: `needsReread` never set (window inert); the aged entry withheld again (this exact finding); a provisional entry mirrored in memory.

### Unverified external claims

Re-derived locally; one was wrong and is gone.

- **CleanShot 9 minutes (5.0.1)** — confirmed today: `duo events --host cleanshot.com` has exactly one request in seven days, `2026-09-18T11:52:42Z`; the page's `last-modified` is `12:01:56Z`. **6 minutes (5.0)** is transcribed from the audit's own record (`22:40:53` fetch vs `22:46:36`), and the comment now says where both timings live.
- **"about a quarter" / 16 of 61** — measured by running the production predicate over this machine's cache files with the key version recovered from each file name.
- **Raycast `2.4` / `2.4.1.0`, Toolbox `3.8.1` / `3.8.1.88030`** — read off the cached entries against the version each is filed under; the comment now says that is where they come from.
- **"Figma and Notion title their entries"** — could not confirm, and the cached Notion entry disagrees: its versions are `7.34.0, 7.33.0, …`. I had carried that example over from `Verify.swift`'s comment without checking it. Replaced with two I did check: Cursor dates its entries (`version` = `Sep 10, 2026`, prose in `title`) and Xcode's key is `27.2 beta 1 (27B5019j)`.
- **Audit timings / "recipe 没坏"** — the request-ledger and `last-modified` figures above, plus `duo verify --changelog --only cleanshot` green and `ChangelogService.loadUncached` against the live page returning `5.0.1 | "Minor bug fixes and improvements"` first.

Also observed in the wild after installing this build: the next pre-warm re-read exactly the provisional entries (8 apps, one burst), and Notion's had caught up — `7.32.0` → `7.34.0`, which is the freeze this PR is about, healing.

jizhi0v0:
### Round 2 verdict

Both findings accepted and fixed in `29983428`, and the round-2 comment is right that two round-1 edits never reached the tree — including one I reported as done. They shared a script with a third replacement whose assertion failed, and the file write came after all three, so the whole file was left untouched. My round-1 reply asserted the result instead of re-reading the diff; that was the mistake, and the review caught it.

**Finding 1 — Notion.** Fixed. Checked at source rather than taken on the comment's word: `NotionChangelogRecipeTests` records that the title-as-version page recipe was replaced by `.notionPageChunk` on 2026-08-22, and this machine's cached `notion.id` entry numbers its versions `7.34.0, 7.33.0, …`. Notion is one of the apps that DOES go provisional, so the example pointed a reader away from the exact case it was there to explain. The sentence now uses Figma (feed of product announcements with no per-entry app version — `com-figma-Desktop.swift:74`-`80`), Cursor (`version` is `Sep 10, 2026`, prose in `title`) and Xcode (`27.2 beta 1 (27B5019j)`), and says explicitly that Notion is not one of them.

**Finding 2 — `prewarmChangelogs`'s doc.** Fixed. It now says a disk hit that carries its own version needs no network, names the second class that does go to the wire (`needsReread`), points at `provisionalWindow` for the sizing, and states that a re-read is painted on top of the disk copy rather than instead of it.

**CleanShot timings provenance.** Now in the comment: request time against the page's own `last-modified`, both written down in `docs/app-audits/pl-maketheweb-cleanshotx.md`.

### A third defect, found by running it rather than reading it

The reported behaviour ("opening the app re-reads") did not happen on the real machine, and the reason is a bug neither review round could see from this diff: **the revalidation debt had no collector.** `changelogRevalidated`'s own doc says `ensureChangelogLoading` runs "every time the user opened the app"; the pane asked for it only from its `.loading` branch, and a key the pre-warm already painted arrives `.loaded` and never reaches that branch. Recorded, never collected — which is why 5.0.1 repeated 5.0's failure three weeks later.

Measured: on the build before this commit, one pane open produced **zero** requests to `cleanshot.com`. With the `.loaded` branch asking too:

```
2026-09-18T17:06:44Z  changelog  200  184247 bytes     ← the pane open
fetchedAt   : 2026-09-18T17:06:44Z
newest entry: 5.0.1 | 18 September, 2026 | ["Minor bug fixes and improvements"]
```

The entry now carries its own key version, so it is immutable from here and asks for nothing further.

### Unverified external claims

The post-install burst the round-2 comment could not observe: `duo events --purpose changelog --client app` shows 12 requests in one batch at `16:42:35`–`16:42:37Z`, immediately after installing the first build, for the hosts backing exactly the provisional entries; `notion.id`'s cached entry went from newest `7.32.0` to `7.34.0` in that batch. The 6-minute and 9-minute CleanShot timings and the 16-of-61 ratio are as described above and in the audit.

Full `make test` green on this commit.

jizhi0v0:
### Round 3 verdict

No blocking findings, and both round-2 fixes confirmed. Took the two observations it made in passing, in `e8156248`:

- **The stale attribution in my own test file.** `ChangelogCarriesVersionTests` still said "Figma and Notion title their entries" — the wording the last commit corrected everywhere else. The assertion never depended on it, which is exactly how a wrong example survives a green test. Now names Figma and Cursor, and says Notion is not one of them and why.
- **"On failure it leaves the key absent so the open path retries."** Pre-existing, as round 2 said, but it sits in the paragraph I rewrote, so it now reads as this PR's claim — and it contradicts the `.failed` settle twelve lines below and that code's own explanation. Corrected to what the code does.

On the duplicate-task race the review checked and dropped (two overlapping appearances during one in-flight re-read): agreed it is not a wrong result — same key, same version, same content — so I am leaving it rather than widening this PR. Worth its own look if anyone sees duplicate changelog requests in `duo events`.

Unverified external claims: unchanged from round 2's answer. The one added this round — "one pane open, zero requests; after the fix, the open fetches" — is `duo events --host cleanshot.com`, one request in the seven days before (`11:52:42Z`, the one that poisoned the entry) and the next at `17:06:44Z`, the moment the pane was opened on the fixed build, after which the entry reads `5.0.1 | 18 September, 2026`.

Full `make test` green on `e8156248`.

jizhi0v0:
### Round 4 verdict

0 blocking, second round running, and nothing in the recheck contradicts the tree. **Not pushing a fifth commit.**

The one observation it made in passing is real and I am deliberately leaving it: the `prewarmChangelogs` paragraph is attached by Swift to `private static let prewarmNetworkGate`, not to the function it describes — the gate's own two-line doc sits between them. That ordering is at the base commit too, so it is not this PR's doing; moving the block is a pure-comment relocation that would cost another commit and another review round for nothing a reader is misled by (the paragraph names `prewarmChangelogs` in its first line). Worth folding into whatever next touches that function.

Unverified external claims carried forward: unchanged, and answered in the round-2 and round-3 replies.

PR #763, opened 2026-09-19
Keep the rollback backups on another disk

jizhi0v0:
Rollback backups are the reason an in-place update feels safe, and they are also
the reason the boot volume fills up: retention is one copy per app, but a copy of
an app is the size of an app. This lets them live on a disk instead.

Supersedes #57, which is still open against `fix/backup-locked-file-undeletable`
and holds only the first four commits of this branch. Everything after
`182bd851` is work that has never been on a pull request.

## What it does

**The store can be on another disk.** A backup is written to the boot volume
first, always, and a transfer drains it to the disk in the background. That is
what makes an unplugged disk a delay rather than a missing rollback point. On the
disk a backup is one Apple Archive (`.aar`) rather than a bundle, which is what
lets a volume formatted for Windows, or a share, hold a macOS app at all.

**Disks are picked from a list, not a file panel.** Every attached volume is
shown with what it can do: free space read from the volume rather than from a
stale enclosing directory, a Time Machine disk named as spoken for instead of
failing to write to it later, and this Mac offered as one of the choices. A disk
that arrives or leaves is noticed from the mount table.

**Backups already taken can be moved, or deliberately kept here.** Choosing a
disk asks about the copies already on this Mac; a backup told to stay says so.

**Reading is wider than writing.** Backups are read from every connected disk
and written to one, because a disk that is merely plugged in may hold another
Mac's backups — and an automatic prune across it would correctly identify every
one of them as an orphan and delete them all.

**`duo backups` gained `disks`, `sync`, `verify` and `probe`**, and names which
disk a backup is on.

**A backup on a disk can be compared with the installed app.** The reader took
`.app`, `.zip`, `.dmg` and `.pkg` and refused everything else, so the comparison
worked only while the backups happened to be on this Mac.

**…and comparing it does not unpack it.** The transfer records what the backup
holds as it goes — it has just read the whole bundle to compress it — and the
comparison reads that instead. Measured on a real backup on a USB stick, release
build, Cursor 3.20.21 (a 280 MB archive holding 930.2 MB and 17,246 files):

| | without the record | with it |
|---|---|---|
| the old side — the same walk either way | 12.29s | **0.06s** |
| deleting the unpacked copy afterwards | 1.30s | **0** |
| the whole comparison, warm | 8.77s | 2.81s |
| the whole comparison, nothing cached | 14.06s | 6.38s |

The first two rows are the honest measure: same work, same conditions, with and
without. The totals move with how warm the *installed* app is, because with the
old side gone that is the entire wall clock, and it cannot be recorded — it is
read fresh every time.

The record is kept under Application Support, not in the backup store: a transfer
rebuilds the key's directory on the disk from the sidecar alone and deletes the
outbox copy it came from, so anything kept in either place would be destroyed by
the very operation that writes it. It is named by the fingerprint of the copy it
describes and by the format version that wrote it, so a stale or foreign entry is
a missing file rather than a decode to be trusted. It is a cache and never a
gate: every miss reads the backup exactly as before, and `BackupManifest` is
untouched, so the restore gate still compares whole structs against every
existing sidecar.

What it does not keep is the strings index — 17–65 MB against 54 KB–3.7 MB for
everything else. That costs the comparison a real answer, and saying so is the
point of the fourth `KeyEvidence` case: without it every key found in the new
binaries would print "in new binaries only", which is a claim *about the old
binaries* made from a search that could not run.

## Cost

A transfer takes 1.6–2.6x as long for the record (Claude 1.99 → 3.25s, Excel
8.64 → 18.55s). Nothing is added to the update path — `save` does not record, and
a test asserts it does not, so moving that decision is something someone has to
do rather than something that drifts. A backup that stays on this Mac gets no
record and loses almost nothing by it: both sides of a comparison are read
concurrently, so skipping one of two equal walks moved the wall clock by 4–10%.

## Verified

`make test` passes on this branch after merging current `main` — 3505 + 378
cases, exit 0.

The record path was exercised end to end on this machine by a real update: four
apps (Conductor, Cursor, VS Code, ChatGPT) went through
`InstallCoordinator.backUp` → `save` → the transfer queue → `transferToDestination`,
and for all four the manifest digest in the sidecar on the disk matches the
entry's filename byte for byte. The comparison was then run from the app's own
workbench and read the record.

Fourteen mutations of the new guards were applied and all fourteen were caught by
the tests — including deleting the lookup, searching an unindexed old side
anyway, dropping the format version from the entry's name, keeping the strings
blob, recording in `save` instead, and not forgetting a record when its backup is
deleted.

### DIFF
--- a/App/Resources/Localizable.xcstrings
+++ b/App/Resources/Localizable.xcstrings
@@ -3117,6 +3117,47 @@
         }
       }
     },
+    "A different disk is mounted here": {
+      "extractionState": "manual",
+      "localizations": {
+        "de": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Hier ist eine andere Festplatte eingebunden"
+          }
+        },
+        "es": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Aquí hay montado otro disco"
+          }
+        },
+        "fr": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Un autre disque est monté ici"
+          }
+        },
+        "ja": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "別のディスクがマウントされています"
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Здесь подключён другой диск"
+          }
+        },
+        "zh-Hans": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "此处挂载的是另一块磁盘"
+          }
+        }
+      }
+    },
     "A few quick steps let Duo Updater keep your apps updated without interrupting you. You set these once — they persist across launches.": {
       "extractionState": "manual",
       "localizations": {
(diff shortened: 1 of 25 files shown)


jizhi0v0:
All three blocking findings were real; confirmed by reading the code, fixed in `4e23e5dc`.

**1 — the Compression setting.** Confirmed: `App/Sources/Preferences.swift` writes `backupCompression`, and nothing outside the CLI ever read it into a transfer. The store is now told it the same way it is told the destination (on change, and at launch, because `didSet` does not fire inside `init`), and `transferToDestination` resolves its default per call rather than capturing one — the setting can change between one transfer and the next.

**2 — the shared `inFlight` slot.** Confirmed, including the second half the finding names: the `defer` cleared whatever was in the slot, not this run's entry, so the two runs could also clear each other. Replaced by `RunningChildren`, which hands each run a token it gives back.

**3 — the assertion against a string no report emits.** Mine, from shortening the phase label to fit the 28 characters `timingSection` pads to. Fixed to `"read the recorded facts"`.

**On the test for finding 1, which I got wrong first.** It asserted that the two archives differed byte for byte — and it *passed under the mutation it existed to catch*, because two `ditto` copies carry different mtimes and the archives differ whatever the algorithm. It now reads the algorithm off the archive's own container magic (`pbze` for lzfse, `pbzx` for lzma, measured with `aa archive -a` on macOS 27; `aa` is byte-deterministic for identical input). Reporting it here because it is the same defect shape as finding 3, one layer up.

**A gap I am not papering over.** The one line in `Preferences` that hands the setting to the store is untested. `Preferences.init` reads the Keychain, no App test constructs it, and doing so in a hostless test bundle risks a password prompt in `make test` and on the self-hosted runner. Everything below that line is tested; that line is not. Say the word if you want it covered and I will look for a seam that does not construct `Preferences`.

**Verification.** `make test` exit 0 — 3508 + 378 cases, plus 51 App cases. Each fix fails under a single-line mutation of itself, run on a clean tree with git restoring between rounds:

```
caught  F1 transfer ignores the configured value
caught  F2 add() keeps a single slot
caught  F2 remove() drops every entry
caught  F2 terminate cancels only the newest
caught  F3 a wrong fingerprint is served anyway
```

The `RunningChildren` tests deliberately never await a task: under a mutation that breaks cancelling, awaiting hangs the suite instead of failing it — which is how the first version behaved, and why this note exists.

**On the unverified-claims list:** those are pre-existing measurements in the #57 stack, not claims this branch introduces, with two exceptions I can speak to. The `stringsBlob` figures at `BundleFacts.swift:55-57` are mine and were measured on Claude/ChatGPT/Excel/DuoUpdater in a release build, reproduced within 2% by an independent re-implementation of the serializer. The timings in the PR body are from this machine against a real backup on a USB stick, and the `make test` counts are the ones above.

jizhi0v0:
CI went red twice on `4e23e5dc`. Neither failure is reachable from this branch's product code; both are tests asking their machine for something a hosted runner could not give. Pushed `b9b5a9ff`.

**Attempt 1 — `StreamedLinesTests.theCapReleasesAWaitThatNeverEnds`.** The test caps a wait at 0.05 s and wraps it in a 30 s watchdog. Both arm on `DispatchQueue.global()`, so the 30 s timer can only win if that queue is starved for thirty seconds — which is a statement about the machine, not about `StreamedLines`. It ran beside `roundTripPreservesTheCodeSignature` (57 s) and `vendorDownloadPassesSignatureGate` (64 s). The file already documents these bounds as "not performance assertions, so they are generous"; 30 s was not generous enough to be the watchdog it says it is. Raised to 120 s. A wait that never returns still fails.

**Attempt 2 — same failure, plus `BackupDestinationProbeTests.theFileSizeCeilingIsReadFromTheFilesystem` on APFS.** Its four formats are parameterized, so they each create, attach and mount a 256 MB sparse image *at the same time*; `hdiutil` refused the APFS one while MS-DOS, exFAT and HFS+ all succeeded. Serialized across formats. Locally all four create fine one at a time, and all five spellings do at 40 MB, so this is contention rather than a format that stopped working.

Neither test is skipped, disabled or quarantined, and no assertion is weakened.

**Why I do not think either is this branch's.** `StreamedLines` is not touched by the diff at all, and `test` passed on this same branch's previous head (`832867f`) an hour earlier with everything in this PR except three review fixes — none of which go near either test. The probe failure is a `#require` on `hdiutil` succeeding, not an assertion about our code. Against that: the suite has grown over this stack, and both tests were sitting near their margins, so the branch is a fair proximate cause even though the code is not.

**Verified:** `make test` exit 0 — 3508 + 378 cases and 51 App cases.

**One thing I deliberately did not "fix".** A local App-test run reported `aMalformedTeamYieldsNoRequirement` as never executed. It had executed and passed — its ✔ reached the log as octal `\342\234\224`, so the harness's own "every declared case ran" grep missed it. It does not reproduce, so I left it rather than guess at it. Noting it because it is the kind of thing that comes back.

I have spent the one re-run these rules allow, on attempt 2. If `test` is red again on `b9b5a9ff` in one of these two places, that is new information and I will stop guessing and root-cause it rather than widen the margins further.

PR #764, opened 2026-09-19
Remember what a backup measures, instead of walking it again on every look

jizhi0v0:
Pressing **Clean Up…** walked every file of every stored backup before the first row appeared, and did it again on the next press. The sheet shows a size on every row, and a size is a recursive walk of a whole app bundle — everything else the listing reads is a handful of small files per backup. That is the one part whose cost grows with the store, which is why it got slower the more backups there were.

A stored backup does not change while it sits there: retention per key is one, `save` builds the new copy in hidden staging and swaps the whole key directory into place, and a transfer rebuilds it. So a measurement stays true until the key directory is replaced — which its inode and modification date report in one `stat`. `BackupSizeIndex` records each measurement against that identity, beside the other state this app keeps rather than on the disk it describes, and retires an entry when the directory it describes has been replaced or deleted.

The settings page and the sheet read the same index, so the page's own measurement — which it already takes when it opens — pays for the sheet. Whatever is left to measure is measured on several threads, which is most of what remains to pay on the first look after an update.

### Measured

A store of 101 backups, 240,308 files and 48 GB on a USB HFS+ disk, filesystem cache warm.

| | before | after |
|---|---|---|
| `listing()`, three consecutive calls (debug) | 1.90 / 1.92 / 1.90 s | 1.32 s, then 0.015 s |
| `storeSizes()`, same three calls | 1.80 / 1.84 / 1.83 s | 0.004 s |
| `duo backups disks` (release, fresh process) | — | 4.46 s with no index file, 0.022 s with one |

### Tests

`BackupSizeIndexTests` covers reuse within and across processes, re-measuring after a supersede (including one inside the same second, which only the inode catches), a sidecar rewritten in place, pruning a deleted backup, not pruning another store's entries, and a directory that cannot be stat'd. `BackupStoreTests.listingReportsTheNewSizeAfterABackupIsSuperseded` pins the wiring end to end.

Each guard was mutation-tested: inode-only, date-only, no prune, prune-everything, never-record, record-without-identity, and reading the identity back through `URL.resourceValues` (which caches on the `URL` value and goes stale) were each killed by the test that exists for them.


jizhi0v0:
Round 1 adjudicated — 0 blocking, and the one non-blocking note is fixed as a claim, not as code.

**The prune gap is real and the comment was wrong about it.** `sizes(of:)` returns before the prune on an empty input, so emptying a store strands that store's last entries. I left the behaviour: closing it means passing the root separately and then deciding, for a store that reads as empty, whether it is *empty* or *unreadable* — and getting that wrong retires a whole disk's entries every time it is unplugged at the wrong moment, which is a worse failure than a few dozen stale bytes that cannot be read back as a size. The comment now says so instead of claiming the prune is "the only thing that keeps the file from growing forever".

**The unverified claims, each checked on this machine:**

- The timings and the `URL.resourceValues` staleness are the measurements the diff was built from; the probe for the second is the one quoted (`freshMtime=…728148 / cachedMtime=…7242494` after a write through the same `URL`).
- `FileManager`'s shared instance: Apple's own documentation says its methods can be called from multiple threads safely, with delegate-based move/copy/remove/link operations as the stated exception — these use no delegate. https://developer.apple.com/documentation/Foundation/FileManager
- **HFS+ whole-second dates — this one I had only asserted, so I measured it.** Three directories created within one second on an HFS+ volume: `mtime=1789817168.000000000` for all three; the same three on APFS: `.931239901`, `.933524619`, `.936083046`. The comment now carries that measurement instead of the assertion, and the test's own comment no longer implies it runs on HFS+ (it runs on whatever `TMPDIR` is, which is why it stamps both copies itself).
- `replaceItemAt` leaving the new inode at the original path: measured in the same probe (`statInode=255957288` → `255957291`, with the modification date moving too), which is what `aSupersededBackupIsMeasuredAgain` then pins.

jizhi0v0:
Round 2 adjudicated — 0 blocking. The one carry-over it found is fixed: the inline comment at the prune no longer repeats "the only thing keeping this file bounded" and points at the doc block that states the exception instead.

PR #768, opened 2026-09-19
Relaunch: hold the backup transfer until the installer has had its window

jizhi0v0:
## 这个 PR 做什么

Relaunch 前的回滚备份把「移到备份盘」留在一个没人等待的 detached task 里，于是它会一直跑到我们 quit 掉 app 之后——也就是厂商自己的安装器唯一那几秒窗口里（macOS 27 下，宿主一退出，它用 `SMJobSubmit` 提交的 launchd job 就被卸载，`ExitTimeOut` 默认 5 秒后 SIGKILL；机制见 https://github.com/Squirrel/Squirrel.Mac/issues/336 ）。

**这个重叠是我们自己造成的，而且每次点 Relaunch 都会发生**——`StagedRelaunchBackup.shouldTake` 只看磁盘版本有没有前进，所以失败之后再点，会重新备份、重新拖一条尾巴。把它移出那个窗口是对的，代价也几乎为零。

## ⚠️ 但它不是失败的原因——这一点我最初判断错了

PR 刚开时我写的是「失败那次重叠、成功那次不重叠」，后续取证把这个因果推翻了：

| 时刻 | 我们的 relaunch | transfer 是否跨窗口 | 结果 |
|---|---|---|---|
| 20:49 | 是 | 是（BackupFacts sidecar birthtime **20:50:10**，ShipIt 20:50:01 已被杀） | 失败 |
| 21:11 | 是 | 是 | 失败（当时挂着 `sample` 探针） |
| 21:34 | **否**（用户自己 Cmd+Q） | **无**（根本没有我们的备份） | **失败** |
| 23:19 | 是 | **是**（transfer 约 23:19:46 → **23:20:01**，横跨整个安装） | **成功 3.08 s** |

我之前把 `backup.json` 的 `savedAt`（23:19:44，save 阶段结束）当成了「备份全部结束」，其实 transfer 在那之后才开始。**成功那次同样重叠。**

而且实测了争用代价——用复刻 `SQRLInstaller` 同样 API 的 harness，在我们的 868 MB→366 MB USB transfer 跑与不跑两种情况下各测一组：

```
idle                     中位 2.194 s
我们的 transfer 在跑      中位 2.278 s     → +0.084 s，占 5 秒预算的 3.8%
```

还有一条否定结论：**备份路径上没有冷/热之分**。`BackupFactsLibrary`（给 bundle-diff 视图）和 `BackupSizeIndex`（给 Clean Up 表）都不在备份路径上；`BackupManifest.compute` 每次都对全部 868 MB 重算 SHA-256。首次与后续实测无差别（save ≈2.2 s、transfer ≈12–16 s，五轮一致）。

**所以：20:49 那次 ShipIt 为什么用了 >7 秒还没走到 `Moving bundle`，仍然没有解释。** 不是这个 transfer（实测 +0.08 s），也不是探针（那轮没挂）。

## 改动

- `backupCurrent` 增加 `transferNow:`；staged-relaunch 路径传 `false`
- 放行点受三个显式条件约束（见 `startOwedBackupTransferIfIdle`）：批量走行期间（`batchRelaunchInFlight`）、还有 relaunch 在飞、还有 quit hand-off armed；让位退出与无法 arm hand-off 的 `.wontQuit` 由 `installerWindowPending` 直接跳过；hand-off 那条由 `relayQuitHandoff` 的 `defer` 放行
- transfer 的两行日志 `.info` → `.notice`：`.info` 在第三方 subsystem 下不落盘，正因如此这三次尝试在统一日志里**一条备份记录都没有**，事后只能靠 sidecar 的文件时间反推

普通安装路径不变，仍然装完立刻 transfer。

## 验证

- `BUILD SUCCEEDED`（Release 全量）
- `3517 tests in 302 suites` + `378 tests in 48 suites` 通过
- ⚠️ **真机行为未验证**：本机 `make install` 因 `errSecInternalComponent`（login keychain 锁定）失败，没能在装好的 app 上走一遍确认顺序。下次真实 relaunch 时看 `.notice` 日志里 `backup: moved` 是否落在 `applied=` 之后即可确认——这正是本 PR 新加的可观测性。


jizhi0v0:
Both blocking findings addressed in `3a6967a8`.

**1. `.wontQuit` / stand-down** — the defer no longer asserts the window has passed; it checks. `startOwedBackupTransfer` is now only reachable through `startOwedBackupTransferIfIdle`, which requires `relaunching` and `quitHandoffs` to both be empty, so an armed hand-off keeps the transfer waiting. `relayQuitHandoff` starts it once that hand-off settles — the first moment on that path where the window really is closed. The stand-down return (`takeRollbackPointBeforeStagedRelaunch` → nil) skips the start outright via a new `updaterTookOver` flag, since that is precisely the case where the app's own updater is starting as we leave.

**2. Update All** — same guard covers it: while the loop is inside the next row's relaunch, `relaunching` is non-empty, so app A's transfer cannot start inside app B's window. The last row to finish is the one that starts it. This is the right shape given `drain()` is global rather than scoped to one app, which the review correctly pointed out.

On the unverified-claims section: all four are accurate as flagged. The macOS 27 / `ExitTimeOut` behaviour and the 2026-09-19 field measurement are not derivable from this repository — the mechanism is written up with the evidence in https://github.com/Squirrel/Squirrel.Mac/issues/336 (including a correction posted there about which of those measurements were contaminated by the probe). Real-device verification of this PR is still outstanding: `make install` fails on this machine with `errSecInternalComponent` (login keychain locked), so the end-to-end ordering still rests on the code reading.

`BUILD SUCCEEDED`; `3517 tests in 302 suites` and `378 tests in 48 suites` pass.

jizhi0v0:
Both round-2 findings addressed in `86526186`.

**1. `relayQuitHandoff`'s release point could never fire** — correct, and the reason is exactly as stated: the call sat at the end of the body, while that id is still in `relaunching` (inserted at the top), so `startOwedBackupTransferIfIdle` could never pass; the six early returns skipped it outright. Moved into that function's own `defer`, after `relaunching.remove(id)`, which also covers every early return.

**2. `relaunching.isEmpty` does not cover Update All** — also correct, and this was the more embarrassing one: the loop awaits one row at a time, so a finished row's `defer` runs in the gap where the set is empty and the next row has not been inserted. The check passed every time. Replaced with an explicit `batchRelaunchInFlight`, set for the whole walk and released once after it, and the doc comment now says why `relaunching` cannot stand in for it.

Also handled the sub-case the round-2 table flagged inside finding 1: the **un-armed** `.wontQuit` branch (staged version unreadable, so no hand-off is armed) left `quitHandoffs` empty while the app was still up with its prompt. Nothing would ever signal that quit, so there is no later moment we can call safe — that branch now sets `installerWindowPending` and leaves what is owed for the next relaunch or install to drain. `updaterTookOver` was renamed to `installerWindowPending` to match what it now covers.

---

**Separately: I have corrected the PR description.** Its original causal claim ("failure overlapped, success did not") does not survive further forensics, and reviewers should not weigh this PR on it:

- The successful 23:19 run **also** overlapped — I had read `backup.json`'s `savedAt` (end of *save*) as the end of the whole backup; the transfer runs after it, and the destination sidecar's mtime puts it at ~23:19:46 → 23:20:01, across the entire install.
- The 21:34 failure had **no** DuoUpdater relaunch at all (user quit the app directly), so our backup is not necessary for a failure either.
- Measured contention, replaying `SQRLInstaller`'s own API sequence with and without our 868 MB→366 MB USB transfer in flight: median 2.194 s → 2.278 s, i.e. **+3.8 % of a ~5 s budget**.
- There is also no cold/warm split to appeal to: `BackupFactsLibrary` and `BackupSizeIndex` are not on the backup path, and `BackupManifest.compute` re-hashes all 868 MB every time (five runs, first indistinguishable from the rest).

So this PR removes an overlap that is genuinely ours and genuinely unnecessary, but **it is not the fix** for the observed failures, and why 20:49's ShipIt needed >7 s remains unexplained.

`BUILD SUCCEEDED`; `3517 tests in 302 suites` and `378 tests in 48 suites` pass. Real-device verification still outstanding (keychain locked, `make install` fails with `errSecInternalComponent`).

jizhi0v0:
New finding addressed in `b5799026`.

`installerWindowPending` was a local of `relaunchStagedUpdate`, and under Update All the batch's release runs after every row's `defer` has already gone — so nothing at that level could see it, and the unconditional `startOwedBackupTransferIfIdle()` after the walk opened exactly the window the hold existed to protect. Correct, and the trigger as described (staged row whose staging cannot be read → no hand-off armed → `quitHandoffs` empty at the batch release) is the cleanest case.

Promoted to `installerWindowPendingIds`, a fourth condition in `startOwedBackupTransferIfIdle`, cleared at the top of `relaunchStagedUpdate` for that row — a fresh attempt supersedes the hold its predecessor left. The comment at the un-armed `.wontQuit` branch ("leave what is owed for the next relaunch or install to drain") is now true on the batch path too.

Nothing is stranded by holding longer: `resumePending()` reads what is owed from the sidecars, and the normal install path still transfers immediately (`transferNow: true`), so a held copy is drained by the next install rather than waiting on another relaunch.

Also noted your two dropped items are right: the batch **restart** loop needs no flag (`restart` takes no backup — 6771 is the only `transferNow: false` call site), and the two early returns ahead of `relayQuitHandoff`'s `defer` lose nothing.

`BUILD SUCCEEDED`; `3517 tests in 302 suites` and `378 tests in 48 suites` pass. Real-device verification still outstanding (`make install` fails with `errSecInternalComponent`, login keychain locked).

jizhi0v0:
Round 4 clean — thanks for the four rounds, each one found something real.

On the three dropped items, agreeing explicitly so the trade-offs are on the record:

- **The hold is sticky.** Correct, and intended as a delay rather than a release: after an un-armed `.wontQuit` or a mid-backup stand-down, `startOwedBackupTransferIfIdle` stays a no-op until that row is retried. What is owed still moves — the normal install path transfers inline, and launch, the volume watcher, "Copy Now" and `releaseHeldBackupsAndSync` all drain. Given the alternative is opening an installer window we were asked to protect, erring toward "later" is the right side to err on.
- **`if !installerWindowPendingIds.contains(result.id)` is redundant** with the guard's `isEmpty` term. Left in deliberately: it states the caller's own reason for holding, independently of how the shared guard happens to be composed. Redundant, and cheap.
- **The not-running early return at 5746.** Agreed it is pre-existing and neither introduced nor widened here. Worth noting for a future change: it is the one place where "a fresh attempt supersedes the previous hold" can drain into a window that is genuinely open.

On the two unverified external claims: both still stand as flagged. The launchd/`ExitTimeOut` behaviour is written up with its evidence in https://github.com/Squirrel/Squirrel.Mac/issues/336 (including a correction there about which measurements were contaminated), and real-device verification of this PR remains outstanding — `make install` fails on this machine with `errSecInternalComponent` (login keychain locked), so the ordering still rests on the code reading. The `.info` → `.notice` change is what will make the next real relaunch diagnosable without a live `log stream`.

PR #773, opened 2026-09-20
SogouInput: one-click, by rotating Contents like the other three input methods

jizhi0v0:
搜狗输入法 gets a one-click, and it runs as the same `Contents` rotation WeType, DoubaoIme and QianwenIME already get.

## Why it wasn't done before

The audit said detection-only, and the main reason turned out to be **about the wrong package**: it described the website installer (two `/Library/LaunchAgents` plists, a QuickLook generator, a user-directory migration, `killall -9`). The self-update package — what the endpoint hands out, and what an ordinary release travels in — does none of that. Read off the real 6.25.1.11973 payload:

| script | what it actually does | why it isn't run |
|---|---|---|
| `pre.sh` | acts only when the bundle is **not** writable, then asks for admin and chowns it back | `InPlaceSwap.stageRotation` refuses that same case up front, with a sentence instead of a password panel — behaviourally equivalent |
| `post.sh` | wrapped entirely in `if [ $SOGOU_INPUT_VERSION == "3.2.0.68597" ]` | dead code on anything modern |
| `switch.sh` | the only teeth, and argument-gated: `switch.sh 1` does `rm -rf ~/Library/Application Support/Sogou/InputMethod` (the learned dictionary) before moving an older location over it; ends in `killall -KILL SystemUIServer` | not calling it is the answer, not a reason to skip the one-click |

⚠️ Those `killAll` lines (capital A) do run — the default volume is case-insensitive, so PATH lookup still finds `/usr/bin/killall`.

## What's new: `ContentsPayload`

Unlike the other three, this package carries **no `.app` at either level** — a zip of scripts wrapped around a second zip holding a bare `Contents<version>/`. `ArchiveExtractor` alone answers `noAppFound`, so `ContentsPayload` unpacks both levels and assembles a bundle for the gates to read.

Assembled, it verifies:

```
codesign -dv    → TeamIdentifier=DFD88F82SU   Identifier=com.sogou.inputmethod.sogou
codesign --verify --deep --strict → valid on disk / satisfies its Designated Requirement
spctl -a -t install               → accepted, source=Notarized Developer ID
```

Gates 2–6 are untouched. **The difference from `nestedArchivePath` is worth stating plainly**: DoubaoIme's stub is signed, so its signature is checked *before* the nested archive is opened. Sogou's outer archive is a plain zip — there is no stub to check. Trust rests entirely on gates 2/3/4 over the assembled bundle, which is where DoubaoIme's ultimately rests too. What this adds is shape guards: exactly one anchored `Contents<version>` directory, not a symlink, not a file, no extra sibling.

## The blocker that had to be fixed first

`InputMethodDataBackup`'s two general rules find **1 of SogouInput's 8** user-data locations. Nothing on disk is called `SogouInput` — the data is filed under the vendor name, and the settings pane writes `com.sogou.SogouPreference` (the same shape as WeType's `com.tencent.WeTypeSettings`, which the name rule exists for and which `SogouInput` cannot reach).

`declaredDataNames` states the extra name rather than deriving it. **The rejected inference**: taking the vendor token out of the bundle id works here and misfires next door — the same rule reads `com.tencent.inputmethod.wetype` as `Tencent` and would snapshot `~/Library/Application Support/Tencent`, shared with every other Tencent app.

## Two decisions worth reviewing

**`SGQuDao` is not preserved.** The installed `Info.plist` carries `SGQuDao = 1111`; the payload's does not, and no script re-injects it — so the vendor's own update drops it too. Verified: still absent after `launchctl kickstart` of `SogouServices`, `Info.plist` unchanged byte for byte. Re-injecting would invalidate a signature that just passed.

**The install URL comes from the response pinned at `v=0.0.0.1`**, not from a second request carrying the real installed version — against what the audit first suggested. Two measurements decide it: the endpoint does not stage (so the pinned request's payload is the newest one), and `sv` gates by OS (so a macOS 28 host's real version is answered with a 2023 build below every real install). Same response for the version and the bytes also removes the compare-one-release/download-another drift surface.

`update_pack_md5` is MD5 where `checksumPattern` is SHA-512/base64, so it is deliberately not wired rather than mis-declared — the same call WeType's `zip_download_md5` got.

## Verification

Full round trip on a real install: 6.24.1.11676 → 6.25.1.11973 → rollback → update again.

| check | result |
|---|---|
| `make test` | Core **3531** / CLI **378** / App **51**, all checker scripts (1 pre-existing known issue) |
| payload integrity | `md5` matches what the endpoint declares |
| `duo install SogouInput --yes` | `applied=true`, 183,534,090 B, 12.7 s |
| outer bundle inode | **unchanged across three installs and two rollbacks** — this is what keeps the input source registered |
| `codesign --verify --deep --strict` | `a sealed resource is missing or invalid` → **exit 0** (the app writes into its own bundle at runtime, so the installed copy was already invalid; don't use before/after signature as a discriminator) |
| user-data snapshot | **7 of 7** locations captured, where the general rules capture 1 |
| rollback | bundle + `SGQuDao` + data restored; a marker planted after the snapshot is gone, as it should be |
| input source | `com.sogou.inputmethod.sogou` still registered throughout |
| human check | typing, account login state and the About panel all fine |

**Mutation-tested**: 3/3 on the discovery change, 5/6 on the assembler. The survivor is the containment check — unreachable behind the symlink guard, and documented in place as the backstop it is rather than pretended to be pinned.

One test was caught being vacuous: the first anchor test named the inner archive after the directory, so the **outer** lookup refused and the anchor never ran (it passed with the anchors removed). The fixture now sets archive name and directory name separately.

## Known, and out of scope here

The About panel shows a stale version after an install or rollback, because rotation replaces the code on disk without interrupting the already-mapped process. Re-selecting the input source fixes it. This is the cost of never force-quitting, and `duo backups restore` already says so.

Separately, and **not fixed in this PR**: backup generations synced to an external backup disk have no `UserData` — the user-data snapshot does not survive the transfer, for any input method. `BackupStore.restore` only logs when it finds no snapshot, so a rollback silently restores the bundle alone. Tracked separately.


jizhi0v0:
Round 1 — 0 blocking. Addressed all three of the non-blocking notes in `d671f13`:

**Block-scoping** (the one you checked and decided not to block on). Scoped the install pattern the same way as `versionPattern`, so the version and the bytes cannot come from different blocks. You are right that it needs multi-block behaviour nobody has seen and that the outcome would be a refusal — I took it anyway because "the two patterns agree about which release is being installed" should hold by construction rather than by the server's habit. `aVersionIsNeverPairedWithAnotherBlocksPayload` already built the body that separates them; it now asserts the install pattern too, and reverting to the unscoped spelling makes it fail.

**The count.** Seven is right, and the inconsistency was mine. A location is what gets captured, and the support directory is captured whole — so its four children in the path list are one location, not four. Fixed in all three comments and stated explicitly next to the list so the next reader does not have to re-derive it.

**HTTPS** — you flagged this as checkable only against the live vendor, so I checked it. Both hosts the CDN alternates between answer over https:

```
pro.cdn.ime.sogou.com    206  application/zip
pro.cdn2.ime.sogou.com   206  application/zip
first 4 bytes: 50 4b 03 04
```

Recorded in the audit's verification table, with the reason it needs checking (`preferHTTPS` rewrites the scheme unconditionally).

Both patterns independently recomputed against the live response: `6.25.1.11973` and the matching payload URL; the sentinel matches neither. `make test` green — Core 3531 / CLI 378 / App 51.

jizhi0v0:
Round 2 — blocking finding fixed in `1e89858`, and you were right that it is the worse half of the change: the code got the safe pattern and the canonical record kept the unsafe one.

**Finding 1 — the audit documented the replaced pattern.** Fixed, and recorded with the reason rather than just the value, so the next reader knows why not to revert it.

**Same defect one section up, and it predates this PR.** While fixing yours I checked the neighbouring line: the audit documented `versionPattern` as `\nversion=(…)[\s\S]*?\nupdate_pack_url=` while the recipe has the `pid=0`-scoped `[^\[]*?` form. Not equivalent — `[\s\S]*?` crosses the brackets of `[end]` / `[product1]` and `[^\[]*?` cannot. Measured on a sentinel block concatenated with an update block:

```
documented spelling → 1.0.0.1     (the sentinel)
actual recipe       → 6.25.1.11973
```

So restoring the recipe from the audit would have reintroduced cross-block pairing through the *version* too. One finding, two instances; both fixed.

**Your non-finding is also correct, and the error was mine.** No revision of this audit ever wrote eight — checked `8c7d9ac7` and `6c09268b`, zero hits, both already `1/7`. The three were in the Swift comments and the test doc, exactly as the previous commit message said. The blockquote now says that, and has moved from `## 更新检测` to the `### 用户数据` section whose count it explains.

`make test` green: Core 3531 / CLI 378 / App 51.

One note on the earlier CI red, since it is on this PR's record: `test` failed on `8c7d9ac` with `App tests: 50 of 51 declared cases ran. Never executed: theClientRequirementIsPinnedForATeam`. `xcodebuild` exited 0; only the name-coverage gate objected, on a pure string-comparison case in `HelperPeerGateTests` — the suite `app_test_coverage.py`'s own comments name as the source of the log interleaving it already fought three times. It did not reproduce on `d671f13` or in six local runs, and this PR changes nothing under `App/`. Tracked separately rather than papered over here.

PR #774, opened 2026-09-20
Backups: carry the input-method user-data snapshot to the backup disk

jizhi0v0:
## The bug

`InputMethodDataBackup` snapshots an input method's user data (learned dictionary, settings, preferences) into `<key>/UserData/` beside the bundle rollback point. That snapshot is the safety net the input-method one-click was reinstated with in 0.3.25→0.3.29.

It did not survive the move to an external backup disk.

Measured on a real machine 2026-09-20: after backups were synced to the external destination, **no generation on that disk had a `UserData` directory** — each held `<App>.aar` + `backup.json` only, including `com.sogou.inputmethod.sogou-…` and `com.qianwen.inputmethod.desktopime-…`, both of which definitely got a snapshot when installed. The local store was empty, so the snapshots were gone.

## Where it was lost — evidence, not inference

Ran the real `transferToDestination` against a scratch destination and printed each step:

```
outbox before transfer  = ["backup.json", "UserData", "Fixture.app"]
destination after       = ["backup.json", "Fixture.aar"]
outbox after, exists    = false
contents of Fixture.aar = ["Contents"]      ← the bundle only
```

**Dropped at pack time, destroyed one line later.** The transfer archived `meta.bundleName` and nothing else, wrote a sidecar that looked complete, then `forceRemove(outboxDir)` deleted the only copy. Not a copy-time or prune-time loss — the snapshot was never a candidate for transfer. The restore side was equally blind: `InputMethodDataBackup.restore(forKey:)` only ever read `BackupStore.root`, which by then holds nothing.

Consequence: a rollback of an input method whose backup lives on the disk silently restored the bundle **alone** — the exact downgrade-onto-newer-data case the snapshot exists to prevent — and said so only in the log.

## What changed

**The snapshot travels.** Packed as its own `UserData.aar` beside the bundle's archive, with its own SHA-256 in the sidecar (`userDataArchiveName` / `userDataArchiveSHA256`, both optional so existing sidecars decode unchanged). Archived rather than copied for the same reason the bundle is: the destination is a filesystem we assume nothing about, and a snapshot restored without its xattrs and ACLs is silently not what was taken. Packed **before** the sidecar is written, so a transfer cut off anywhere leaves the outbox copy — snapshot and all — as the one complete generation.

**And comes back.** Restore unpacks it into the rollback scratch **before** the swap, so a rollback either has both halves in hand or knows it does not. `InputMethodDataBackup.restore(forKey:from:)` is a new overload that takes the directory instead of assuming the outbox. A corrupt or tampered snapshot archive **degrades** the rollback to bundle-only rather than failing it — the bundle half stays fatal, but a wrong dictionary written over the right one cannot be undone.

**The user is told.** `BackupStore.restoreReportingUserData` returns `RestoreOutcome{version, userData}` with `.notApplicable / .restored(n) / .noSnapshot / .failed`; the existing `restore(forKey:over:)` is now a one-line wrapper, so no caller broke.
- App: a bundle-only rollback sets the row's `installNotes` — *"X is back at the earlier version, but its dictionary and settings are not — this rollback point has no copy of them."* Plus a `.failed` variant. Both strings added to `Localizable.xcstrings` in all six translated languages.
- CLI: `duo backups restore` prints the warning under `restored`; `--json` gains `userDataRestored` / `userDataMissing` / `userDataError`.

**Generations already affected.** `BackupStore.generationsMissingUserDataSnapshot()` finds them, recognised by `UpdatePolicy.isInputMethod(meta.originalPath)` in whichever store, requiring the archive to actually exist rather than merely be named. `duo backups verify` marks them `[bundle only]` with an explanation. It does **not** fail the run: the bytes are fine.

No repair is offered, deliberately. The obvious one — snapshot the live data now — would attach the **newer** version's data to the **older** generation, so a later rollback would restore the bundle and then put back exactly the data the rollback was escaping, while reporting success. That is worse than the gap. The data lost on machines this happened to is not recoverable; the next update of each input method takes a snapshot that does travel.

**Sizing.** `duo backups list` now measures a destination generation by its directory rather than the bundle archive alone — DoubaoIme's snapshot is larger than most apps, and reading one file would report a 600 MB rollback point as the 40 MB half of it. `BackupStore.storeSize` already walked the directory, so Settings was already correct.

## Tests

New `BackupStoreUserDataTransferTests` — 7 tests, written as data round trips (write a dictionary → back up → transfer → outbox gone → an update mangles the data → roll back **from the disk** → the dictionary is back), plus no-snapshot, tampered-snapshot, ordinary-app, and detection cases. Four CLI tests for the warning, the JSON row, and the destination size.

Every new guard mutation-tested; each mutation confirmed red:

| mutation | caught by |
|---|---|
| don't pack `UserData` at transfer (the original bug) | 3 tests, incl. both round trips |
| restore never finds the destination snapshot | 2 round-trip tests |
| `snapshotExists` ignores `userdata.json` | the manifest test |
| destination size branch ignores its files | the new CLI size test |

## Verification

```
make test
━ Test run with 3524 tests in 303 suites passed after 34.250 s with 1 known issue
✔ Test run with 382 tests in 49 suites passed          (CLI)
✓ App tests — 51 of 51 cases executed
```
plus all 11 Python check/test scripts green. The one known issue is pre-existing (`EventStoreSchemaDriftTests`, schema drift) and unrelated. Core went 3517→3524 and CLI 378→382 from the new tests.

## For the reviewer

- **One untested seam:** `AppListModel.rollback` itself has no unit test — the App test bundle only executes pure functions and the model needs a whole app. The guard is at the Core level, which is tested; the App change is a `switch` over that outcome verified only by compilation.
- The new `switch` in `rollback` puts the user-data note **ahead of** the App Store note. They cannot both apply today (an input method is not a Mac App Store app); the ordering states which matters more if that ever changes.
- `transferToDestination` now throws if the snapshot archive fails, aborting the transfer. That matches what it already did for the bundle archive: no sidecar is written, so the half-made destination directory is not a backup, and the outbox copy survives for the queue to retry.


jizhi0v0:
Blocking finding confirmed and fixed in `f217144b` — the comment asserted the opposite of the truth. `archiveName(forBundle: "UserData.app")` is byte-for-byte `"UserData.aar"`, and `BundleArchive.archive` removes its target before renaming the `.partial` in, so the snapshot overwrote the bundle archive whose digest had just been recorded, with `forceRemove(outboxDir)` one line later. The snapshot name is now derived — `Foo.aar` against `Foo.UserData.aar` — so the two differ by construction for every bundle name. Regression test `anAppCalledUserDataDoesNotOverwriteItsOwnSnapshot` added as a round trip; reverting to the fixed name fails it with `backupCorrupted("UserData.aar")`, the predicted failure exactly.

All three unverified claims corrected too:
- The backup-disk measurement is now my own rather than relayed — listed the disk directly: 46 generations, **none** with a `UserData` directory, both named input methods holding `<App>.aar` + `backup.json` only.
- Dropped the invented "600 MB / 40 MB" for the 578 MB DoubaoIme figure `InputMethodDataBackup` actually records.
- "until 0.4.1" named a release this may not ship in (tree is at 0.4.0 build 109) → "before this change".

`make test`: 3525 Core / 382 CLI / 51 app cases, all checkers green.

jizhi0v0:
Round-2 blocking finding fixed in `f2891be5`. The example on `Meta.userDataArchiveName` is now `WeType.UserData.aar`, and the field records *why* the obvious name is the wrong one — `UserData.aar` is what `archiveName(forBundle:)` produces for `UserData.app`, so anything recognising snapshots by that string would be reading a bundle archive. The note points at `userDataArchiveName(forBundle:)` for the collision that cost.

Left alone deliberately: `Meta.archiveName`'s neighbouring `e.g. Slack-4.35.121.aar`. It has the same shape of error and predates this PR, and the review explicitly did not report it — out of scope here rather than overlooked.

The two "unverified external claims" need no change: the disk listing is my own measurement from this Mac (46 generations, none with a `UserData` directory), and the 578 MB DoubaoIme figure is the repository's own recorded measurement, now cited rather than restated.

`make test`: 3525 Core / 382 CLI / 51 app cases, all checkers green.

PR #775, opened 2026-09-20
App tests: read "which cases ran" from the .xcresult, not the console log

jizhi0v0:
## Why

CI run [35492361595](https://github.com/jizhi0v0/duo-updater/actions/runs/35492361595) failed the App name-coverage gate with:

```
✗ App tests: 50 of 51 declared cases ran. Never executed:
    theClientRequirementIsPinnedForATeam
```

`xcodebuild` exited 0 — Core, CLI and the App bundle all passed. Only the gate objected, on a commit that changes nothing under `App/`. That case is a pure string comparison with no I/O, no timing and no environment dependency, so it cannot "not run" for a real reason.

It was the console interleaving the gate already documents: a stderr log line and a swift-testing record glued together mid-line, in a tear shape the existing repair does not recover.

Evidence it was a flake, not a regression: the next run on the same branch (35492877522) passed, `scripts/app-tests.sh` was 6/6 green locally at 51 of 51, and the 12 most recent runs on `main` were green.

### 1. The sweep is committed this time — `scripts/sweep_app_test_coverage.py` (new)

The gate's comments quote a 2026-09-17 brute-force sweep "with a script that is not committed, so the numbers can't be re-run from this repo". That is exactly what made this false red expensive to classify: there was no way to tell "a tear shape we already know we miss" from "the parser regressed".

The sweep tears a record and a log line at **every** position, interleaves them every way that preserves each writer's order, and asks two separate questions:

- **SAFETY** — can a non-passing case be made to look like it ran? Any hit is a hard failure; this is the guarantee the gate exists for.
- **RECALL** — is a genuine pass still found? A miss is a false red, which costs a rerun.

Re-run 2026-09-20:

```
SAFETY — a non-pass record must never be counted, however it is torn
  record in 1, log in 1:      100 texts, 0 false passes  [ok]
  record in 2, log in 1:     5160 texts, 0 false passes  [ok]
  record in 1, log in 2:    11190 texts, 0 false passes  [ok]
  record in 2, log in 2:   769872 texts, 0 false passes  [ok]
  + a donor pass record, all three torn:   500000 texts, 0 false passes  [ok]

RECALL by interleaving order (both writers torn in two)
  log1|log1|record|record          149200/149200 recovered
  log1|record|log1|record           86800/149200 recovered   (62400 missed)
  log1|record|record|log1          149200/149200 recovered
  record|log1|log1|record          149200/149200 recovered
  record|log1|record|log1           84380/149200 recovered   (64820 missed)
  record|record|log1|log1          149200/149200 recovered
```

The misses land in exactly the two orderings the file's own comments name — the ones where the log line's own newline arrives *after* a piece of the record, so cutting the line out takes record text with it. The three-writer configuration puts a real `✔` from another case's pass record on hand, which is the dangerous one, and it is clean.

Quick mode runs on every `make test`.

### 2. The console parser is **not** loosened

Recovering those two orderings means allowing text between `)` and `passed` that the log line could have written — precisely how a torn non-passing record borrows a verdict it never earned (the counterexample family from the review of #716).

Three candidate repairs were measured against the sweep rather than argued about: stop the log-line cut at a marker; cut only the timestamp header; both. Recall moved **86.2% → 86.3%** across 179040 torn texts, with safety unchanged. Not worth the invariant, so none were taken. Both orderings are pinned as *known misses* in `KnownConsoleMisses`, so anyone who later makes them pass owes an argument for how the verdict still cannot be borrowed.

### 3. The gate now prefers the `.xcresult`

`xcrun xcresulttool get test-results tests` renders the result bundle as JSON: one node per case, a database rather than a byte stream, so no amount of concurrent logging can tear it. Measured on this toolchain (Xcode 27.0 27A266a, xcresulttool 25115, schema 0.4.0): 51 `Test Case` nodes for 51 declared cases, with the parameterized case as a single node spelled `aMalformedTeamYieldsNoRequirement(_:)`.

xcodebuild already wrote this bundle — CI extracts it on failure today, for exactly this reason ("the per-case results, which answer 'did that case run' without going through the console log the gate parses"). It is now named per process via `-resultBundlePath`, like the log, so "the newest bundle" is never a guess between concurrent runs.

The console parser stays as the **fallback**. An unreadable bundle is evidence about the toolchain, not about the tests, so a missing `xcrun`, a non-zero exit, unparseable output, or a schema this doesn't understand all hand over to the log rather than reporting an empty run. A genuinely empty run still fails, because the console finds nothing either. The gate prints which source answered and `app-tests.sh` echoes it, so a future red says whether it came from the structured source or the interleaving-prone one.

## The vacuity guard — and the trap inside it

Counting bundle nodes would have silently reintroduced the first of the three counting bugs the gate's docstring already records. Measured by putting `.disabled("mutation probe")` on `theClientRequirementIsPinnedForATeam` and rerunning:

```
Test Case nodes: 51
results: Counter({'Passed': 50, 'Skipped': 1})
[('theClientRequirementIsPinnedForATeam()', 'Skipped')]
```

The disabled case is **still listed, by name**. Unlike the console, which simply has no record for a case that never ran, the bundle names it anyway. So `Skipped` is excluded, and the real red→green was walked on the actual suite rather than inferred:

```
✗ App tests: 50 of 51 declared cases ran. Never executed:
    theClientRequirementIsPinnedForATeam
  Source: result-bundle. ...
EXIT=1
```

and restored:

```
✓ App tests — 51 of 51 cases executed (source: result-bundle)
```

## Mutation testing

Nine mutations of the new reader, all killed by `scripts/test_app_test_coverage.py`:

| Mutation | Result |
|---|---|
| drop `Skipped` from `DID_NOT_RUN` | killed |
| ignore `result` entirely | killed |
| count any node, not just `Test Case` | killed |
| empty bundle stops falling back | killed |
| anchor the name to empty parens | killed |
| console wins over bundle | killed |
| drop the JSON guard | killed |
| ignore xcresulttool's exit code | killed |
| drop the pass-marker class from `RAN` | killed |

Two of these initially **survived** and the tests were made discriminating rather than the mutations dropped:

- *count any node* — real `Arguments` nodes carry **no `result` key at all**, so the `Skipped` rule cannot reject them; only `nodeType` can. The test now uses a skipped parameterized case whose `Arguments` child is spelled like a Swift call.
- *ignore the exit code* — a stub that exits 1 with no output is rejected by the JSON guard anyway. The test now has xcresulttool print a well-formed but **incomplete** report and exit 1, which is the half-written-bundle case; trusting it would be a false red on a green run.

## Verification

```
━ Test run with 3517 tests in 302 suites passed after 23.055 seconds with 1 known issue.
✔ Test run with 378 tests in 48 suites passed after 6.415 seconds.
✓ App tests — 51 of 51 cases executed (source: result-bundle)
```

Gate's own tests: 33 passing. Both sources cross-checked against the same real run — `51 51 / result-bundle` and `51 51 / console-log`.

## For the reviewer

- `.github/workflows/ci.yml` needed a one-line glob change: `-resultBundlePath` moves the bundle out of `Logs/Test/` into the derived-data root, verified on a local run. The extracted `ci-xcresult/` output path is unchanged, so the upload step still matches. The raw bundle is still not uploaded (the 107 MB artifact problem in run 35087278533).
- The gate's stdout gained a third line (the source). `app-tests.sh` is its only caller.
- **Known residual, unchanged by this PR:** the console fallback still misses ~14% of torn texts. That is the safe direction — false red, never false green — and it now only applies when xcresulttool is unavailable. It is not fixed, and the sweep is committed so the next person can confirm that rather than re-derive it.


jizhi0v0:
Fixed in `9f54148f` (blocking finding) and `a928b592` (the two actionable unverified claims).

**Blocking — RECALL inventory never checked:** correct, and implemented rather than removed from the docstring, since that check is the reason the sweep is worth committing. The inventory is now two pinned sets (four orderings that must be fully recovered, two known to miss), compared in **both** directions: an ordering that starts missing is a regression; one that stops missing is good news but falsifies the percentages quoted in `app_test_coverage.py` and `KnownConsoleMisses`, so it has to be a deliberate edit; an ordering the inventory does not name also fails.

Verified with your exact trigger — tightening `SPLICED_LOG_LINE` so a whole log line no longer rejoins leaves SAFETY clean at zero false passes, and drops `record|log1|log1|record` from `149200/149200` to `86922/149200`; the sweep now exits 1 naming it and `make test` goes red. With the new check block deleted, that same mutation survives at exit 0 — so it is load-bearing, not decorative.

**A second hole of the same kind**, found while reading for the first: the sweep is itself a gate, and an enumeration that silently produced nothing would have printed `0 texts, 0 false passes [ok]` for every section and exited 0. Every section's text count is now checked; making `tears()` return `[]` exits 1 naming all six empty sections.

`test_app_test_coverage.py` asserts both claims separately plus vacuity (35 tests). That last assertion caught a mistake of mine in review: `assertNotIn("0 texts,", …)` also matches `20000 texts,`, so it was red against a healthy run — it is anchored on the whole field now.

**Unverified claims:** two were worth acting on.

- `--compact` **is** accepted here (it is how the 51-node measurements were taken), but you are right that it is a silent-demotion risk for nothing: the output is parsed, never read. Dropped, so it now matches `ci.yml`'s invocation.
- "xcodebuild refuses to overwrite an existing bundle" I had genuinely not tested. Measured now — `xcodebuild: error: Existing file at -resultBundlePath "…/existing.xcresult"`, with the build stopping before running a test. The claim holds; the comment quotes the error instead of asserting the behaviour.

The rest were measured during the change and already carry their toolchain versions: the node shapes and the `.disabled()` probe, the bundle landing at `$DD/app-tests-$$.xcresult` rather than `Logs/Test/`, and the sweep totals.

`make test`: Core 3517 / 302 suites, CLI 378 / 48 suites, App 51 of 51 (source: result-bundle), gate tests 35.

jizhi0v0:
Round 2 had no new blocking findings; this addresses the test-strengthening note and the one actionable unverified claim. Pushed as `c2096ecb`.

**"The success line is printed unconditionally, so deleting the check block leaves the test green."** Fair, and worth closing rather than leaving as a slow signal. The comparison is now `inventory_problems(table, totals)`, called by `main()` and called *directly* by the tests with synthetic tables — unchanged, an ordering that starts missing, one that stops missing, one the inventory doesn't name. Five mutations of it are killed: returning three empty lists, dropping any one of the three terms, and dropping a label from the pinned sets. A second test asserts the two pinned sets partition the six orderings `recall_by_shape` actually produces, so a typo in a label can't quietly create a hole.

Writing those tests found a real wart you didn't flag: an ordering the inventory doesn't name was reported **twice** — as `unknown` and also as `improved` — so the output told the author to go update percentages that don't mention it. Unknown labels are now excluded from the other two terms and reported once.

**`--compact` measurement not re-taken without the flag:** re-taken. Identical structure — 51 `Test Case` nodes all `Passed`, 7 `Arguments` nodes, parameterized case spelled `aMalformedTeamYieldsNoRequirement(_:)`. Only whitespace differs, as expected.

On the other two unverified items: the `-resultBundlePath` error text and the `86922/149200` figure are both my measurements from this machine, which the sandbox can't reproduce — flagging that honestly rather than claiming otherwise. The sweep is committed and runs in quick mode on every `make test`, so a CI run reproduces the inventory check even where it can't reproduce that specific number.

**On the failing `test` check —** it is not this PR. `BackupDestinationProbeTests.theFileSizeCeilingIsReadFromTheFilesystem(_:)` failed because `hdiutil create` returned 1 for the APFS case on the hosted runner; this PR touches only `scripts/` and `ci.yml`, nothing in Core and nothing affecting disk pressure. That test's own doc comment records the same failure on #763 ("four images being minted together is enough for `hdiutil` to refuse one on a hosted runner's disk"), and the last 8 completed runs on `main` are all green. I re-ran the job rather than touching the test.

Worth noting: that Core failure aborted `make test` **before** `scripts/app-tests.sh` ran, so the result-bundle path and the new `ci.yml` glob have not yet been exercised on a hosted runner — only the 35 gate tests were, and they passed. A green run is still needed to confirm that half.

`make test` locally: Core 3517 / 302 suites, CLI 378 / 48 suites, App 51 of 51 (source: result-bundle), gate tests 37.

jizhi0v0:
Round 3 had no new blocking findings. Both items you checked-and-dropped are now closed anyway in `e9d3ea37` — neither was a defect, both were cheap, and the first is squarely this PR's own theme.

**Residual: deleting the `regressed`/`improved`/`unknown` block from `main()` left every test green.** You were right that the docstring stated this honestly rather than claiming otherwise, but a guard that can silently stop checking is exactly what this PR exists to fix, so leaving it documented wasn't good enough. The sweep now takes `--selftest-inventory`, which injects one ordering's regression into an otherwise healthy run — the exact shape of the 2026-09-20 false red — and a test asserts end to end that `main()` reports it and exits 1. It only ever makes the sweep stricter.

Both call-site mutations are now killed, verified: deleting the `regressed` branch from `main()` → exit 0, test fails; dropping just its `failed = True` so the message prints but the sweep still returns 0 → test fails on the return code alone.

**Miscount:** correct, and fixed. Only three of the four assertions are change cases; the first is the unchanged control. Reworded precisely, since as you note `unittest` stops at the first failing assertion — the mutation makes the test fail *at* the first change case. The control is there to separate "the comparison reports nothing" from "it reports the wrong thing".

**On CI —** the `test` check has now failed twice, and neither failure is this PR:

- `a928b592`: `BackupDestinationProbeTests.theFileSizeCeilingIsReadFromTheFilesystem(_:)` — `hdiutil create` returned 1 for the APFS case.
- `c2096ecb`: `HomebrewSelfUpdateTests.theTimeoutCountsFromTheShellsLaunch()` — a real `zsh -l -i` didn't finish inside its 5 s timeout; the case took 46 s.

Different tests, both in `DuoUpdaterCore`, which the Makefile runs several steps *before* `scripts/app-tests.sh` — so nothing this PR changes has executed yet when they run. Both are timing/load sensitive and say so in their own comments ("four images being minted together is enough for `hdiutil` to refuse one on a hosted runner's disk"; "the task hops around it on a loaded 3-core runner"). The last 8 completed runs on `main` are green.

I'm treating them as runner-load flakes rather than touching either test. Being explicit about the limit of that call: two failures on two runs, against zero in the last 8 on `main`, is a pattern worth a real look rather than endless re-runs — if the run on `e9d3ea37` fails again, especially with a *third* different test, that deserves investigation on its own rather than another retry.

Still outstanding either way: because both failures aborted `make test` before `scripts/app-tests.sh`, the result-bundle path and the new `ci.yml` glob have **never yet run on a hosted runner**. Only the gate's own tests have, and they passed. That half still needs a green run before this merges.

`make test` locally: Core 3517 / 302 suites, CLI 378 / 48 suites, App 51 of 51 (source: result-bundle), gate tests 38.

jizhi0v0:
**CI is green on `e9d3ea37`** — `test` pass (7m9s), `review` pass. No fix was needed; the failure notice reflected the earlier `c2096ecb` state.

```
Ran 38 tests in 22.250s                       ← gate's own tests
━ Test run with 3539 tests in 304 suites passed … with 1 known issue.
✔ Test run with 382 tests in 49 suites passed …
✓ App tests — 51 of 51 cases executed (source: result-bundle)
```

**Correcting myself:** I said twice that the result-bundle path "has never yet run on a hosted runner". That was wrong, and I should have checked the full run history before asserting it — `97c692c8` was fully green and already printed `✓ App tests — 51 of 51 cases executed (source: result-bundle)`. I had looked only at the two failing runs. The path had been working on CI since the first commit; `e9d3ea37` now confirms it again with the `--compact`-free `xcresulttool` call as well.

Also correcting the count: there were **two** genuine CI failures, not three — `a928b592`'s hdiutil case and `c2096ecb`'s Homebrew timeout. The runs I listed as failures for `9f54148f` and `a928b592` are recorded as *cancelled*, superseded by the next push.

So the flake read holds, and this is the evidence for it: two different timing-sensitive `DuoUpdaterCore` tests, on runs that bracket two fully green ones (`97c692c8`, `e9d3ea37`) on the same branch with the same code path. Neither test was touched.

One residual, stated rather than claimed closed: the `ci.yml` glob change runs only under `if: failure()`, and on both failing runs `make test` aborted in Core before `app-tests.sh` produced a bundle, so the glob matched nothing (`nullglob`, no-op). It is still unexercised on CI. It is a diagnostics-only step guarded by `|| echo "::warning::"`, so a mismatch degrades to a warning rather than a failed build, but I have verified it only locally.

jizhi0v0:
Round 4: 0 blocking findings, and CI was green on `e9d3ea37` (`test` 7m9s, `review` 3m9s). One item acted on, in `5da6f852`.

**"Exact shape of the 2026-09-20 false red" is loose.** You filed this as not-a-defect, but it is a comment of mine that says something untrue, and the repo's rule about claims in comments has no severity threshold — so it is fixed rather than accepted.

You are right about the distinction and it is worth more than a wording tweak: the 2026-09-20 incident was a tear landing in an ordering the parser is **already known to miss**, which is exactly why it sits in `EXPECTED_WITH_MISSES` and why the sweep stays *green* on it. What `--selftest-inventory` injects is the opposite case — an ordering that currently recovers fully starting to miss. Same consequence (a genuine pass reported as never run), different cause, and that difference is the entire reason the inventory is two sets rather than one. The comment and the test docstring now both say so, including what the earlier draft got wrong, so the distinction is not re-lost.

No behaviour change. Verified anyway: healthy run exits 0, `--selftest-inventory` exits 1, 38 gate tests green.

**On the two unverified claims:** both are local runs the sandbox can't reproduce, and I'd rather point at independent evidence than repeat myself. CI on `e9d3ea37` ran the whole thing on a hosted runner — `Ran 38 tests` for the gate, then `Test run with 3539 tests in 304 suites passed`, `382 tests in 49 suites passed`, and `✓ App tests — 51 of 51 cases executed (source: result-bundle)`. The counts differ from my local ones (3517/302, 378/48) because CI sets `DUO_DOWNLOAD_GATE=1`, which enables additional cases.

Since that also settles it: the result-bundle path is confirmed working on a hosted runner, including the `--compact`-free `xcresulttool` call. The one thing still unexercised on CI is the `ci.yml` glob, which only runs under `if: failure()` and needs a failure that gets as far as `app-tests.sh` producing a bundle.

jizhi0v0:
Round 5: 0 blocking findings, and CI is green on `5da6f852` (`test` 6m44s, `review` 3m41s). Nothing to fix.

One thing in round 5 *is* worth resolving — the note that "only the author can confirm which run produced which number", about local `3517 tests in 302 suites` / `378 in 48` versus CI's `3539 in 304` / `382 in 49`.

**I gave the wrong cause for that gap, and the hypothesis in round 5 is also wrong.** I said CI sets `DUO_DOWNLOAD_GATE=1` "which enables additional cases" — it can't: that variable is a `guard … else { return }` *inside* a test body (`VendorInstallTests.swift:928`), so the case still runs and still counts. Round 5 suggested environment-conditional suites (`.enabled(if: HomebrewInstaller.brewPath() != nil …)`, `.enabled(if: … hdiutilAvailable)`); those are satisfied on this machine too, so they don't explain it either.

Rather than offer a third guess I diffed the suite *names* — CI's log against a local `swift test` of both packages. Exactly three suites run on CI and not locally:

```
BackupStoreUserDataTransferTests
ContentsPayloadTests
BackupsUserDataWarningTests
```

None of them exists in this branch's tree. They exist on `origin/main`:

```
origin/main:DuoUpdaterCore/Tests/DuoUpdaterCoreTests/BackupStoreUserDataTransferTests.swift
origin/main:DuoUpdaterCore/Tests/DuoUpdaterCoreTests/ContentsPayloadTests.swift
origin/main:CLI/Tests/DuoKitTests/BackupsTests.swift        ← BackupsUserDataWarningTests
```

The cause is simply that **GitHub builds the PR merge commit, not the branch tip.** This branch forks at `c3c5e65c`; `main` has since gained #773 and #774, which add those suites. CI therefore runs my branch merged with current `main` — two extra Core suites and one extra CLI suite, which matches the deltas exactly (+2 Core suites, +1 CLI suite; CLI 48→49, 378→382).

So both sets of numbers are real: mine are this branch alone, CI's are this branch merged with `main`. Nothing environment-conditional is involved. Those two commits touch backups and input-method `Contents` rotation, nothing under `scripts/`, so there is no interaction with this PR — and CI has already exercised the merged result green.

PR #782, opened 2026-09-21
Stop offering updates that never land: Air vs Toolbox, and a leftover App Store copy

jizhi0v0:
Two rows that kept offering an update the user could not get.

## Air: compare against what Toolbox actually offers

`ToolboxSource` checked Air (no `productCode` in Toolbox's `state.json`) against Air's own Sparkle feed, retargeted to the Toolbox channel. On 2026-09-21 that feed had moved ahead of everything the user can install:

| Source | Latest |
|---|---|
| `fleet-feed/AIR/eap/macos_aarch64/feed.xml` (Sparkle) | **262.991.1** (pub. 09-18) |
| `data.services.jetbrains.com/products/releases?code=AIR&type=preview` | 262.834.44 |
| Toolbox catalogs `eap.feed` / `public-feed-arm.feed` (decoded) | 262.834.44 |
| `download.jetbrains.com/product?code=AIR&latest&type=preview` → | `Air-262.834.44-aarch64.dmg` |

The row read `262.834.44 → 262.991.1 [Toolbox]`, and Toolbox showed nothing to install.

**Change:** Air now goes through the same releases-API path as the IDEs, with code `AIR`. Toolbox calls its Public Preview quality `eap`, but the API calls it `preview` (`type=eap` returns `{}` for AIR), so the non-release channel asks for `eap,preview`. Fleet keeps the Sparkle path. The API does list Fleet as `FL`, but it is discontinued and I had no install to verify its build ids against.

## AndroMeld: the App Store update landed on a different copy

The vendor renamed the bundle from `AndDrive.app` to `AndroMeld.app`. Both copies are in `/Applications` with the same bundle id `com.catchingnow.andfiles`, ADAM 6762439757, and a `_MASReceipt`. The App Store keeps updating `AndroMeld.app`, which is already at 1.10.0. `AndDrive.app` stays at 1.9.0. Each Update click on the AndDrive row went like this:

1. It ran `mas`. FSEvents show the writes landing on `AndroMeld.app` at 15:21:59, 15:35:06 and 15:35:25, each before `mas` returned.
2. It backed up the untouched AndDrive.app, 98 MB each time.
3. It re-read AndDrive.app and logged `install done: AndroMeld now 1.9.0`.
4. It showed **Updated ✓**, because `appliedNothing` deliberately skips `.appStore`, and then offered the same update again.

**Change:** a new `AppStoreLeftoverCopy.sibling(of:target:among:)` finds another store copy with the same bundle id (and the same ADAM id when both have one) that is already at or past the target version. `AppListModel.performInstall` uses it in two places:
- **Before the install**, and before the backup: the install is skipped and the row says `The App Store updates AndroMeld.app (already 1.10.0), not this copy — this one is left over and can be moved to the Trash.`
- **After the install**: if the row's own version didn't move but a sibling now holds the target, the row shows the same message instead of ✓. This covers the case where the store creates the renamed copy during the click itself.

Candidates come from rows that share the bundle id plus `NSWorkspace.urlsForApplications(withBundleIdentifier:)`, and each one is re-read from disk. The CLI already refuses `.appStore` routes (`Install.swift`), so it needs no change.

### Follow-up: the orphan row stops offering Update at all

Making a click that could never work say so was not enough. The row should not offer the click in the first place. `AppStoreLeftoverCopy.orphans(in:)` marks a store row as an orphan when a sibling row already holds its offered version. It is memoized off `results.didSet`, so it holds through rescans. An orphan row:
- is not counted by `UpdatePolicy.isActionableUpdate` / `needsAction` (new `isOrphanedStoreCopy` parameter), so it is left out of the badge, "N updates available", Update All and notifications;
- gets a new `RowActionState.orphanedStoreCopy(updatedCopy:version:)` rung instead of `.updateAvailable`. The popover shows an orange ⚠︎ that opens an explanation with **Show in Finder**. The workbench shows an orange "Left-over copy" label, with the explanation on hover.

The click-time disk check from the first commit stays, for a sibling the store creates during the click. The post-install arm now reads `updated.hasUpdate`, not `isActionableUpdate`, because the latter now excludes exactly the case that arm is for.

Live check: with Amp's update settled, the rows reporting `updateAvailable` were AndroMeld 1.9.0, Surge and Xcode. After `make install`, the app logged `refresh done: 2 updates` and the menu-bar badge read 2, so AndroMeld is no longer counted. The new rows have not been seen in the popover yet, because that needs a human click.

## Verification

- `swift test` (Core): 3555 tests passed (after the orphan commit).
- New tests: `ToolboxAirReleasesAPITests` stubs the network with the 09-21 answers (API 262.834.44, Sparkle 262.991.1) and expects no update and one request, to `code=AIR` with `preview`. `AppStoreLeftoverCopyTests` has 7 cases.
- Mutation checks: emptying the Air code map fails the Air test (4 issues). Dropping the target-version condition fails `aSiblingStillBehindTheTargetIsNotTheAnswer`. Dropping the ADAM-id check fails `onlyStoreCopiesOfTheSameProductCount`.
- `check_localizable_keys.py` and `check_localizable_specifiers.py` pass. The new string is translated into de/es/fr/ja/ru/zh-Hans.
- After `make install`, the live app logs `Air [com.jetbrains.air] v262.834.44: Toolbox → 262.834 (hasUpdate=false)`.
- **Verified live (2026-09-21 16:11):** a real click on the AndDrive row logged `install skipped: AndroMeld at /Applications/AndDrive.app — the App Store copy at /Applications/AndroMeld.app is already 1.10.0`. There was no `mas` run and no backup, and the row showed the note. The later wording change (98c0cb8) is installed but hasn't been re-clicked.

- **Verified by the user (2026-09-21):** the orphan row, in both the popover badge and the workbench label, opens the explanation panel, and Show in Finder works.


jizhi0v0:
Round 1 had 0 blocking findings. I checked the unverified claims on the dev machine, so no code change:

- **`type=eap,preview` → `releases.first` is safe.** With `latest=true` the API returns a single entry: the newest release across all the requested types. Commands run 2026-09-21:
  - `code=AIR&latest=true&type=eap,preview` → `[('262.834.44','preview','2026-09-16')]`
  - `code=IIU&latest=true&type=eap,release` and `type=release,eap` → both return only `[('263.5153.40','eap','2026-09-18')]`, so the result doesn't depend on type order.
  - `type=eap` for AIR alone → `{}`.
- **Bundle id.** `mdls ~/Applications/Air.app` (the Toolbox-managed install) → `kMDItemCFBundleIdentifier = "com.jetbrains.air"`. There's no EAP suffix.
- **Air 262.991.1 vs 262.834.44.** `curl -L` on `fleet-feed/AIR/eap/macos_aarch64/feed.xml` → `<sparkle:version>262.991.1`, pubDate 18 Sep. Toolbox's `eap.feed.xz.signed` and `public-feed-arm.feed.xz.signed` (CMS-unwrapped, then `xz -d`) contain `262.834.44` and never `262.991.1`. The `download.jetbrains.com/product?code=AIR&latest&type=preview` redirect → `Air-262.834.44-aarch64.dmg`.
- **Fleet.** `code=FL` → last entry `Fleet-1.48.261`, dated 2025-05-20. Fleet isn't installed here.
- **AndroMeld FSEvents.** From `log show` on UserEventAgent `fsevents.matching` for `/Applications/AndroMeld.app`: writes at 15:21:59, 15:35:06 and 15:35:25. DuoUpdater's `mas … install ok` lines are at 15:22:04, 15:35:09 and 15:35:27. `AndDrive.app` Info.plist is still 1.9.0 (mtime 09-13).
- **`_MASReceipt` premise.** This is a design choice, not something I measured. A copy without a receipt isn't App Store provenance, so it's left out on purpose rather than guessed at.
- **Tests and l10n scripts.** I ran these on macOS: `swift test` passed 3551 tests, the three mutation checks each failed as expected, and both `check_localizable_*` scripts reported ✓.

jizhi0v0:
Round 2 had 0 blocking findings. I checked both unverified claims on the dev machine:

- `PlistBuddy -c "Print :CFBundleName" -c "Print :CFBundleDisplayName" /Applications/AndDrive.app/Contents/Info.plist` → `AndroMeld` / `AndroMeld`. The two copies really do share one display name, and the file name is the only thing that tells them apart.
- Click path: a real click at 16:11:04 logged `install skipped: AndroMeld at /Applications/AndDrive.app — the App Store copy at /Applications/AndroMeld.app is already 1.10.0`. There was no `mas` run and no backup. I updated the PR body to match. 98c0cb8 only changes the wording, and `strings` on the installed binary plus `zh-Hans.lproj/Localizable.strings` confirm the new string shipped. It hasn't been re-clicked since.

jizhi0v0:
Round 3 blocking finding 1 (no gallery tile for `.orphanedStoreCopy`): fixed in the latest commit.

- Added `50-orphaned-store-copy` (the badge) and `51-orphaned-store-copy-explanation` (the panel, drawn through `orphanHintPopover`, with the workbench half registered in `mayBeBlank`).
- The popover badge is the same amber borderless triangle as 18/22/43/46/47. So it goes in `notFaithful` under cause 1, and its five collisions are listed in `mayLookAlike` (not claimed as tooltip-differentiated).
- `make gallery` output: "no state renders blank", "no unregistered ImageRenderer placeholder pixels", "no two states draw the same tile". The four new PNGs are committed.

On the stale test count in the PR body: after the orphan commit, `swift test` (Core) passed 3555 tests on macOS. `AppStoreLeftoverCopyTests` now has 7 cases and `BadgeAndNeedsActionTests` gained `anOrphanedStoreCopyDoesNotCount`. Deleting the count guard or the rung makes its test fail.

jizhi0v0:
Round 4 blocking finding 1 (stale count above `mayBeBlank`): fixed in the latest commit. The comment now reads "for the later four, 43/46/47/50; these seven names".

On the two unverified claims:
- `make gallery` exits 0 (re-run on this commit, `echo $?` → 0). One note: on this Mac it also rewrites `27-managed-app-store` / `31-up-to-date-app-store` (both surfaces). Those tiles draw the host's App Store icon, the drift is local, and this PR changes nothing that draws them, so they're restored and not committed.
- The PR body already says 3555 tests and 7 cases (edited after round 3). `.claude-review/pr.md` is a snapshot taken before that edit.

PR #784, opened 2026-09-21
Workbench: the Left-over copy label opens the same panel as the popover

jizhi0v0:
Follow-up to #782, which was squash-merged at `0d7a179`. This one commit landed on the branch after that and never reached main.

## What changed

In #782 the workbench drew `.orphanedStoreCopy` as a hover-only "Left-over copy" label. The popover's ⚠︎ opens a panel that explains why the row can't update and offers **Show in Finder**. The workbench had no way to reach that panel, so the two windows disagreed.

- The panel now lives in one view, `OrphanedCopyPanel` (`RowActionViews.swift`). The popover's `orphanHintPopover` returns it.
- The workbench label is now a `.plain` button that opens the same panel in a popover.

## Verification

- `make gallery` exits 0 ("no state renders blank", "no unregistered ImageRenderer placeholder pixels", "no two states draw the same tile"). `workbench/50-orphaned-store-copy.png` is byte-identical, because the plain button draws exactly what the label drew. The only other changes, the host-dependent `27-managed-app-store` / `31-up-to-date-app-store` tiles, were restored and are not committed.
- `make install`, then the user clicked the label in the workbench on the real AndroMeld (AndDrive.app) row: the panel opens and Show in Finder works.
- `git diff 0d7a1797 origin/main -- App DuoUpdaterCore` is empty, so this is the verified commit cherry-picked onto the merged main (`ea30082` = `e34619d`).


jizhi0v0:
Round 1 blocking finding 1 (workbench/51 exempted as "no panel" while the workbench now opens one): fixed in 2713d50.

- `workbenchTile` draws `OrphanedCopyPanel` for `51-orphaned-store-copy-explanation`. It's the same panel the workbench label opens, and the committed `workbench/51` PNG is now that panel instead of blank.
- `workbench/51` is removed from `mayBeBlank`, and its comment is back to six names / 43/46/47, with a line saying why 51 isn't there.
- The `workbenchTile` doc comment and the `explanationCaseNames` comment now name 51 as the exception.

On the unverified claims:
- **plain button vs placeholder:** `make gallery` on this commit exits 0 with "no unregistered ImageRenderer placeholder pixels". `workbench/50-orphaned-store-copy.png` is unchanged in the diff, so the `.plain` button renders the glyph the same as the bare `Label` did. Cause 1 is specific to `.borderless`.
- **live click:** confirmed by the user on the real AndDrive.app row: the panel opens and Show in Finder works.
- **27/31 App Store tiles:** again rewritten by this host's App Store icon and restored, not committed.

